PR #52, opened 2013-01-08
Support for parallel test configuration

marchof:
I've maven-surefire plugin configured with this options :
Concurrency config is parallel='classes', perCoreThreadCount=true, threadCount=4, useUnlimitedThreads=false  

This means, that maven runs 4 separate JVM, each JVM has own jacoco-agent and all agents needs to write to the same target/jacoco.exec file.

After all tests are finished, I've malformed target/jacoco.exec file.

```
Caused by: java.io.UTFDataFormatException: malformed input around byte 0
    at java.io.DataInputStream.readUTF(DataInputStream.java:639)
    at java.io.DataInputStream.readUTF(DataInputStream.java:547)
    at org.jacoco.core.data.ExecutionDataReader.readExecutionData(ExecutionDataReader.java:145)
```

Is there a synchronisation/write locking for target/jacoco.exec ?

I think that this is continuation of http://sourceforge.net/apps/trac/eclemma/ticket/191

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -25,7 +25,8 @@
   <li>Support for offline instrumentation (GitHub #4).</li>
   <li>JaCoCo agent exposes runtime API for direct integration with application
       under test (GitHub #61).</li>
-  <li>Support for parallel test execution (GitHub #52).</li>
+  <li>Support for parallel test execution: Different agents can now safely write
+      to the same <code>*.exec</code> file (GitHub #52).</li>
 </ul>
 
 <h3>Fixed Bugs</h3>


Raynevun:
Still reproducible for me, when using TeamCity 9.1.1 (build 37059 which should contain JaCoCo 0.7.5) integration and running unit and integration tests in parallel with two agents writing in one exec file.

Godin:
@Raynevun The best place to get support is our users mailing list - https://groups.google.com/forum/#!forum/jacoco , so could you please switch to it? And describe problem more detailed, at least provide stacktrace of exception, also maybe you know a way to reproduce this issue without need to install TeamCity?

Thanks for your understanding.

Issue #50, opened 2012-11-27
include commandline tool for merge/reporting of jacoco results in main source tree/final jars


judovana:
Many projects are using eg Make or another commandline based toll as build framework and for launching of different testsuites.

It would be nice to have (except excellent API you have already now) direct commandline tool for merging of exec files and for report generating. I have write one by myself for this task - http://mail.openjdk.java.net/pipermail/distro-pkg-dev/attachments/20121127/ba8f6a1e/jacoco-tool-0001.diff - ( from http://mail.openjdk.java.net/pipermail/distro-pkg-dev/2012-November/020984.html) and I will be happy to help with inclusion of such a tool into your repository.

Thanx for this amazing tool,
J.

marchof:
Will be available with PR #525.

PR #86, opened 2013-03-11
Add CLI tools for offline instrumentation and report generation

johnkeeping:
These tools help people who are using a build system other than Ant/Maven, as discussed in issue #50.

To that end, I also include a commit adding some OSGi metadata to the runtime agent JAR.  This allows that JAR to be stored in a bundle repository and downloaded from there during the build.

### DIFF
--- a/org.jacoco.agent.rt/META-INF/MANIFEST.MF
+++ b/org.jacoco.agent.rt/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JaCoCo Agent Runtime Implementation
+Bundle-SymbolicName: org.jacoco.agent.rt
+Bundle-Version: 0.6.3.qualifier
+Bundle-Vendor: Mountainminds GmbH & Co. KG
+Bundle-RequiredExecutionEnvironment: J2SE-1.5
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -79,6 +79,7 @@
     <module>../org.jacoco.agent.rt</module>
     <module>../org.jacoco.agent</module>
     <module>../org.jacoco.ant</module>
+    <module>../org.jacoco.cli</module>
 
     <module>../jacoco-maven-plugin</module>
 
@@ -125,6 +126,7 @@
     <!-- Dependencies versions -->
     <asm.version>4.1</asm.version>
     <ant.version>1.7.1</ant.version>
+    <args4j.version>2.0.16</args4j.version>
     <junit.version>4.8.2</junit.version>
 
     <!-- ================== -->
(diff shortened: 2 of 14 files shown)


somechris:
Since this PR is still waiting for a merge, and I needed a CLI tool for the report
generation part, I threw something similar together at

  https://github.com/quelltextlich/jacoco-toolbox

. The jar (available [from central](http://search.maven.org/remotecontent?filepath=at/quelltextlich/jacoco/jacoco-toolbox/0.4103.1/jacoco-toolbox-0.4103.1.jar), so no compilation necessary) allows
to access JaCoCo's functionality to

* merge inputs, and
* generate CSV, XML, and HTML reports

directly from the command line.


Maybe it's useful for other people that find this PR in the future.

mathjeff:
This would be helpful for me too. Android would like to be able to have code coverage metrics via jacoco, but the classes later get processed further and executed in a vm that isn't a jvm (hence wanting the offline instrumentation), and the build system is a custom (though the portion that would interact with Jacoco is essentially Make) build system (hence wanting a command-line tool that both starts a couple seconds faster than Ant or Maven, and that is slightly simpler than including Ant or Maven in the build process).

@marchof you indicated in https://github.com/jacoco/jacoco/issues/520 that some of the main blockers for getting this merged were tests and documentation. Could you elaborate on what kinds of documentation and tests are expected? For documentation, are you thinking we'd just add some html under the file path /org.jacoco.doc/docroot in the same commit (as opposed to also separately updating some external wiki)? For tests, are you thinking full integration tests that build the .jar, execute it with command-line arguments, and check the resultant output files, or do you think it would be fine to just test the parser itself in memory?

I'd be happy to write some tests and/or documentation if that would get this to be merged

Thanks

marchof:
@mathjeff Thanks for following this up here! Regarding missing parts:

### Documentation

All official documentation goes to module `org.jacoco.doc`. External wiki is only for development purpose (e.g. design documents for new features) and not considered documentation. For the command line tool I envision that the documentation is generated from the annotated source code, like we do for [Maven](http://www.jacoco.org/jacoco/trunk/doc/maven.html). Actually I started a new implementation some time ago to do exactly this. The idea was to work "ducumentation first" to identify the minimal required feature set before starting the actual implementation.

### Test

Testing the built cli.jar as a separate process would probably the integration test. I think we already do the same for our [Maven goals](https://github.com/jacoco/jacoco/tree/master/jacoco-maven-plugin.test).

mathjeff:
@marchof Thanks so much; that looks like it should give me a good understanding of what you're looking for.

I'm starting some updates at https://github.com/mathjeff/jacoco/commits/updating-cli-of-johnkeeping but I haven't yet implemented the test and documentation changes that you're talking about; once I do that then I'll turn it into a proper pull request.

Thanks so much!

marchof:
@mathjeff I also started working on the documentation and test part. I hope I can push something by next week so we can see what is left for your use cases.

marchof:
We follow-up the command line interface implementation in a new pull request: #525

PR #97, opened 2013-05-02
Add aggregate-report that reports cross module code coverage for the maven plugin

johnoliver:
We have had issues with gaining accurate cross module code coverage numbers with jacoco. For instance when code in module A is touched during tests from module B, in our case when they are both part of a multi-module build.

This pull request makes it possible to report coverage numbers on classes outside of the current module. By adding a list of sourceFolders and classFolders which is a list of source and class folders outside of the current module to be scanned for classes to report coverage on.

The aggregate-report then concatenates the jacoco.exec files from reactor projects so that a unified report can be created.

PR #217, opened 2014-05-23
Allows the report mojo to reference binary/source files in other modules

chrisruffalo:
I've been working with a multi-module report where I wanted the ability to create coverage reports based on the merged reports.  I've really only been able to find good ways of doing this that all involve hacks.

So here's my counter-hack.

I'm not sure this would be accepted cleanly but I would appreciate some feedback.  

Thanks!


marchof:
We will go with a more generic solution as proposed in #97 or #388.

PR #237, opened 2014-09-17
Added 'verbose' mode for the check goal/task

nesvold:
```
    - Each limit check now potentially reports the current coverage for that particular metric even when there are no violations.
    - verbose is by default 'false' for the maven plugin. The equivalent for ant is running with the -v (verbose) flag set.
```


gzm55:
any updates for this pr?

PR #272, opened 2015-01-19
Conflict between jmockit and jacoco

marchof:
Related to #35 but it seems to still happen.

This project reproduces the issue: https://github.com/henri-tremblay/jacocomockit

When the line in the readme is executed, you will get this exception:
`java.lang.UnsupportedOperationException: class redefinition failed: attempted to change the schema (add/remove fields)`

jmockit 1.14 and jacoco 0.7.2.201409121644 is used

### DIFF
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
@@ -21,6 +21,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.instrument.IllegalClassFormatException;
+import java.security.CodeSource;
+import java.security.ProtectionDomain;
+import java.security.cert.Certificate;
 
 import org.jacoco.core.JaCoCo;
 import org.jacoco.core.runtime.AbstractRuntime;
@@ -41,6 +44,8 @@ public class CoverageTransformerTest {
 
 	private ClassLoader classLoader;
 
+	private ProtectionDomain protectionDomain;
+
 	private StubRuntime runtime;
 
 	@Before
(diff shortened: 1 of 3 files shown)


marchof:
This change has caused at least two regressions: #288 and #294 
We need to find a way to fix this:

1) Rollback
2) Make it optional (off by default)

Godin:
@marchof what about option 3 - updates in upstreams (#294 - Powermock , #288 - something in stack for Android) ? ;)

marchof:
If that is a realistic option...

Godin:
@marchof and what about another option, which is originated in [JMockit issue (#125)](https://github.com/jmockit/jmockit1/issues/125): remove field on our side, so that to avoid root cause of this PR - "class redefinition failed" ?

marchof:
@Godin This would cause a significant performance problem: For *every* method invocation we would need to access the JaCoCo runtime.

Godin:
@marchof As stated in referenced ticket - they don't observe such problem with jmockit-coverage. Would be nice to have a benchmark as a proof of such problem. Also I'm wondering how this is done in JCov.

henri-tremblay:
I'm a bit lost between the issue. However, provided with a before / after
code. I'm willing to perform the benchmark

Godin:
@henri-tremblay currently there is no before/after , because we were discussing possible options.

rliesenfeld:
I believe this fix should stay as it is. See discussion at https://github.com/jacoco/jacoco/issues/294

A performance benchmark comparing JaCoCo to JMockit Coverage would be interesting; I may create one myself, eventually. Just for awareness, one significant performance issue I noticed in JaCoCo is that, by default (ie, when users don't specify the classes to be included or excluded, which seems to be the common case), it will instrument all classes from third-party libraries; obviously, users generally don't want to measure coverage for such classes, resulting in a lot of unnecessary instrumentation. JMockit Coverage avoids this issue by only instrumenting classes that don't come from a jar file, except for classes explicitly selected for coverage; this may result in a significant performance gain for typical usage scenarios.

marchof:
@Godin I added a separate PR #296 for instrumentation without adding members.

pedrocborges2:
> This will prevent any dynamically generated class from being instrumented.

We use apache-tapestry-ioc and this change just moved all our tapestry service coverage to 0%
This change prevents any code in actual classes called through dynamically generated classes from being considered.
You should skip dynamically generated classes from being processed, but you still need to dig further because there can be actual code coverage further down the stack trace.

Godin:
@pedrocborges2 JaCoCo 0.7.6 has configuration option "inclnolocationclasses" ( #288 ).

pedrocborges2:
Thank you @Godin I somehow missed that.

PR #296, opened 2015-03-09
Experimental: Instrumentation without adding members

marchof:
This is a experimental branch to avoid the class re-transformation problem caused by the additional members added by JaCoCo. This comes with the following costs:

- the runtime access is inlined in every method of instrumented classes
- the runtime is accessed with every method invocation

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactoryTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactoryTest.java
@@ -17,6 +17,7 @@ import static org.junit.Assert.assertNull;
 import org.jacoco.core.runtime.IExecutionDataAccessorGenerator;
 import org.jacoco.core.runtime.OfflineInstrumentationAccessGenerator;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
@@ -30,6 +31,7 @@ import org.objectweb.asm.Opcodes;
  * {@link IProbeArrayStrategy} implementations. The verifies the behaviour of
  * the returned {@link IProbeArrayStrategy} instances for different classes.
  */
+@Ignore("Factory logic temporarly disabled ")
 public class ProbeArrayStrategyFactoryTest {
 
 	private IExecutionDataAccessorGenerator generator;
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ClassFileVersionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ClassFileVersionsTest.java
@@ -32,6 +32,7 @@ import org.jacoco.core.JaCoCo;
 import org.jacoco.core.instr.Instrumenter;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.SystemPropertiesRuntime;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
@@ -69,16 +70,19 @@ public class ClassFileVersionsTest {
 	}
 
 	@Test
+	@Ignore("no frames required when accessing the runtime everytime")
 	public void test_1_6() throws IOException {
 		testVersion(V1_6, true);
 	}
 
 	@Test
+	@Ignore("no frames required when accessing the runtime everytime")
 	public void test_1_7() throws IOException {
 		testVersion(V1_7, true);
 	}
 
 	@Test
+	@Ignore("no frames required when accessing the runtime everytime")
 	public void test_1_8() throws IOException {
 		testVersion(V1_8, true);
 	}
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactory.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactory.java
@@ -15,7 +15,6 @@ import org.jacoco.core.internal.data.CRC64;
 import org.jacoco.core.internal.flow.ClassProbesAdapter;
 import org.jacoco.core.runtime.IExecutionDataAccessorGenerator;
 import org.objectweb.asm.ClassReader;
-import org.objectweb.asm.Opcodes;
 
 /**
  * Factory to find a suitable strategy to access the probe array for a given
@@ -40,35 +39,15 @@ public final class ProbeArrayStrategyFactory {
 			final IExecutionDataAccessorGenerator accessorGenerator) {
 
 		final String className = reader.getClassName();
-		final int version = getVersion(reader);
 		final long classId = CRC64.checksum(reader.b);
-		final boolean withFrames = version >= Opcodes.V1_6;
 
-		if (isInterface(reader)) {
-			final ProbeCounter counter = getProbeCounter(reader);
-			if (counter.getCount() == 0) {
-				return new NoneProbeArrayStrategy();
-			}
-			if (version >= Opcodes.V1_8 && counter.hasMethods()) {
-				return new FieldProbeArrayStrategy(className, classId,
-						withFrames, InstrSupport.DATAFIELD_INTF_ACC,
-						accessorGenerator);
-			} else {
-				return new LocalProbeArrayStrategy(className, classId,
-						counter.getCount(), accessorGenerator);
-			}
+		final ProbeCounter counter = getProbeCounter(reader);
+		if (counter.getCount() == 0) {
+			return new NoneProbeArrayStrategy();
 		} else {
-			return new FieldProbeArrayStrategy(className, classId, withFrames,
-					InstrSupport.DATAFIELD_ACC, accessorGenerator);
+			return new LocalProbeArrayStrategy(className, classId,
+					counter.getCount(), accessorGenerator);
 		}
-	}
-
-	private static boolean isInterface(final ClassReader reader) {
-		return (reader.getAccess() & Opcodes.ACC_INTERFACE) != 0;
-	}
-
-	private static int getVersion(final ClassReader reader) {
-		return reader.readShort(6);
 	}
 
 	private static ProbeCounter getProbeCounter(final ClassReader reader) {


Godin:
@marchof Cool! And I suppose that next step on this subject - is to do benchmarking. Might be tricky to define usual workloads to do this in isolation, so I propose to start with quick tests on real-life projects with addition of results here as comments. I'll start few such tests.

Godin:
@marchof And FYI seems that probe in JCov - is just a call to static method: http://hg.openjdk.java.net/code-tools/jcov/file/f42b2ac06163/src/classes/com/sun/tdk/jcov/instrument/Instrumenter.java#l42

marchof:
@Godin Unfortunately this gives creates a static dependency from the instrumented classes to the coverage library -- which does not work in most containers. JaCoCo uses different strategies to communicate to the agent runtime through JDK APIs only. With the exception of the "Offline" runtime which creates a static dependency on JaCoCo (requiring deploying jacocoagent.jar with your app in the runtime container specific way).

marchof:
@Godin BTW, we have a pluggable runtime access API even today. We could easily add a new implementation which uses a direct dependency.

marchof:
Apache Commons Collections (min test runtime out of 10 runs):

  * run 6066ms
  * coverage 7176ms (+18%)
  * coverage/nomembers 14429ms (+138%)

andrioli:
@Godin @marchof JCov works more like Cobertura. But I know that Cobertura has a high execution overhead.

Anyway, recently I executed JaCoCo in offline mode with 10 open-source tools. The goal was to compare the overhead of an experimental coverage tool that I was developing with another effective tool (JaCoCo).

I can re-execute this benchmark with this branch if you are interested.

Godin:
@marchof I know that static access not suitable for OSGi containers...

marchof:
@Godin Same problem with web and j2ee containers. The agent is only part of the container's application class path, not visible to the deployed apps itself.

With Java 1.6 the agent API got a new API to extend the bootstrap classpath. But you need to supply the location of a physical JAR file: https://docs.oracle.com/javase/6/docs/api/java/lang/instrument/Instrumentation.html#appendToBootstrapClassLoaderSearch%28java.util.jar.JarFile%29

Godin:
@marchof Do you know environments in which physical location of JAR file can't be determined? And as an idea - instrument classloaders, so that to inject JaCoCo runtime into each.

Godin:
SonarSource C++ Plugin, single shot, wall clock time:
* 1m13.069s = "mvn test"
* 1m35.259s (+30%) = "mvn org.jacoco:jacoco-maven-plugin:0.7.4.201502262128:prepare-agent test"
* 2m40.044s (+119%) = "mvn org.jacoco:jacoco-maven-plugin:0.7.5-SNAPSHOT:prepare-agent test"

marchof:
@Godin What I mean is that we would need a physical JAR file in addition to the agent where the location is known to the agent. Of course we could include that with jacocoagent.jar and export it to a temp location but this adds more complexity.

BTW, the runtime we're actually already using static access to a instrumented JRE class (ModifiedSystemClassRuntime).

Godin:
@marchof Yep I also came to ModifiedSystemClassRuntime in parallel with you :smile:

And so I don't believe that what we observe - is overhead due to call of static method. This needs to be checked, but: method is hot, so it will be inlined, eliminating cost of method call, and thus we observe something else what happens in this method.

marchof:
@Godin It's not just one method call, its a bunch of methods involving synchronization and a map lookup -- in comparison to local field access.

marchof:
BTW, it wouldn't be a big deal to make NNMM ("no new members mode") optional and configurable via a agent parameter.

Godin:
@marchof The main question in the story with bunch of methods - maybe they can be simplified? For example: maybe classId and name can be avoided in such calls using a kind of open addressing? And should be a way to avoid or at least minimize synchronization.

Godin:
For the more precise example: current access-call anyway done thru ``Object#equals`` with creation of  array and I'm really not sure about efficiency of this. So far this is just a speculation. Proper profiling and benchmarking will definitely help.

marchof:
That is tricky in case of ModifiedSystemClassRuntime: It would be easy to add a method to the modified system class, but this method cannot call any JaCoCo runtime class. That's why we use Object.equals(Object) as the "API".

Godin:
Let me dump another idea here - using ``invokedynamic`` possible to remove overhead of a second invocation of each probe turning it into NoOp. Support of reset for coverage data might tricky, but without it we did a PoC of a such coverage engine at SonarSource BrainDevDay.

marchof:
Same issue to solve: the bootstrap method (which is part of the agent runtime) has to be visible from the instrumented classes. Beside that this would disable reset it only works for Java runtime AND Java classfiles >= 1.7.

Godin:
@marchof yes, but IMO restriction on Java runtime isn't different from addition of an option for "no new members mode" :wink:

marchof:
@Godin Agreed, but the impact of changing the probe strategy at all is a bit bigger. Looks like we need a plugable framework to configure any aspect of the runtime implementation ;-)

But seriously: I don't have a good feeling in adding config options to adjust the internal implementation details and put it on the end user to find a suitable configuration for her/his scenario. I think the success of JaCoCo -- so far -- was that it works out of the box for most use cases.

Godin:
@marchof indeed

Godin:
@marchof while I was in process of writing benchmarks for different implementations of IRuntime, another idea came into my mind - what about generation of "companion" class for each instrumented class? This "companion" will simply hold field which we currently inject into instrumented classes.

marchof:
@Godin Interesting idea, the challenge is how to provide and load these classes.

Godin:
@marchof we know ClassLoader, so can call ``defineClass`` in it, no?

marchof:
@Godin Not sure whether this will "persist" the class in the classloader so it can be resolved from the instrumented class. Would need some investigation, but I really like the idea.

Godin:
@marchof And coming back to usage of ``Object#equals`` for communication with JaCoCo Runtime - here is another idea: why not inject Class with storage into each ClassLoader, so that it can be accessed directly by instrumented classes, JaCoCo runtime can hold list of such storages and visit them during operations such as dump and reset.

marchof:
@Godin Any idea how to inject a class into a classloader?

Godin:
@marchof something like this - https://github.com/raphw/byte-buddy/blob/master/byte-buddy-dep/src/main/java/net/bytebuddy/dynamic/loading/ClassLoaderByteArrayInjector.java#L119

Godin:
@marchof Choosing between last two ideas - I would prefer last one, since in this case we'll introduce one additional class per ClassLoader instead of per class. Do you want me to craft PoC of a new Runtime based on this PR?

marchof:
@Godin Ups, seems as simple as calling defineClass(). Than we could pick up the original idea to to create a companion jacoco runtime class for every instrumented class which holds the probe array as well as the jacocoInit() method. Nice, I'll give it a try.

Godin:
@marchof See my comment above - one additional class per instrumented class most likely will introduce bigger memory overhead than one additional class per classloader.

marchof:
@Godin The additional class shouldn't be a performance issue. The nice thing here that the runtime access could be done in a static initializer, so the instrumented class just needs to refer to a static member of the companion class. That is probably even faster that today's init() method calls.

Godin:
@marchof I'm concerned about memory and not about performance, i.e. PermGen in case of Java < 8 and MetaSpace in case >= 8.

marchof:
@Godin Some random remarks

  * This aspect is orthogonal to the different IRuntimeAccess implementations: The companion class still needs to deploy some magic to access the JaCoCo runtime (as it has the same class loader than the instrumented class).
  * The companion class makes the different IProbeArrayStrategy implementations obsolete.
  * We currently have a option to instrument bootstrap classes. Unfortunatelly you can't access their class loader.
  * For offline instrumentation companion classes must also be emitted.

Godin:
@marchof 
- Indeed communication magic will stay, but it won't be used for each probe - either during initialization, either during dump/reset.
- IMO case of instrumentation of bootstrap classes is rare one. And I don't see what prevents us from handling bootstrap classes differently than others.
- If I'm not mistaken, then in case of offline instrumentation direct access to runtime is already required.

marchof:
@Godin 

  * Would be nice to have a single instrumentation strategy only, just for simplicity.
  * Right, in case of offline instrumentation we have direct dependency (basically to start the JaCoCo runtime). But lookup is still expensive so we also cache probe array instances here in a new field

Godin:
@marchof Then I'm wondering - is there something what can prevent us from call ``defineClass`` for Bootstrap ClassLoader? :wink:

marchof:
@Godin Yep, there is no ClassLoader instance for the bootstrap classloader :-(

Godin:
@marchof ah, indeed, you right - parent of System ClassLoader is Extensions ClassLoader, not Bootstrap.

Godin:
@marchof ``-Xbootclasspath`` in addition to ``-javaagent`` and we'll be inside System ClassLoader? :smile:

marchof:
@Godin -Xbootclasspath specifies a static location for additional classes. We would need to add the companion classes on the fly as we instrument classes. Probably we can emit these classes to a file system  location specified with -Xbootclasspath ... too much of complexity.

Godin:
@marchof My point was - forget about addition of the classes on the fly: ``java -javaagent:jacoco.jar -XBootclasspath=jacoco.jar`` :smile:

andrioli:
Hey Guys, just FYI (slightly off topic)

I "tried"  execute the programs that I mentioned in the previous comment to analyze the cost of look up the probe array in each method invocation. I was using with the off-line instrumentation, so there NO cost to create an array to pass data through the `equals` method. Furthermore, I only instrumented the program classes (i.e., I do not instrumented test classes nor JAR dependencies).

But it was very slow, I had no patience to wait. Inline the probe array lookup seems unfeasible.

To give you an idea.

JTopas-0.8 (wall clock time)
 * JaCoCo 0.7.5 : 913.272 seconds
 * JaCoCo 0.7.4 : 62.023 seconds

I know that the class re-transformation problem is only related to on-the-fly instrumentation. So in the off-line instrumentation can still make a cache of the probe array in a class field :)

More interesting than the large difference of this branch and  version 0.7.4 is the difference between earlier versions of JaCoCo. When I compared the runtime overhead of JaCoCo with the experimental data-flow coverage tool that I was developing, I used JaCoCo version 0.7.1. But the JTopas 0.8 execution time was much higher than these 62 seconds. So I re-ran to be sure. 

Look this: (wall clock time, just one run)
 * JaCoCo 0.7.5 : 913.272 seconds
 * JaCoCo 0.7.4 : 62.023 seconds
 * JaCoCo 0.7.3 : 61.184 seconds
 * JaCoCo 0.7.2 : 156.816 seconds
 * JaCoCo 0.7.1 : 157.956 seconds

Looks that the instrumentation changes added in 0.7.3 modify some very sensitive things to JIT and JTopas. I'm just sharing my thoughts :)

Godin:
@andrioli Thank you for sharing this.

Godin:
@marchof Finally I wrote benchmark for various methods of accessing data (https://github.com/Godin/jacoco-benchmarks-playground). And here is raw data:
```
Java(TM) SE Runtime Environment, 1.7.0_76-b13
Java HotSpot(TM) 64-Bit Server VM, 24.76-b04
Linux, 3.17.8-gentoo-r1, amd64
8 available processors

Running with 1 threads:
        DirectAccess:   24.481 +-  0.230 ns/op
 ModifiedSystemClass:   24.805 +-  0.254 ns/op
                NoOp:    3.182 +-  0.063 ns/op
    SystemProperties:   54.189 +-  1.016 ns/op
    URLStreamHandler:  119.173 +-  0.774 ns/op

Running with 2 threads:
        DirectAccess:   27.836 +-  0.266 ns/op
 ModifiedSystemClass:  202.553 +- 21.028 ns/op
                NoOp:    3.299 +-  0.014 ns/op
    SystemProperties:  306.387 +- 64.820 ns/op
    URLStreamHandler:  180.719 +-  1.801 ns/op

Running with 4 threads:
        DirectAccess:   30.235 +-  7.357 ns/op
 ModifiedSystemClass:  397.545 +- 17.006 ns/op
                NoOp:    3.474 +-  0.204 ns/op
    SystemProperties:  586.242 +- 42.835 ns/op
    URLStreamHandler:  704.170 +-  9.153 ns/op
```
which should be investigated/explained, especially differences between single and multiple threads.
To check sanity of those measurements, we most probably can correlate them with overheads observed earlier for real projects.

Godin:
@marchof You'll find patch with PoC for companion classes in branch https://github.com/jacoco/jacoco/tree/companion-class . Patch is a bit dirty, because not formatted properly, contains some duplication with existing code and lacks unit tests for new code, but it passes existing tests and can be executed to gather coverage for JaCoCo itself! And speaking about performance - execution for SonarSource C++ Plugin (single shot, wall clock time): 1m35.885s.

Godin:
@marchof out of today's discussion I didn't fully get idea of how we can have companion class with many fields without knowing in advance number of probes for not-yet-instrumented classes. As was said - your PoC avoids counting of probes, so could you please publish it? Even if it is not complete and ready for integration, it will give nice food for my brain.

marchof:
> without knowing in advance number of probes for not-yet-instrumented classes

Indeed, I missed this one :-(

 My implementation actually does count probes -- but during instrumentation. For certain cases (static initializers < Java 8) we need to know the total number of probes before (https://github.com/jacoco/jacoco/blob/master/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactory.java#L47).

I will push my prototype.

Godin:
@marchof ok, but that maybe not a big issue - just inversion of control: if during instrumentation we can trigger initialization of field, then companion classes with many fields are possible. I suppose that this won't actually change time of initialization, and might be not a big deal even if will change. Since in this case mandatory to keep track of already-created companion classes per classloader, special care should be taken to not prevent GC for them.

Godin:
But actually before jumping into implementation we should answer on two important questions:
* do we really want to avoid addition of fields into instrumented classes?
* is it an problem, if we'll generate many additional classes?

marchof:
> do we really want to avoid addition of fields into instrumented classes?

I see two some cases for this:
* Better cope with other tools dynamically changing class files (mocking)
* Attaching agent to running JVM (and instrument classes already loaded)
* Instrument bootstrap classes which have been loaded before (e.g. to load JaCoCo agent itself)

Here is my try: https://github.com/jacoco/jacoco/tree/companion-class-2

Godin:
@marchof Euh, IMO under the hood they are not much different :laughing:

marchof:
@Godin Actually I did a first look just now and felt the same... good to know that there seems to be a common understanding how JaCoCo actually works :wink:

Godin:
note: initialization of probe array at a time of instrumentation is different from initialization at execution time, for example in cases when there is reference to class, but its members not used:
```
public class Example {
  public static final String FIELD = null;
  static {
    System.out.println("clinit");
  }
}

public class ExampleTest {
    System.out.println(Example.class);
    System.out.println(java.util.Arrays.toString(Example.class.getDeclaredFields()));
  }
}
```

marchof:
@Godin Speaking about initialization: If we create companion classes on the fly during instrumentation we could also directly load them and inject the corresponding probe array instance into the constants. This would obsolete all the IRuntime-Hacks and the $jacocoInit() methods.
I will test on my prototype branch.

Godin:
@marchof this is exactly what I meant speaking about "initialization of probe array at a time of instrumentation".

marchof:
@Godin Sometimes it takes me some time to understand things :wink:

Godin:
@marchof if you don't have time to try, then I'll do this - we already know that our vision / approaches aligned ;)

marchof:
@Godin Too late ;-) I just pushed it to companion-class-2. It actually works, needs API cleanup though. Now we need to find a cool solution for bootstrap classes.

Godin:
For the record: Seems that this will be really useful for interoperability with PowerMock, which when asked mocks **all** static methods including synthetic one generated by JaCoCo - MEZk/checkstyle#1 .However for me not clear why syntetic methods are not ignored by  PowerMock in general.

PR #310, opened 2015-05-18
Provide better coverage data in case of implicit exceptions

marchof:
This is an alternative approach to the PR proposed in #261. The objective is to improve covered result in case implicit exceptions are thrown by method invocations. Instead of adding a extra probe for every method invocation this PR adds probes only between source lines -- and only if the subsequent line contains at least one method invocation.

This reduces the number of extra probes needed and at the same time provides better results in case of implicit exceptions in multi-line blocks.


Godin:
For comparison

this PR:
pr-310

PR #261:
pr-261

marchof:
@Godin Nice, thx! Another useful comparison would be the total number of probes and the execution duration for a reasonable test suite.

mackerl:
I plan a testrun with this pull request this weekend. Afterwards I will give you a comparison about runtime and coverage results between the two PRs

mackerl:
Testsuite with 6000 JUnit Tests, 3 Million LoC under Test, 48% Coverage
without Coverage: 12 hours
with Coverage from Master without PR #261 29 hours
with Coverage from Master with PR #261 30 hours
with Coverage from Master with PR #310 coming soon

marchof:
@mackerl Wow, 12h for 6000 Unit-Tests? Are this actually *unit* tests? Or do you fork a VM for each test case? This would also explain why the runtime with coverage more than doubles. Jsu ta comparison: JaCoCo runs 1000 Tests with coverage in 80sec:

http://www.eclemma.org/jacoco/trunk/test/index.html

Anyways, many thanks for your ongoing feedback, very helpful!

marchof:
Here is a build for this PR: http://download.eclipselab.org/jacoco/preview/jacoco-0.7.5.201505190939.zip

mackerl:
@marchof You are welcome. A few hundred of the tests are more on the API Layer. They have a runtime from a few minutes up to one hour. We also do not mock everything - there are actually around 1000 tests involving a real Oracle or SqlServer Database, where a specific datastage is set up during the bootstrap of the Unit Test Framework. Those tests typically run up to a few minutes. Nevertheless the biggest part of our testsuite are real unit tests with a runtime of around a few milli seconds.
Thanks for the build ....  would it be an easy task for you to provide an eclemma update site, where this change in jacoco is included? Otherwise I can also do it by myself

marchof:
@mackerl Sorry, my local build infrastructure currently only allows to build EclEmma from Maven repo. Could work on this but not on short notice.

mackerl:
@marchof no problem ... i will hack maven myself again then :)

marchof:
@mackerl What counter does your coverage percentage refer to? Would be nice to know the percentages for onstructions, branches and lines separately.

marchof:
I will create a comparison for commons-collections tests tonight, camparing the following figures:

* execution time
* covered and total instructions
* covered and total branches
* covered and total lines
* total number of probes
* exec file size

Godin:
@marchof I'm wondering if possible to update your PR to display lines with exceptions as partially covered as it was done in PR #261 ? IMO this makes report in such cases more readable.

andrioli:
+1 for @Godin proposal. IMO the partially covered lines make the report more "understandable".

Godin:
Indeed - I meant "understandable", when wrote "readable" :wink: Thx for correction @andrioli

Godin:
@marchof out of [InstrumentationBenchmark](https://github.com/Godin/jacoco-benchmarks-playground/blob/master/src/main/java/org/jacoco/benchmarks/InstrumentationBenchmark.java) - comparison of "cold" performance of instrumentation of jar, supposed that offline instrumentation not much different from online ( 10 forks, 1 iteration, no warmup, JDK7u80 ) :

elasticsearch = org.elasticsearch:elasticsearch:1.1.2 , 13M
guava = com.google.guava:guava:18.0 , 2.2M
rt = rt.jar of used JVM , 60M

master:
```
  elasticsearch: 4372.632 ± 142.458 ms/op
          guava: 1763.230 ±  33.403 ms/op
             rt: 8443.019 ± 257.801 ms/op
```

this PR:
```
  elasticsearch: 4373.558 ± 106.255 ms/op
          guava: 1815.460 ±  77.315 ms/op
             rt: 8574.634 ± 435.716 ms/op
```

PR #261:
```
  elasticsearch: 4366.380 ± 167.617 ms/op
          guava: 1822.976 ± 141.603 ms/op
             rt: 8502.746 ± 291.181 ms/op
```

Note that benchmark can be improved to exclude time of jar reading, but IMO it already shows that we can forget about impact of changes on instrumentation time for both PRs.

Godin:
@marchof and btw I believe that file size computable from number of probes, so why you plan to track it?

marchof:
@Godin exec files contains meta information (class name, id) and probes. As probes are stored as bits they make only 50%. But doesn't actually matter, I'll post my results in a few minutes.

marchof:
Here are the results for commons-collections (14,594 test cases):

Branch | Runtime | Exec File Size | Probes | Instructions      | Branches      | Lines
------ | ------- | -------------- | ------ | ----------------- | ------------- | ---------------
Run    | 5.919s  |                |        |                   |               |
Master | 7.222s  | 58,182         | 15,988 | 109,092 / 130,209 | 6,005 / 8,364 | 25,804 / 30,841
PR 310 | 7.421s  | 59,944         | 29,851 | 109,455 / 130,209 | 6,004 / 8,364 | 25,869 / 30,841
PR 261 | 7.571s  | 62,570         | 50,086 | 111,475 / 130,209 | 6,010 / 8,364 | 26,460 / 30,841

While we see a significant increase in probes between the different approaches, runtime seems not be really effected. Also memory overhead should be minimal as we need 1 byte per probe. As a conclusion probes are pretty cheap and we can actually heavily use them!

Godin:
@marchof SonarSource C++ Plugin, modular project but all classes included in instrumentation and so in dump files , 48K of source code and 1K of classes not counting tests , 2400 tests and 92.7% coverage as reported by SonarQube with JaCoCo 0.7.4 , single shot, wall clock time, first measurement - "mvn test", subsequent - "mvn jacoco:prepare-agent test" :

without coverage = 1m18.3s
master = 1m36.9s
this PR = 1m36.8s
PR #261 = 1m37.1s

We don't have aggregated JaCoCo reports, so can't quickly provide all other numbers, but can work on this if needed.

Taking into account run-to-run variance I consider that all 3 branches actually demonstrate same time, so changes are insignificant.

Godin:
@marchof even, if addition of probes has insignificant impact on performance, we anyway should be careful at least due to limits on size of methods :smiling_imp: and what about impact on size of class files on disk and in memory, and impact on size of compiled code in memory? :laughing:

Godin:
@marchof did you noticed difference in covered branches and lines in results for commons-collections? I just wanted to make a note on this measurement - don't exclude possible existence of run-to-run variance, I fighted a lot to get rid of it in project mentioned before.

marchof:
@Godin yes, there are differences, see my table above. The columns Instructions, Branches and Lines show covered vs. total items each.

marchof:
Regarding coverage status of methods with exceptions: If we spent a probe per method invocation we can even mark those lines as "green".

BTW, there are a couple of other implicit exceptions (NPE, IndexOutOfBounds etc.) which will still not result in coverage.

Godin:
@marchof my point was that maybe you'll see difference even between multiple runs of exactly the same JaCoCo build

marchof:
@Godin I just tried with master: The figures seem to be reproducible.

Godin:
@marchof not sure that I understand your point about "green" lines. Original proposal was to show lines 12, 27, 36, 44, 48 as partially covered (see screenshots). You propose to show them as green aka fully covered? If so, then understandably is the same as in case of red aka not-covered - let's take line 12 as example with slightly changed names of methods:
```
a(); // green
b(); // if green or red, then where exception happened? in method "b" or in method "c"?
     // and in case of yellow it is clear that exception happened in method "b".
c(); // red
```

Godin:
@marchof ok, so then I suppose that figures are different exactly because of difference of branches on not-covered and partially-covered.

marchof:
@Godin To be honest I have know idea yet why the lines becomes yellow in #261. Currently we have no semantics in our coverage results for "exception happens here". Every instruction is executed or not. A yellow line simply means that some instructions are covered, others not. A static void method call is a single instruction -- so yellow shouldn't be possible.

Godin:
@marchof I think that explanation is here - https://github.com/jacoco/jacoco/pull/261/files#diff-3e9be84b8e6db90bab251bbb404fef07R294

marchof:
@Godin A, now I see: your methods nop() and ex() have a return type. Change them to void and the lines probably becomes green.

Godin:
@marchof 
selection_004

So I still think that this is because of specific changes in PR #261 - see my previous comment.

Godin:
@marchof However you right that there is difference in bytecode:
```
  static void implicit();
    flags: ACC_STATIC
    Code:
         0: invokestatic  #2                  // Method nop:()Ljava/lang/Object;
         3: pop           
         4: invokestatic  #5                  // Method ex:()Ljava/lang/RuntimeException;
         7: pop           
         8: invokestatic  #2                  // Method nop:()Ljava/lang/Object;
        11: pop           
        12: return        
      LineNumberTable:
        line 11: 0
        line 12: 4
        line 13: 8
        line 14: 12

  static void implicitNoResult();
    flags: ACC_STATIC
    Code:
      stack=0, locals=0, args_size=0
         0: invokestatic  #6                  // Method nopNoResult:()V
         3: invokestatic  #7                  // Method exNoResult:()V
         6: invokestatic  #6                  // Method nopNoResult:()V
         9: return        
      LineNumberTable:
        line 17: 0
        line 18: 3
        line 19: 6
        line 20: 9
```

marchof:
@Godin That's why the lines with exceptions are red now. But if we spend a a extra probe for each method invocation we probably could introduce the exception semantic to our coverage data.

Godin:
@marchof now I see two scenarios, while originally I had in mind only first one:
1. in case if all tests touch same code and endup in exception thrown from method (or more realistic - execution of single test from IDE), then for me seems that there is no need for additional probe - fact of exception can be derived by looking on subsequent probe
2. in case of multiple tests, where some touch exception and nother not, then indeed I see what you mean by "exception semantic" - even if line fully covered might be interesting to know that some tests were stopped in the middle due to exception

Godin:
@marchof and so then current behaviour correct and logical, whereas showing of line as green aka fully-covered IMO will be not really correct, thus probably would be better to go ahead with this PR as it is now and to consider introduction of "exception semantic" as separate enhancement additional to this.

marchof:
@Godin Fine to me. May I ask you to review the code changes?

org.jacoco.core/src/org/jacoco/core/internal/flow/LabelInfo.java
```diff
@@ -105,6 +107,43 @@ public static boolean isSuccessor(final Label label) {
...
+		final LabelInfo info = get(label);
+		return info == null ? false : info.methodInvocationLine;
+	}
+
+	/**
+	 * Determines whether the given label needs a probe to be inserted before.
+	 * 
+	 * @param label
+	 *            label to test
+	 * @return <code>true</code> id a probe should be inserted before
```
Godin: typo: "id" -> "if"

org.jacoco.core.test/src/org/jacoco/core/internal/flow/LabelInfoTest.java
```diff
@@ -90,6 +91,42 @@ public void testMultiTarget3() {
...
+	@Test
+	public void testNeedsProbe() {
+		testNeedsProbe(false, false, false, false);
+		testNeedsProbe(true, false, false, false);
+		testNeedsProbe(false, true, false, false);
+		testNeedsProbe(true, true, false, false);
+		testNeedsProbe(false, false, true, false);
+		testNeedsProbe(true, false, true, true);
+		testNeedsProbe(false, true, true, true);
+		testNeedsProbe(true, true, true, true);
```
Godin: not sure that such style of tests is readable, but that's minor

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -36,6 +38,12 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/301">#301</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+  <li>The exec file version has been updated and is not compatible to previous
+      versions.</li>
```
Godin: typo "to" -> "with" ?

org.jacoco.ant.test/src/org/jacoco/ant/CreateExecFiles.java
```diff
@@ -0,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2015 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Brock Janiczak - initial API and implementation
```
Godin: why?

org.jacoco.doc/docroot/doc/faq.html
```diff
@@ -45,13 +45,13 @@
...
-<h3>Code with exceptions shows no coverage. Why?</h3>
+<h3>Source code lines with exceptions shows no coverage. Why?</h3>
 <p>
   JaCoCo determines code execution with so called probes. Probes are inserted
   into the control flow at certain positions. Code is considered as executed
   when a subsequent probe has been executed. In case of exceptions such a
-  sequence of instructions is aborted somewhere in the middle and not marked as
-  executed.
+  sequence of instructions is aborted somewhere in the middle and the
+  corresponding source code lines is not marked as executed.
```
Godin: I feel that this should be refined - after the change I still get impression that exceptions are completely not covered by JaCoCo.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/LabelInfoTest.java
marchof: Any other proposal? Alternatively I can create scenario tests for this (as already used in this test class).

org.jacoco.core/src/org/jacoco/core/internal/flow/LabelInfo.java
marchof: Done.

org.jacoco.doc/docroot/doc/changes.html
marchof: Done.

org.jacoco.ant.test/src/org/jacoco/ant/CreateExecFiles.java
marchof: Copy & paste... done.

org.jacoco.doc/docroot/doc/faq.html
marchof: Right, implicit exceptions are not covered by JaCoCo. But with this change on line level, not on block level any more. That's why I adjusted the FAQ entry from blocks to lines.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/LabelInfoTest.java
Godin: @marchof fine to keep it as is.

marchof:
@Godin Anything that prevents us from merging this?

Godin:
@marchof nothing! you can go ahead or you wanna me to do squash/merge?

scottcarey:
The new coverage form does not make sense to me.  I understand why it is happening, but if you call

ex();  

which always throws.

Then it is always thus 'red' at the call site, yet the inside of its method is green!  This is inconsistent.  How could you have coverage inside ex() if you never called it?  That is what is implied, to me, by ex() being red.  Execution never got there, and stopped the line before.   

Hopefully the new 'exception semantic' fixes this, because yellow made way more sense to me here -- we entered, but did not leave, ex().

edisonh:
+1 for @scottcarey proposal.
Before introduction of "exception semantic", yellow is more semantic.

C-Otto:
+1 from me. I still prefer my original PR #261 for what it's worth.

C-Otto:
Did you realize there are red squiggly lines all over the place? These make the text unreadable.

YogiKhan:
sorry my bad!!!
Added new:
image

C-Otto:
The `@Transactional` annotation may be involved here. Could you please open an issue instead of commenting on a pull request?

Godin:
@YogiKhan @C-Otto I'm removing comments here since new issue has been opened

PR #317, opened 2015-05-28
Character range table attribute

Godin:
Proof of concept of usage CharacterRangeTable attribute for mapping of bytecode branches back to conditions in source code.

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -409,6 +409,18 @@
     </pluginManagement>
 
     <plugins>
+      <!-- FIXME(Godin): experimental: -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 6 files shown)


Godin:
@marchof Could you take a look please? Some notes:
* I'm not confident in change in MethodSanitizer, but without it label objects not the same at time of reading of attribute and at time of analysis.
* I don't know (yet?) how to integrate this into reports. But message is displayed in console for each uncovered branch as proof of concept. You can observe this on build of JaCoCo itself.
* During initial testing I found that not all branches can be found in attribute, but this is IMO not a problem. For one of examples - "for-each" loop.

marchof:
@Godin How cool is that: I wasn't aware that ASM is extensible for custom attributes. I will have a closer look how this could be integrated in ICoverageNodes and finally reports.

zhaoweiqiii:
now how can we match which branchs was missed?

PR #318, opened 2015-05-28
ClassCoverageImpl should not be part of API

Godin:
I don't see any good reasons for this class to be part of API:
* it is used only by other internal classes and tests
* and IMO not supposed to be used directly outside of JaCoCo - there is public interfaces
* was introduced in public package in first JaCoCo version, but I suppose that this was done by mistake

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/analysis/CounterComparatorTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/analysis/CounterComparatorTest.java
@@ -19,6 +19,7 @@ import java.util.Comparator;
 
 import org.jacoco.core.analysis.ICoverageNode.CounterEntity;
 import org.jacoco.core.internal.analysis.CounterImpl;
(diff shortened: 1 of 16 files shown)


Godin:
@marchof could you verify please? Notes:
* No changes for OSGi metadata were done, as all places of use already should have access to the new location of this class, but this wasn't verified by execution in OSGi container, only manually reviewed.
* Not sure that changelog item is good enough.

Godin:
And BTW thanks to @benzonico for pointing this out.

marchof:
@Godin Have you noticed that the class is used within org.jacoco.report bundle? This will fail in OSGi containers, e.g. the EclEmma plugin. Also as an API user if you want to create aggregate nodes this is a quite useful class.

Godin:
@marchof indeed, you're right - I missed usage in AbstractGroupVisitor, the use for aggregate nodes also not obvious - maybe add a comment. In any case I'm closing this ticket as "invalid".

Issue #219, opened 2014-06-12
Exclude annotation


rlogiacco:
If an exclude annotation with RUNTIME retention policy is added users could easily determine methods to skip in code coverage, an example being generated code/classes.
I'm willing to contribute a patch if the approach is considered valid.

PR #330, opened 2015-06-27
For format version mismatches, report version number in a human friendly way

somechris:
Version numbers have been reported as hex string, but without any hex
string marker. So effectively when seeing a mismatch reported around
version '1006', it came down to version '0x1006', which is version
4102 in decimal. Those high format version numbers and mismatch in hex
vs. decimal was confusing.
Hence, we now get rid of the byte-marker 0x1000, and report the format
version as plain decimal.
Thereby, nice low format versions get reported. And the confusion
between (markerless) hex and decimal is gone.

What is 0x1007 format version on-disk, is now reported as version 7.
What is 0x1006 format version on-disk, is now reported as version 6.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
@@ -113,7 +113,7 @@ public class ExecutionDataReaderWriterTest {
 		assertEquals(0x01, 0xFF & header[0]);
 		assertEquals(0xC0, 0xFF & header[1]);
 		assertEquals(0xC0, 0xFF & header[2]);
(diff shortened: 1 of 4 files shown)


Godin:
IMO this relates to #319

Godin:
@somechris I didn't get why "6" or "7" is more user-friendly and less misleading than any other numbers like "1006" and "1007"? In both cases remediation action is unknown, when user receives such internal version format number, which is IMO root of the problem.

somechris:
Thanks for the pointer. You're right, they are definitely related.
I'll respond on #319 to your questions, so all the discussion is in one place.

marchof:
@somechris In a separate PR #219 I fixed the error message and created a specialized IOException. Though the decission that we write version numbers as hex strings and started with version 0x1001 I decided not to hide this fact in the error message. But version numbers are now properly written as hex literals (0x1007).

Thankf for your support!

PR #333, opened 2015-09-30
Fix MBeanClient example

marchof:
* JavaDoc is outdated - there is no "mbean" output mode
* IProxy:
 * method ```byte[] dump(boolean reset)``` should match corresponding method from IAgent, which is ```void dump(boolean reset)```
 * method ``` byte[] getExecutionData(boolean reset)``` should be added
 * to avoid such issues consider execution of example as part of build


marchof:
@Godin Fixed this and added a JUnit test to ensure both APIs are in sync.

PR #336, opened 2015-08-14
added support for multi-module projects

prgp:

PR #349, opened 2015-10-02
jacoco-maven-plugin options for report generation plus external dependencies

cristcost:
Hello,
I propose this pull request that provides improved configuration of the jacoco-maven-plugin and capability to include external dependencies in reports. 

## Rationale
I want to find "dead" code and how much external dependencies are used. This is because I want to keep the project clean by minimizing external dependencies. Often, we import an external library which takes many transitive dependencies which are not used, other time a big library is imported for using just a single static utility method. I want to used JaCoCO to detect these situations, but I need coverage report for these external dependencies.

This pull request is different from #336 as I don't focus on reporting other projects from the reactor project; I rather focus on being able to include any external Jar from the dependencies of the project, more similarly to how it is possible to do in EclEmma when you import external session and it allows you to "include binary libraries", see picture:

schermata 2015-10-02 alle 09 51 05

The implementation has been inspired from the implementation of the Report Ant task.

## Usage
The plugin can the be configured in this way:
```
<configuration>
	<!-- If you don't want cvs report set it false -->
	<csvReport>false</csvReport>
	<!-- If you don't want xml report set it false -->
	<xmlReport>false</xmlReport>
	<!-- 
		The htmlReport options to true will generate the html report.
		It can be omitted as all reports are true by default to keep 
		comaptibility with previous behaviour.
	-->
	<htmlReport>true</htmlReport>
	<!-- 
		If you want coverage report also for test sources set this to true
	-->
	<includeTests>true</includeTests>

	<!-- 
		If your project for example depends, directly or indirectly, on Google 
		Guava, Apache Common-IO and Jetty, and you want to inspect how much you make use 
		of those libraries, include them in the report like this:
	-->
	<includeArtifacts>
		<!-- format is [groupId]:artifactId -->
		<includeArtifact>com.google.guava:guava</includeArtifact>
		<includeArtifact>org.apache.commons:commons-io</includeArtifact>
		<!-- groupId for org.eclipse.jetty can be omitted if artifactId is unambiguous -->		
		<includeArtifact>:jetty-server</includeArtifact>
	</includeArtifacts>
</configuration>
```

### Mojo Configuration
New configuration options have been included in AbstractReportMojo.java:
```
boolean csvReport;
boolean xmlReport;
boolean htmlReport;

boolean includeTests

List<String> includeArtifacts;
```
### Grouped Reporting
The method ```void createReport(final IReportGroupVisitor visitor)```  has been changed to manage reporting of the dependencies in different report bundles. This will only happen in case the condition ```(includeTests || (includeArtifacts != null && includeArtifacts.size() > 0))``` is true, else the report will not be grouped, so to keep the output and behavior consistent with previous version.

### Report of a Dependency Artifact
The method ```private void createDependencyReport(final IReportGroupVisitor visitor, final Artifact artifact)``` has been added to manage the creation of a report bundle for a Maven Artifact.

### Selection of type of Reports
The method ```IReportVisitor createVisitor(final Locale locale)``` has been changed in order to add visitors for specific report types only if the relative configuration option is set to true

### Sources in src/test/java
The method ```List<File> getCompileSourceRoots()``` has been changed to search sources also in src/test/java, as this is required if the option includeTests is set to true.

### Coverage of dependency Artifact
The class ```ArtifactBundleCreator``` has been created in order to manage the coverage of a dependency Artifact. 

*Note*: it is very similar to class ```BundleCreator``` and some refactoring to reduce duplicated code is advised in future.

### Report bundle name in BundleCreator
The class ```BundleCreator``` has been modified in order to take an external name for the report bundle. This is because the BundleCreator is used bot for report on src/main/java and src/test/java, and in this case we want to specify two different names and not using for both ```this.project.getName())```

### Locating sources of external artifacts
The class ```MavenSourcesArtifactLocator``` has been created in order to resolve sources in maven sources jar, using the standard maven convention that the sources are are located in the jar with classifier sources (i.e. myproject-1.0.0-sources.jar will contain sources for myproject-1.0.0.jar)

*Note*: sources are not downloaded if not available in local repo. User must invoke manually the ```mvn dependency:sources``` or use other solutions to download sources.

### Configured Mojo to require dependecies
The mojo option ```@requiresDependencyResolution runtime``` has been added to ReportMojo.java and ReportITMojo.java in order to make it work the method ```getProject().getArtifacts()``` that it is used to find the Jar files in the local repo and use them for report generation (this was an hard issue for me to find and solve :-D ) 

*Note*: Probably it is better to use ```@requiresDependencyResolution test``` to allow resolution also of test dependencies?

## Final notes
Please consider this pull request to become part of the standard distribution of JaCoCo. If you need some refinement such as code quality or code style improvement or other features, please ask and I'll be glad to do my best to improve this pull request and meet your needs.

I didn't implemented any Integration Test, I've developed and tested the patch on a real project of mine and I've never written IT tests for a maven plugin before. I could develop some IT, but please give me suggestion on what to do, what code to use as a template or which existing IT test to modify.

If some comment is not clear, please ask and I'll improve it.

Also, while working on the code, some automatic code format command given while editing the code in Eclipse may have changed the format (indentation and new lines). I've tried to keep the format the same as much as possible, but I have committed some format change anyway. Sorry as I know this makes it harder to compare the diff.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
@@ -303,8 +303,8 @@ public abstract class AbstractReportMojo extends AbstractMavenReport {
 					throws IOException {
 		final FileFilter fileFilter = new FileFilter(this.getIncludes(),
 				this.getExcludes());
-		final BundleCreator creator = new BundleCreator(this.getProject(),
-				fileFilter, getLog(), reportName);
+		final BundleCreator creator = new BundleCreator(fileFilter,
+				getLog(), reportName);
 
 		final IBundleCoverage bundle = creator
 				.createBundleOfDirectory(executionDataStore, directory);
--- a/jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
@@ -11,16 +11,6 @@
  *******************************************************************************/
 package org.jacoco.maven;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
-
 import org.apache.maven.doxia.siterenderer.Renderer;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.project.MavenProject;
@@ -40,6 +30,16 @@ import org.jacoco.report.csv.CSVFormatter;
 import org.jacoco.report.html.HTMLFormatter;
 import org.jacoco.report.xml.XMLFormatter;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+
 /**
  * Base class for creating a code coverage report for tests of a single project
  * in multiple formats (HTML, XML, and CSV).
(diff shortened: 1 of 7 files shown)


marchof:
@cristcost As already described on the mailinglist please help us to make PRs more consumable:

1) Do not mix different aspects in a single PR (here: report format selection vs. coverage of dependencies)
2) Include tests (here: no tests at all)

cristcost:
Thank you for the feedback @marchof ,
could you please give me some suggestion on what tests to implement? what do you expect?

Ok for not mixing PRs, I'll remove the report format selection for the moment and submit it as a different PR later.

Finally, when do you expect to make next release (I mean how much time approximately I have for fixing  this PR and be able to have it included in next release?)
Thanks
Cristiano

marchof:
This feature should become part of the new `report-aggregate` goal, see #388.

I'll opened a separate request #404.

Issue #14, opened 2012-08-29
Investigate filtering with annotations


marchof:
**brock_j@SF**

 It may be useful to tag a method/class as unable to be covered. Elements that are tagged with this annotation will not be instrumented and will be skipped during structure analysis.

The annotation would have to have a retention of 'class' so it doesn't have to be present at runtime. The annotation could eventually be replaced by one provided by the JDK (if one is ever added)

It might be a good idea to actually mark the element as explicitly 'skipped' in reports so the total element counts are accurate.

**marchof**

We should put this on the list of filtering szenarios. Some concerns:
- If the annotation is JaCoCo API, we get a dependency from production code to a particular code tool :-(
- When we have a reasonable set of filtering patterns (e.g. for private, empty default constructors) the annotation requirement might become obsolete. 

**brock_j@SF**

You are correct about the dependency issue. It is a compile time only link, which is slightly better, but probably still not acceptable.

How about something a little bit different?

We don't need to filter out members during instrumentation, instead, it could be done at the reporting/view stage. We could simple provide a file with instructions on what is to be skipped in the report. This would probably fit better with your idea of filtering options. This would just be a 'static' filtering rule (ignore the method called x in class org.example.Foo)

Eclipse integration could mark the ignored elements in some way (perhaps a special annotation of background colour)

**brock_j@SF**

How common would it be to not want to record coverage information for test classes? Could we look for classes that have test methods defined in them?

Or might we in future want to instrument test classes differently to put extra information into the execution data stream?

**marchof**

Test cases itself should not be in the scope of the coverage analysis. But as the test implementation typically has a different class path entry it can be easily ignored by the coverage analysis (see JaCoCo coverage report).

What could be useful indeed is a annotation, that defines the scope of a test case (something like @TestScope(MyClass.class). When the coverage runtime encounters this annotation e.g. on a method it would track execution information only for the given class(es). This would result in precise 1:1 test coverage data, filtering out all execution data that results from side effects from other tests. Actually for the JaCoCo build I already implemented this on bundle level by specifying an instrumentation filter that for the classes in the bundle under test only (so e.g. report tests do not produce any execution data for the core bundle, even though core classes are heavily used by the tests).

marcpawl:
[...]

PR #351, opened 2015-10-07
Annotation to mark method as fully covered in reports

marcpawl:
Issue #14 
While annotations bring in licensing problems for some commercial projects, by dual licensing the annotations with EPL or Apache, most closed source projects should be able to include the annotations.  Note only the annotations are dual licensed, the remainder of the code of course has not had its licensing changed.

An annotation has been added to indicate that all instructions in a method have been executed.

The use case to handle is:  marking as covered generated code that is trusted, even though not all possible branches.  Even though at the class level a report will show that the method has been fully executed, you can drill down and see what has not been covered. 

A soon to come feature is an other annotation that will mark the code for a method as executed only if at least part the code is executed.

While this solution does not solve all the problems that are desired, it makes progress and solves a highly desired use case.

The use case I have is that I trust generated code from Eclipse for toString(), hashCode(), and equals(Object), even if fully executed.  And due to the null checks in the code not all branches may be covered since it violates the objects integrity constraints; however it is not desirable to modify the code to remove these changes. Note that a lot of the time these three messages  are only be generated for debugging purposes.

Note that even though in a report a method will be marked as fully covered, if you drill down you can still see what was missed.  A nice to have would be to have some indication in the reports beside the method name to indicate that the coverage numbers have been messaged.


seanf:
Thanks for this work!

Regarding the annotation licence/dependency problem, Cobertura lets you specify the classname of any annotation, and then uses that annotation. Having a similar feature would make it very easy to switch from Cobertura to JaCoCo.

eg see this partial config for cobertura-maven-plugin:

            <instrumentation>
              <ignoreMethodAnnotations>
                <ignoreMethodAnnotation>org.zanata.util.CoverageIgnore</ignoreMethodAnnotation>
              </ignoreMethodAnnotations>
              <ignoreTrivial>true</ignoreTrivial>
            </instrumentation>

https://github.com/cobertura/cobertura/wiki/Coverage-Annotations

Issue #358, opened 2015-11-10
AgentOptions fails to parse options where destfile contains a comma


jochenberger:
```
Caused by: java.lang.IllegalArgumentException: Invalid agent option syntax "destfile=build/jacoco/foo, bar.exec,append=true,dumponexit=true,output=file,jmx=false".
	at org.jacoco.agent.rt.internal_932a715.core.runtime.AgentOptions.<init>(AgentOptions.java:192)
	at org.jacoco.agent.rt.internal_932a715.PreMain.premain(PreMain.java:43)

**Comment by marchof** [comment]
@jochenberger Well, this is a limitation of the current agent options syntax. Any ideas for a syntax extension to properly handle such cases?

**Comment by jochenberger** [comment]
If your keys are always lowercase letters, I suggest a RegExp-based approach
```groovy
String optionString = "destfile=build/jacoco/foo, bar.exec,append=true,dumponexit=true,output=file,jmx=false"
Pattern p = ~",(?=[a-z]+=)"
String [] items = p.split(optionString)
assert items == ["destfile=build/jacoco/foo, bar.exec", "append=true", "dumponexit=true", "output=file", "jmx=false"]
```

bjkail:
I'm not convinced it's worth the effort, but you could check if the first key is escaped=true and then parse the option string "manually" (without split) so that '\' or whatever can be used as an escape char.

marchof:
The agent parameter format has its limitations. Also file names with = characters will fail to parse.

Also built-in agents fail to parse such file names, e.g.

    > java -Xrunhprof:file=a,b.txt Example
    HPROF ERROR: hprof option error: Unknown option: b.txt (file=a,b.txt) [hprof_init.c:495]

Therefore I tempted to close this as "wontfix". If we have a compelling use case I would rather suggest to offer an alternative way to configure the agent (e.g. property file) which does not have such syntax limitations. For offline instrumentation this is already supported.

jochenberger:
My suggested fix also works for `=` in the file name.

Godin:
I also wonder about strong use case for comma in filename.

jochenberger:
Well, I have a directory that contains a comma. And I wonder why jacoco should explicitly forbid that.

marchof:
Thanks for contributing this fix! I just merged #359.

Godin:
@marchof I'm reopening this ticket, because I had hard time debugging patch for Gradle for new configuration option and this change was the root cause:
```
[...]

PR #359, opened 2015-11-12
Test and fix for #358

jochenberger:


org.jacoco.core.test/src/org/jacoco/core/runtime/AgentOptionsTest.java
```diff
@@ -435,5 +435,13 @@ public void testPrependVMArgumentsReplace() {
 				String.format("-javaagent:%s= a b c",
 						defaultAgentJarFile.toString()), vmArgument);
 	}
+	
+	@Test // issue #358
+	public void testDestFileWithComma() {
+		AgentOptions options = new AgentOptions();
```
mkordas: This line should be removed

PR #361, opened 2015-11-23
Don't count class stats twice for a package

### LINKED ISSUE #360: Incorrect package stats for duplicate class binaries
If duplicate classes with debug source file information are analyzed, the stats for the class are counted twice for the package.

I'll send a pull request shortly.

bjkail:
Fixes #360.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/analysis/CoverageBuilderTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/analysis/CoverageBuilderTest.java
@@ -165,6 +165,27 @@ public class CoverageBuilderTest {
 	}
 
 	@Test
+	public void testCreateSourceFileDuplicateClassNameIdentical() {
+		final MethodCoverageImpl method1 = new MethodCoverageImpl("doit",
+				"()V", null);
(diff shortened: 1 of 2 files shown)


marchof:
Nice catch, thanks!

Just out of curiosity: How did you discover this? What was you use case?

There still is a little flaw: If one would supply different coverage data for the same class the second event would be silently ignored. But I don't think this is a likely use-case here.

bjkail:
(Sorry for the duplicate issue, I'm new to github.)

My use-case is aggregating the execution data for the tests of many components of a large product in order to produce a full analysis of the entire product.  For various reason, classes from some components are repackaged multiple times across the product, so Analyzer.analyzeAll finds the same class multiple times.  We noticed by chance that the numbers weren't adding up in those cases.

The issue with duplicate IClassCoverage is somewhat pre-existing due to the classes.put, but you're right that this patch would not handle that consistently: SourceFileCoverageImpl would contain the data from the first IClassCoverage, but CoverageBuilder.classes would contain latter IClassCoverage.  Perhaps SourceFileCoverageImpl could be recomputed if a replacement IClassCoverage is added, or perhaps it's unimportant since the primary use-case is Analyzer, which will always recompute the same values anyway.

PR #365, opened 2015-12-08
Added coverage export to MongoDB

ihr:
Hi guys,

I've added an export to MongoDB in order to facilitate (microservice) coverage report generation/exportation when app is deployed in a Cloud.

Please, let me know what  you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,97 @@
+# Created by .ignore support plugin (hsz.mobi)
+### OSX template
+.DS_Store
+.AppleDouble
+.LSOverride
+
+# Icon must end with two \r
+Icon
+
+# Thumbnails
+._*
+
+# Files that might appear in the root of a volume
+.DocumentRevisions-V100
+.fseventsd
+.Spotlight-V100
+.TemporaryItems
+.Trashes
+.VolumeIcon.icns
+
+# Directories potentially created on remote AFP share
+.AppleDB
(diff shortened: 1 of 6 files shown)


marchof:
I don't think JaCoCo should get an dependency on MongoDB. As you can see from the build status the build failed as JaCoCo's minimum Java version is 1.5. MongoDB seems to be 1.6.

-1 sorry

marchof:
I would rather go e.g. for a proxy e.g. serves as a JaCoCo TCP Server and dumps incoming requests to a MongoDB instance. Anyways this would be a separate project.

ihr:
That was my initial thought still, I had some problems with the TCP server. Will dig further and if suitable create a proxy project.

Cheers

marchof:
The ExecutionDataServer example should be a good starting point:

http://eclemma.org/jacoco/trunk/doc/examples/java/ExecutionDataServer.java

ihr:
Thanks, much appreciated! A problem concerning the proxy approach is that most often network connectivity among cloud processes is via HTTP/S.

marchof:
Well, but mongoDB also uses a plain TPC connection. So what is the difference here to this PR?

ihr:
I was thinking of an approach over HTTP/S, no the one using MongoDB sink. Since you said that you don't like having an additional dependency.

PR #288, opened 2016-01-12
>= 0.7.3 breaks Gradle + JUnit + Robolectric coverage (Android)

marchof:
Version info:
Gradle 2.2.1
Android Gradle plugin 1.0.1
Robolectric Gradle plugin 0.14.1
JUnit 4.10
Robolectric 2.4

Setting up a simple test using a configuration like:

```groovy
apply plugin: 'jacoco'
jacoco.toolVersion = project.property('jacocoVersion')
```

and running my unit tests with:

    $ ./gradlew -PjacocoVersion=0.7.1.+ clean :testDebug :jacocoTestDebugReport
    ...
    $ ls -l build/jacoco/testDebug.exec
    -rw-r--r--+ 1 jhansche  staff  278218 Feb 26 16:45 build/jacoco/testDebug.exec
    $ strings build/jacoco/testDebug.exec | wc -l
        4325

This all works correctly, and I get a proper JaCoCo report generated.  Changing **nothing else** but the `toolVersion`, however, breaks the execution data that is gathered:

    $ ./gradlew -PjacocoVersion=0.7.4-SNAPSHOT clean :testDebug :jacocoTestDebugReport
    ...
    $ ls -l build/jacoco/testDebug.exec
    -rw-r--r--+ 1 jhansche  staff  36910 Feb 26 16:56 build/jacoco/testDebug.exec
    $ strings build/jacoco/testDebug.exec | wc -l
         639

You can see that the execution data file is much smaller, and contains far fewer strings.  Looking at the "Sessions" link in the top-right of the HTML report lists only *one* class from my application's package (the Android app's `R$styleable` class).  The majority of the classes listed are in one of the `org.gradle`, `org.junit`, or `org.robolectric` packages.  Using 0.7.1, it included all the `android.**` classes, and more importantly, the classes from my application.

However, when I set the `test.jacoco.classDumpFile = file("${project.buildDir}/jacoco/dump")` property, I **do** see most of my application's (tested) classes listed in that directory -- they just don't show up in the execution data file.

One other difference that I notice in the `dump/` directory (for a single example test class):

0.7.1:
build/jacoco/dump//com/example/MyClass.class
build/jacoco/dump//com/example/MyClassTest.class

0.7.4:
build/jacoco/dump//com/example/MyClass.9b700a23ba643d32.class
build/jacoco/dump//com/example/MyClassTest.22115a9aa9d0d3ee.class

I'm in a difficult position as now I am forced to use two different versions of JaCoCo:  0.7.1 with Robolectric+JUnit to get coverage of unit tests on the local JVM;  and 0.7.3+ with Calabash to get coverage of instrumentation tests running on the device's runtime -- I am forced to update to 0.7.3 in order to allow support for the Android ART runtime, as otherwise the instrumented classes cannot be installed on a Lollipop device.

What other information can I provide to help track down what is going wrong here?  If it's not a bug in JaCoCo, then who is to blame? (I realize there is a very long list of potential candidates for where it should be fixed).

### DIFF
--- a/jacoco-maven-plugin.test/it/it-customize-agent/pom.xml
+++ b/jacoco-maven-plugin.test/it/it-customize-agent/pom.xml
@@ -26,12 +26,15 @@
     <jacoco.destFile>${project.build.directory}/coverage.exec</jacoco.destFile>
     <jacoco.append>false</jacoco.append>
     <jacoco.exclClassLoaders>sun.reflect.DelegatingClassLoader:MyClassLoader</jacoco.exclClassLoaders>
+    <jacoco.inclBootstrapClasses>true</jacoco.inclBootstrapClasses>
+    <jacoco.inclNoLocationClasses>true</jacoco.inclNoLocationClasses>
     <jacoco.sessionId>session</jacoco.sessionId>
     <jacoco.dumpOnExit>true</jacoco.dumpOnExit>
     <jacoco.output>file</jacoco.output>
     <jacoco.address>localhost</jacoco.address>
     <jacoco.port>9999</jacoco.port>
     <jacoco.classDumpDir>${project.build.directory}/classdumps</jacoco.classDumpDir>
+    <jacoco.jmx>true</jacoco.jmx>
   </properties>
 
   <build>
--- a/jacoco-maven-plugin.test/it/it-customize-agent/verify.bsh
+++ b/jacoco-maven-plugin.test/it/it-customize-agent/verify.bsh
@@ -17,12 +17,15 @@ String agentOptions = "destfile=" + basedir + File.separator + "target" + File.s
     + ",includes=*"
     + ",excludes=java.*:sun.*"
     + ",exclclassloader=sun.reflect.DelegatingClassLoader:MyClassLoader"
+    + ",inclbootstrapclasses=true"
+    + ",inclnolocationclasses=true"
     + ",sessionid=session"
     + ",dumponexit=true"
     + ",output=file"
     + ",address=localhost"
     + ",port=9999"
-    + ",classdumpdir=" + basedir + File.separator + "target" + File.separator + "classdumps";
+    + ",classdumpdir=" + basedir + File.separator + "target" + File.separator + "classdumps"
+    + ",jmx=true";
 
 //backslashes will be escaped
 agentOptions = agentOptions.replace("\\","\\\\");
--- a/jacoco-maven-plugin/src/org/jacoco/maven/AbstractAgentMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/AbstractAgentMojo.java
@@ -81,6 +81,12 @@ public abstract class AbstractAgentMojo extends AbstractJacocoMojo {
 	 */
 	Boolean inclBootstrapClasses;
 	/**
+	 * Specifies whether classes without source location should be instrumented.
+	 * 
+	 * @parameter property="jacoco.inclNoLocationClasses"
+	 */
+	Boolean inclNoLocationClasses;
+	/**
 	 * A session identifier that is written with the execution data. Without
 	 * this parameter a random identifier is created by the agent.
 	 * 
@@ -188,6 +194,10 @@ public abstract class AbstractAgentMojo extends AbstractJacocoMojo {
 		}
 		if (inclBootstrapClasses != null) {
 			agentOptions.setInclBootstrapClasses(inclBootstrapClasses
+					.booleanValue());
+		}
+		if (inclNoLocationClasses != null) {
+			agentOptions.setInclNoLocationClasses(inclNoLocationClasses
 					.booleanValue());
 		}
 		if (sessionId != null) {
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
@@ -63,51 +63,25 @@ public class CoverageTransformerTest {
 	}
 
 	@Test
-	public void testHasSourceLocationNegative1() {
-		CoverageTransformer t = createTransformer();
-		assertFalse(t.hasSourceLocation(null));
-	}
-
-	@Test
-	public void testHasSourceLocationNegative2() {
-		CoverageTransformer t = createTransformer();
-		ProtectionDomain pd = new ProtectionDomain(null, null);
-		assertFalse(t.hasSourceLocation(pd));
-	}
-
-	@Test
-	public void testHasSourceLocationNegative3() {
-		CoverageTransformer t = createTransformer();
-		CodeSource cs = new CodeSource(null, new Certificate[0]);
-		ProtectionDomain pd = new ProtectionDomain(cs, null);
-		assertFalse(t.hasSourceLocation(pd));
-	}
-
-	@Test
-	public void testHasSourceLocationPositive() {
-		CoverageTransformer t = createTransformer();
-		assertTrue(t.hasSourceLocation(protectionDomain));
-	}
-
-	@Test
 	public void testFilterAgentClass() {
 		CoverageTransformer t = createTransformer();
 		assertFalse(t.filter(classLoader,
-				"org/jacoco/agent/rt/internal/somepkg/SomeClass"));
+				"org/jacoco/agent/rt/internal/somepkg/SomeClass",
+				protectionDomain));
 	}
 
 	@Test
-	public void testFilterIncludesBootstrapClassesPositive() {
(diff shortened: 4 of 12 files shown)


marchof:
@Godin Can you please review the fix? I propose to introduce a new agent option 'inclNoSourceClasses' which when set to true instruments all classes as it was the case till 0.7.2.

Godin:
@marchof it looks like both entry for changelog and commit message mention incorrect option.

marchof:
@Godin arghh, damn copy&paste. Fixed & Thx.

Godin:
@marchof except of being sad that we forced to introduce new configuration option :disappointed: looks good to me :thumbsup: note that I only looked on changes without end-to-end testing

TommyDevAll:
I faced this error on jenkins too, with jenking-jacoco-plugin 2.0.0, gradle 1.5.0 and exec files built with JaCoCo 0.7.5

Godin:
@marchof btw - something strange in this thread: @jcdom said that he faced this issue with 0.7.2 , while #272 was in 0.7.3

marchof:
@Godin Not sure what the exact problem with in @jcdom setup is. For sure it is not related to the new filtering which has been intriduced with 0.7.3.

marchof:
@jhansche @jaredsburrows @sneuberger-amazon @T3rm1 @mbells @aveuiller @TommyDevAll 

Can anyone of you please re-test with 0.7.6-SNAPSHOT? You need to specify the new option inclnolocationclasses=true for the JaCoCo agent.

jaredsburrows:
Is `inclnolocationclasses=true` a new option for `gradle`? Like:

````
jacocoTestReport {
   inclnolocationclasses=true
}
````

TommyDevAll:
I tried this way:
    jacoco {
        toolVersion = "0.7.6-SNAPSHOT"
        inclnolocationclasses=true
    }

got this error:
Error:(30, 0) No such property: inclnolocationclasses for class: org.gradle.testing.jacoco.plugins.JacocoPluginExtension_Decorated

jaredsburrows:
@marchof @TommyDevAll Actually, this is the `jacoco` repo. The `jacoco` DSL source is in the `gradle` repo. I am not sure where `inclnolocationclasses` is supposed to go. It seems like it is easier to use `maven` to add parameters for the `jacoco agent`.

marchof:
@jaredsburrows At least till we have a JaCoCo release which is picked-up by gradle.

jaredsburrows:
@marchof Why introduce this new config option? This will prevent most `gradle` users from upgrading.

marchof:
@jaredsburrows because we have a conflict between different different use cases:

1) Exclude classes without source location as such classes are typically dynamically created e.g. by mocking or persistence frameworks and cause problems
2) Include classes without source location in case of Android where no proper source location is reported.

marchof:
@jaredsburrows Alternatively you might report this issue to Android/Robolectric to fix the root cause.

jaredsburrows:
@marchof I can use different Robolectric and Android with Jacoco 7.2 just fine. I just can't upgrade jacoco. Even with a normal groovy/java project, I can't use 7.3+, I have to use 7.2.

I am wondering if this is a gradle issue. It seems gradle only supports up to 7.1: https://github.com/gradle/gradle/blob/6277a4dc70fbeea83c111e75c95ba851d1e56ffc/subprojects/jacoco/src/integTest/groovy/org/gradle/testing/jacoco/plugins/JacocoVersionIntegTest.groovy#L25.

TommyDevAll:
@jaredsburrows have you tried with gradle 1.5.0?

    jacoco {
        toolVersion = "0.7.5.201505241946"
    }

this works for me (jacoco 0.7.5 i mean, not the coverage)

jaredsburrows:
@TommyDevAll Gradle 1.5.0? That is a really old version. What do you mean, jacoco works for you but not the coverage? jacoco == coverage.

TommyDevAll:
sorry i mean 2.9.... with android plugin for gradle @1.5.0 
i'm able to update to 0.7.5...

everybody here know that jacoco is for coverage.... we are trying to help each other.... do you remember?

jaredsburrows:
I am using gradle 2.10, Android plugin 1.5.0. So now you are saying 0.7.5
works? Did you do a gradle clean testDebug?

TommyDevAll:
i'm trying to figure out why you can't upgrade to 0.7.5. Did you faced some problem?
i never told you that jacoco 0.7.5 works for me... i told you that i'm able to upgrade to that version.

Now i'm trying to use the 0.7.6-SNAPSHOT with the new attribute inclNoLocationClasses=true but can't find the right place to add this new attribute.... is it clear?

jaredsburrows:
By saying that you "can upgrade", I would assume you mean that that version
works. But then you said, the coverage doesn't? To be clear, 0.7.5 coverage
does not work.

TommyDevAll:
ok

Godin:
@marchof FYI my colleague pointed out that there is an issue in combination JaCoCo(>= 0.7.2) + Gradle, even without Android and Robolectric, so I'll do end-to-end testing today.

Godin:
@marchof after investigations JaCoCo works just fine with Gradle, when something else not involved into process of execution of tests:
* in our particular case seems that issue might come from JetBrains/gradle-intellij-plugin#65 or even from custom ClassLoaders in IntelliJ itself
* retested Gradle+Android+JUnit+Robolectric (https://github.com/MeetMe/jacoco-288)
* new option `inclNoLocationClasses` solves the issue in both of the cases :thumbsup:
* but JaCoCo Gradle Plugin needs to be updated to expose this new option, because seems that there is no way to specify arbitrary option - https://github.com/gradle/gradle/blob/master/subprojects/jacoco/src/main/groovy/org/gradle/testing/jacoco/plugins/JacocoTaskExtension.groovy

Hence I'm wondering what is better on our side - enabled by default or disabled with PR for Gradle?

marchof:
@Godin Many thanks for retesting!
I also think Gradle plugin needs to be updated. If we change the default we cause a regression for users of mocking libraries and lock them out from Gradle usage -- assuming Gradle is not only used for Android projects.

Godin:
@marchof ok, btw quick and dirty patch for Gradle, which allows to specify arbitrary options can be found in Godin/gradle@7e4f886cf5b076ce983b2482da92cbd77cb89672

emartynov:
Guys, when to expect 0.7.6 release? Any workaround with snapshot usage and some gradle snipet?

TommyDevAll:
@emartynov +1
In my case all the "non-robolectred" test classes was correctly analysed and included in the report. Is it the same for you?

Godin:
### Note for Gradle users
JaCoCo 0.7.6 has been released, however
- you need to wait for Gradle/Gradle#575
- either without it find a way to pass new option (`inclnolocationclasses`) to agent

jaredsburrows:
:+1: 
@Godin Thanks for the update!

rliesenfeld:
If I am not mistaken, this problem has the same root cause as when using PowerMock, which is that the custom JUnit test runner creates copies of classes without preserving the original protection domain. I opened an issue in the PowerMock tracker for it: https://github.com/jayway/powermock/issues/564

I believe the same applies to the Robolectric custom test runner, which does very much the same thing as PowerMock.

dampcake:
Small Android sample with the changes here:
https://github.com/dampcake/Robolectric-JaCoCo-Sample
dampcake/Robolectric-JaCoCo-Sample@f9884b96ba5e456cddb3d4d2df277065bb26f1d3 shows the change to make it work with JaCoCo 0.7.6

RomainPiel:
Just catching up with this, so to sum up:
- there is no potential fix for jacoco <2.7.6
- jacoco 2.7.6 requires gradle 2.13

Correct?

That's a bit annoying. Any chance there could be a hacky fix for previous versions (on robolectric or jacoco)?

Godin:
@RomainPiel JaCoCo doesn't require Gradle. Options:
 1. upgrade Gradle
 2. do not use Gradle Plugin for JaCoCo and attach JaCoCo 0.7.6 agent to your tests by yourself - it is just "-javaagent:..." option for JVM
 3. use JaCoCo <= 0.7.2
 4. ask Gradle to backport changes ( gradle/gradle#575 ) to older version or create you own version with such backport

RomainPiel:
@Godin cool thanks!

mkjensen:
Thank you all for pointing me in the right direction. I had no luck using JaCoCo 0.7.6 and Gradle 2.13 but I was not really comfortable using that anyway since Android Studio 2 uses Gradle 2.10 per default. However, I also had problems getting JaCoCo 0.7.2 to work since toolVersion didn't seem to get respected. I ended up with the following hack that I'm looking forward to deleting as soon as possible: https://github.com/mkjensen/danish-media-license/commit/5d9eb52a0d07790b132fb0a6818aa78a5bbbfed1

PR #374, opened 2016-02-11
Implement probes as abstract type and provide alternative probe types

oazmon:
This feature is a big commit. 
It is currently in use in a large company in Mountain View.
I am contributing it back as a feature branch so it is easier to check and look at.
This provides count and parallel count as alternative implementations to
boolean[] so as to collect Empirical Big O and parallel percent and
report on them when available.
All documentation, tests, etc. have been included.
Backward compatibility is 100% to existing 0.7.5 JaCoCo.
I understand that alternatives to boolean[] were not implemented in the past because they are slower and their sync time is an issue. By making the old and new both available and some performance tweaking this has been overcome.

The following write-up summarizes the inspiration for this feature. I hope that after reading it and looking at the code you will see the benefit of accepting this pull requests. It also contains some minor bug fixes that we detected along the way. They can be separated out if needed.

*\* Continuous Performance

Performance testing might be defined as “the technical investigation done to determine or validate the speed, scalability, and/or stability characteristics of the software under test”. The reasons for such testing might be grouped into the following categories: (a) Determining the required tuning and current limits of an application, (b) identifying performance related architectural design faults, and (c) identifying performance related code faults.

The most common performance concerns related to Web applications are “Will it be fast enough?”, “Will it support all of my clients?”, “What happens if something goes wrong?”, and “What do I need to plan for when I get more customers?”. In casual conversation, most people associate “fast enough” with performance testing, “accommodate the current/expected user base” with load testing, “something going wrong” with stress testing, and “planning for future growth” with capacity testing. Collectively, these risks form the basis for the four key types of performance tests for Web applications.

However, in a component continuous testing arrangement, only the third category applies: finding code faults. Furthermore, under these conditions, any traditional load, stress, or capacity testing results are not easily translatable into production expectations. Furthermore, these tests tend not point to the problematic segment of code and leave us guessing.

What we need is a performance testing-tool that does not require a special controlled environment, that will clearly identify performance risks in those conditions, and that will identify the offending section of code. Big-O is such a predictor, but is hard to calculate for large bodies of code. Furthermore, it tends to flag issues where there are none, such as when using an O(n2) sort for lists of less than 10 items.

In April 2009, SF Goldsmith published his PHD thesis titled “Measuring Empirical Computational Complexity” in which he demonstrate a technique of arriving at an empirical Big-O while side stepping the over-flagging issue. A search of the Internet has not shown any open-source or commercial implementations for any platform, but has found already over 53 papers referencing this work. [Ref: http://sfg.users.sonic.net/berkeley/trendprof-fse-2007.pdf and http://www.eecs.berkeley.edu/Pubs/TechRpts/2009/EECS-2009-52.pdf ]

There is one type of problem not identified by this work that is important in an elastic cloud situation. This metric has been identified by a luminary of our field, Gene Amadhl, and is known as Parallel Percent. This metric predicts the speed up achieve when adding processors to a computation. When a piece of software is, for example, 50% parallel, it will achieve no performance gain beyond double its single processor performance, no matter how many processors are added. And a 95% parallel application will achieve no performance gain beyond 20 times its single processor performance. This metric can be safely measured empirically as the percentage is independent of the test machine’s performance.


oazmon:
I see the build failed. I don't have JDK 5 on my MAC. Setting up a VM and rebuilding with JDK 5.

marchof:
@oazmon Thanks for this impressive contribution! In the past I had strong opinion against mixing code coverage with performance profiling and stay focused on code coverage for JaCoCo.
I promise I will study your PR carefully and reconsider the scope of JaCoCo! Please give me some time for this as I'm quite busy with my day job.

oazmon:
@marchof take as much time as you need. In any case I am available for any Q.

bjkail:
I'm not a maintainer and have not extensively reviewed the code, but a few comments:

1.  It appears you've modified all the instrumentation code to have direct references to internal JaCoCo classes (e.g., ProbesArrayService, IProbeArray), but there is a large IRuntime infrastructure within JaCoCo to avoid these direct references.  See the "Coverage Runtime Dependency" section of [Implementation Documentation](http://eclemma.org/jacoco/trunk/doc/implementation.html).  Perhaps the contract could be adjusted to "return" a List object rather than a direct boolean[], and you could "abuse" List.get to indirectly call the increment methods, similar to how Object.equals is currently "abused" to return the boolean[] in the first place.

2.  It appears that you've changed the API (non-internal packages).  I don't know whether the JaCoCo project allows API breaking changes since it's not 1.0, but it would be preferable for me if there were a way to make these changes without breaking the non-internal API.  For example, have ExecutionData.getProbes lazily initialize a probes array as needed, and introduce a new method like getProbesObject (or whatever) to return IProbes.

marchof:
Due to our limited resources and other important tasks like filtering we will not add more functionality beyond plain code coverage to JaCoCo. Sorry about this.

PR #376, opened 2016-02-15
AgentOptions should handle invalid args properly

marchof:
Fix for regression caused by fix for #358

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/runtime/AgentOptionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/runtime/AgentOptionsTest.java
@@ -362,7 +362,7 @@ public class AgentOptionsTest {
 
 	@Test(expected = IllegalArgumentException.class)
 	public void testInvalidOptionKey() {
-		new AgentOptions("destfile=test.exec,something=true");
+		new AgentOptions("destfile=test.exec,Some-thing_1=true");
 	}
 
 	@Test(expected = IllegalArgumentException.class)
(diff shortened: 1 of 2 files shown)


marchof:
@jochenberger Hi can you please review this fix for your PR? We detected a serious regression.

jochenberger:
Looks good to me. Could you please add another test for a `destfile` with a dash or an underscore, just to be sure?

marchof:
@jochenberger Done.

Godin:
@marchof I don't want to open Pandora box here, but to be honest - in general I don't like the initial case, which we are trying to cover here, and the way to cover it. It looks like a kind of undocumented workaround. Also one could say that ``=`` (equals sign) not supported as part of path to dump file and we'll get an ambiguity:
```
java -javaagent:destfile=jacoco,output=file
```
What is it - name of destination file or two options? Maybe would be better to allow usage of Java property file to specify options ("response file")
```
java -javaagent:jacocoagent-0.7.5.jar=@path-to-file-with-options
```
?

However this patch improves situation compared to what was before. So if now this looks safe for you, then let's go ahead with this change and keep in mind "response file" option for future.

marchof:
@Godin I like the idea of supporting property files in addition. We actually do today for the offline mode and AgentOptions supports this.
In this case we need to check how we configure the agent in our Ant and Maven goals: Should these create propertly files to also support special characters? But I prefer to discuss this in a separate issue and apply the fix here instead of roll-back.

Godin:
@marchof ok :+1:

PR #386, opened 2016-03-20
Add support for non-java source files which are not organised by package.

aardvark179:
Not all languages on the JVM follow the Java way of organising source files by package. In order to produce line-by-line coverage reports for such languages we need to check source file locations relative to the supplied source directories but not taking account of the java package the class belongs in.

### DIFF
--- a/org.jacoco.ant.test/src/org/jacoco/ant/AntFilesLocatorTest.java
+++ b/org.jacoco.ant.test/src/org/jacoco/ant/AntFilesLocatorTest.java
@@ -59,11 +59,19 @@ public class AntFilesLocatorTest {
 		assertContent(source);
 	}
 
+	@Test
+	public void testCanGetResourceWithWrongSlashes() throws IOException {
+		locator.add(createFile("org/jacoco/example/Test.notjava"));
+		final Reader source = locator.getSourceFile("",
+				"org\\jacoco/example/Test.notjava");
+		assertContent(source);
+	}
+
 	private Resource createFile(String path) throws IOException {
 		final File file = new File(folder.getRoot(), path);
 		file.getParentFile().mkdirs();
-		final Writer writer = new OutputStreamWriter(
-				new FileOutputStream(file), "UTF-8");
+		final Writer writer = new OutputStreamWriter(new FileOutputStream(file),
+				"UTF-8");
 		writer.write("Source");
 		writer.close();
 		return new FileResource(folder.getRoot(), path);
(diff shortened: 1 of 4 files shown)


marchof:
@aardvark179 What JVM language are you referring to?

aardvark179:
In this case Magik, a dynamically typed language which I've helped port to the JVM, but the same properties hold true for classfiles generated by Nashorn, and probably other languages ported to the JVM.

marchof:
@aardvark179 So in this case all source files are in the same folder but the compiler creates some package names for the class files, right?

aardvark179:
The source files are organised according to the existing conventions of the language. They are organised in products and modules, and can be organised arbitrarily within a module. So they are not all in one directory, and are not organised quite according to the java packages of the classes (though I have ensured that that those also reflect the product/module structure as this makes finding classes for debugging and understanding coverage reports much easier).

marchof:
There are a couple of JVM languages that do not have such a problem, see our examples for Groovy, Scala, Kotlin and Xtend here: 

https://github.com/marchof/jacoco-maven-examples

I do not understand how source lookup should work unambiguously when source files of all packages are "flattened" into a single directly.

aardvark179:
The languages you've listed were all designed for the JVM, and so could make decisions about package and source structure that fit with the assumptions of the JVM. When porting other languages we cannot force a change in how source is normally laid out on disk.

When we record a source path in a class file we do so from a root source directory, so it might be something like, "modules/my_modules/source/some_file.magik". and we will record the package as magik.my_product.my_module. It's true that the system could be fooled by a user making the file magik/my_product/my_module/modules/my_modules/source/some_file.magik, but you can already generate nonsense reports from Jacoco by specifying incorrect source paths so I don't see it as a major issue. This change will never match some_file.magik in the root source directory because the relative path is stored as part of the source file data, and Jacoco doesn't try to alter that in any way.

I agree that a source file recorded as "MySource.java" should only be looked up in the relevant Java package, so I did add a check so no other attempts would be made on Java files, and would be happy to make that more configurable so that only files with particular extensions are considered.

marchof:
Thanks for the detailed explanation of your use case!

I don't want to change the look-up behaviour for Java class files as I expect issues with non-unique local class names. So what other options do we have?

1. Ignore Java package when source attribute already includes folders (i.e. slashes)
2. New configuration option "ignorepackage" for source lookup
3. Ignore package for specific file extensions

Which of those would work for your scenario?

aardvark179:
I think any of those would work, I'll ask on the mlvmmailing list and see what would work well for other language implementations.

marchof:
@aardvark179 Then number 1 would be my preferred solution: No need for extra configuration and does not affect source lookups for Java class files.

marchof:
@aardvark179 Any update on this?

marchof:
No input, no fix.

PR #387, opened 2016-03-21
Don't write empty probe arrays to exec files

marchof:
On the mailing list it was proposed to not dump empty probe arrays to execution data files:

https://groups.google.com/forum/?fromgroups=#!topic/jacoco/wQhTVf2zEtI

In case of repeated dumps (per-test coverage) this can lead to a significant performance improvement:

* test time: 12hrs -> 1hr
* jacoco.exec size: 93GB -> 1GB

Within JaCoCo there is no use for empty probe arrays vs. non existing ExecutionData entries. Therefore there is no need to dump such entries.


marchof:
@Godin May I ask for a review, please?

Godin:
@marchof sure! This seems to be a huge improvement. Will do today.

marchof:
@Godin Thx, feel free to merge if you agree with the implementation.

Godin:
:thumbsup:

Issue #18, opened 2012-08-29
Multi-report merging (within single project)


marchof:
**gelephantastic@SF**

There are a few threads about Maven plugin datafile merging. I'd like to write in about the 2 major Maven merge use cases so you're aware. One is trivial, one is not. The trivial case is for multiple test types: unit tests, functional tests, and perhaps even integration tests or acceptance tests (like Concordion). For this case, I would expect that you could list additional, optional extra report files to conditionally be merged in at the report generation phase. This is trivial because it's all within a single Maven module. The harder case is merging the data file in a multi-project module, where I may want the parent pom to identify every report file that was used in any reactor module, and merge that into a master report for the whole build. This would ideally include all the "extra" reports from the trivial case above. I might try to code up the first use case. I expect I will tackle the problem exactly as I've described; the main dataFile parameter (defaulted to target/jacoco.exec) will still be expected. I might add either a list of files to (optionally) merge, or a directory that contains all the additional data files to be merged. It might be good to switch from target/jacoco.exec to using a directory, like target/jacoco/execution.exec, and then have the default behavior to merge everything in there, perhaps even then outputting the merged report to target/jacoco.exec. It's getting fuzzy here, since I'm out on a limb, but just a thought.

* recursive-merge-patch.diff: https://gist.github.com/3518923

**marchof**

I like the idea to create a report on multiple exec-Files (like the Ant task also supports).

But to be honest I don't like the idea to recursively search directories. This might cause searching thousands of directories. Also there might be other *.exec files in the directories that should not be considered.

What is best practice here? Can't we supply a file set?

**gelephantastic@SF**

I picked the recursion simply because it was the smallest change (and also backwards compatible), but I certainly agree with your concerns. I don't know what the best practice is.. probably some sort of includes/excludes and a directory, or just statically listing all the reports that are expected?

In the case of the latter, do you use an API-breaking dataFiles/dataFile arrangement in the XML config, or do you keep dataFile and add an extraDataFiles/dataFile configuration block?

In the case of the former, you would keep a single dataFile but also add, I suppose, a dataFileFilter/include=*.exec or something of that nature. That option still requires recursion, but at least it is selective about what it picks up instead of trying to merge everything.

**marchof**

We can add a new property ("dataFiles") and deprecate the old one.

**gelephantastic@SF**

[...]

PR #388, opened 2016-03-22
Aggregate Report Maven Goal

marchof:
To finally offer a solution for multi module Maven projects (#18) I did a detailed analysis of the issue together with @jwloka. We documented the results here:

https://github.com/jacoco/jacoco/wiki/MavenMultiModule

As part of the analysis we also reviewed the existing PR #97. We came up with a different solution which should cover the described use cases and should be more robust in terms of build configuration and execution.

Please feel free to add comments regarding the general approach as well as the implementation


velo:
I would say that having a "report" module is not feasible....

To have a report module, this modules must depend on ALL other modules.

IMHO, there should be an option that would make the reports to be written on the top level project.
OR
Something like deployAtEnd... https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#deployAtEnd

marchof:
@velo The report has to be written after all modules have been built and all tests have been executed. If you know how this can be properly implemented in a Maven goal of the top level project please let us know. From our understanding the goal is executed **before** the sub-modules. See linked document for a detailed discussion.
And no, we don't want extra goals specified on the command line.

velo:
not if you enable threads in order to have a speeding build...  in that case, modules order are ignored.

marchof:
@velo Ok, then how do you implement deferred execution of a top level project's goal in a single threaded build?

velo:
Something like this:
https://github.com/apache/maven-plugins/blob/trunk/maven-deploy-plugin/src/main/java/org/apache/maven/plugins/deploy/DeployMojo.java#L155

marchof:
@velo WTF?!? Do we want to maintain such a hack...? Anyways, many thanks for that reference!

velo:
well, I'm my opinion assuming that projects will be executed at right order is a big problem.

Specially when you add mvn -T to the mix.

velo:
I made this to show what I mean:
https://github.com/jacoco/jacoco/pull/389/files

jacoco-maven-plugin.test/it/it-report-aggregate/report/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>report</artifactId>
+  <name>Aggregate Report</name>
+  
+  <dependencies>
+    <dependency>
+      <groupId>jacoco</groupId>
+      <artifactId>child1</artifactId>
```
velo: this is why multiple threads did not fail....


now, if someone create a child3, child3 may or may not appear on the reports...

unless the same someone remember to add child3 into this dependency list...

marchof:
@velo Absolutely, this is the advantage and disadvantage of the approach:
* + explicit dependencies allows to control what should be in the report
* + proper execution with the Maven live cycle (also for multi-threading)
* - manual, redundant maintenance of the report scope

velo:
Also
* - need for a report module

marchof:
> child3 may or may not appear on the reports... 

This is not true: The report-aggregate goal only adds modules to the reportwhich are declared as a dependency. So if a module is not declared as a dependency it will never show up. If it is declared as a dependency Maven should ensure a proper execution sequence even with multiple threads.

Godin:
@marchof I didn't looked deeper yet, but it seems that this covers case of JaCoCo itself only partially, am I right?

marchof:
@Godin Sure? I think the goal could be directly used in the doc module (or a new one). I'll try to use report-aggregate within our build.

velo:
So, if I created something similar to deployAtEnd ( run coverage report on the last module inside reactor and record results onto reactor project would you guys be willing to consider it?

marchof:
Absolutely! We need to make sure "something" is robust enough to work in most scenarios. For example usage of static tracker in "deployAtEnd" seems very unstable (Maven embedded in other JVM, Multiple aggregate Reports in a more complex module structure etc.)
Maybe someone can get in touch with Maven developers and ask for best practices.

Godin:
@marchof I had a talk with @rfscholte at FOSDEM about `deployAtEnd` and as far as I understood as of today there is no way to do better. Don't know how it plays in case of embedding of Maven.

Godin:
@marchof about usage within our build - I made a wrong guess, because didn't found such a change :wink:

rfscholte:
I have several proprosals on how to improve the lifecycle handling in Maven. Once https://issues.apache.org/jira/browse/MNG-5666 there's no need for installAtEnd/deployAtEnd anymore and it will work for all Maven projects. installAtEnd/deployAtEnd works for about 90%+ of the projects, it fails for custom packaging types and lifecycle forks because it Maven creates a new classloader, which makes it impossible for the maven-deploy-plugin to detect when the build is really finished. So the real solution means improving Maven core.

marchof:
> works for about 90%+ of the projects

This doesn't sound like we should build a solution on this approach. The 10% of projects will cause lots of support requests here.

marchof:
@Godin Of course, one should always eat your own dog food. So here we eat: 2401392a76300674ff80143a1b6add2a3de5d084

org.jacoco.tests/pom.xml
```diff
@@ -33,6 +33,7 @@
     <module>../org.jacoco.ant.test</module>
     <module>../jacoco-maven-plugin.test</module>
     <module>../org.jacoco.examples.test</module>
+    <module>../org.jacoco.tests.coverage</module>
```
Godin: @marchof any particular reasons why this can't be part of existing module `org.jacoco.doc` ?

jacoco-maven-plugin.test/it/it-report-aggregate/invoker.properties
```diff
@@ -0,0 +1 @@
+invoker.goals = -T2 clean install
```
Godin: IMO approach based on dependencies does not require special testing of parallel build by definition. Parallel builds are available only since Maven 3.x , so that this test can't be executed with Maven 2.x.

org.jacoco.tests/pom.xml
marchof: Actually I also tried this: I had trouble with dependencies to same module with different qualifiers (it makes appearing those bundles twice in the report). Nut sure whether we actually need these dependencies as we now create the download zip in a separate project.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,163 @@
...
+
+	@Override
+	void loadExecutionData() throws MavenReportException {
+		final ExecFileLoader loader = new ExecFileLoader();
+		for (final MavenProject dependency : findDependencies(
+				Artifact.SCOPE_COMPILE, Artifact.SCOPE_TEST)) {
+			// TODO Use configured location from project
+			// TODO Clarify when to include exec data from integration tests
+			final File execFile = new File(
+					dependency.getBuild().getDirectory(), "jacoco.exec");
```
Godin: Ouch.

marchof: Not sure how to fix this: Where do we want to retrieve the location from?

mfriedenhagen: Not sure how retrieving the location could be done, but IMO one of the primary use cases for aggregated reports are integration tests which typically use `jacoco-it.exec` by default, so I would revert the default here?

mfriedenhagen: What about two goals, `report-aggregate` and `report-aggregate-integration` which would just pick up `jacoco-it.exec`?

marchof: I don't like duplication of goals, should be configurable which files to pick (test, it or both).

Godin: @mfriedenhagen I also don't like duplication of `goals`. The most tricky part here IMO - what if user customized location(s)?

marchof: I'm thinking of a file filter which is applied to each project's root directory. Default would be includes="target/jacoco.exec,target/jacoco-it.exec".

mfriedenhagen: As a last comment concerning goals for a JaCoCo user it could be more convenient to write
```
<executions>
<execution>
<goals>
<goal>report-aggregate</goal>
<goal>report-aggregate-integration</goal>
</goals>
</execution>
</executions>
```
instead of fiddling around with multiple executions. And we already have `report-integration`, so it would be more consistent.

johnoliver:
I have given this a try and can confirm that it does do what #97 is after. In general it looks fine, I have not been able to run your tests, but I have never got jacoco's tests to run successfully so that is a different issue.

I think all of my issues with this implementation have already been mentioned and discussed. Having to add an extra module with dependencies is not great, but that issue is a long standing maven problem and I have no better solution to it. Making the exec file name configurable would probably be an improvement.

I think the biggest improvement would be if you added the `reactorProjects` to the places scanned for exec files. i.e I added:

```
List<MavenProject> reactors = new ArrayList<MavenProject>(reactorProjects); reactors.remove(getProject()); 
result.addAll(reactors);
```
To `findDependencies`. This will then allow the option of compiling an aggregate report with:

```
mvn clean install jacoco:report-aggregate
```

And we would not need to create the extra report module. Although I understand you don't want to have to enforce adding extra goals to the command line, this code allows you the option to do so, and I cant think of any reason why it would cause any problems. Since we build our report on a CI server I would much prefer to add this extra goal to the build rather than add the extra module for reporting. It is preferable since it is far more robust to changes in the code base than adding the extra module, not to mention a reduction in overall code. 

I think this is a good addition as it just allows the option between adding an extra goal or the extra module. 

If you don't want to do this, then also reactorProjects is an unused field and should be removed.

Other than that, good work, thanks.

marchof:
@johnoliver Many thanks for your detailed feedback! I like the idea of having a goal which supports both scenarios, so users can choose. I'll add a IT test and try the implementation you proposed.

Godin:
@johnoliver Let's take an example of project with sub-module "lib" used by two other sub-modules "app1" and "app2". Extra module with report allows to show in a single place coverage of "lib" by tests in "app1" and "app2".

But I don't see which use case is addressed by your proposal regarding "reactor". Execution of "mvn clean install jacoco:report-aggregate" will generate coverage of "lib" in two places "app1" and "app2", because they have dependency on "lib", which will be scanned for "jacoco.exec".

So could you please elaborate?

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,163 @@
...
+ * 
+ * @goal report-aggregate
+ * @since 0.7.7
+ */
+public class ReportAggregateMojo extends ReportMojo {
+
+	/**
+	 * The projects in the reactor.
+	 * 
+	 * @parameter expression="${reactorProjects}"
```
Godin: This generates warning during build:
```
...
[INFO] --- maven-plugin-plugin:3.3:descriptor (default-descriptor) @ jacoco-maven-plugin ---
[INFO] Using 'UTF-8' encoding to read mojo metadata.
[INFO] Applying mojo extractor for language: java-annotations
[INFO] Mojo extractor for language: java-annotations found 0 mojo descriptors.
[INFO] Applying mojo extractor for language: java
[WARNING] org.jacoco.maven.ReportAggregateMojo#reactorProjects:
[WARNING]   The syntax
[WARNING]     @parameter expression="${property}"
[WARNING]   is deprecated, please use
[WARNING]     @parameter property="property"
[WARNING]   instead.
...
```

johnoliver:
@Godin yes there was a mistake with the description in the last post. The build and `aggregate-report` would have to be 2 separate calls to maven. i.e

```
mvn test
mvn jacoco:report-aggregate
```

What will happen here is tests will execute for all the reactor projects `lib`, `app1`, `app2` and the `jacoco.exec` files will be generated for each module with the cross module coverage contained. The `report-aggregate` is now run for every reactor after the testing has been executed giving the parent a chance to aggregate reports after the `jacoco.exec` files have been generated.

Now I accept multiple calls to maven is not a great solution by any means, however, in all this we are making the best of a bad situation introduced by maven. Given our CI setup, making multiple calls to maven to generate the report is not an issue, and as I said before between the bad choice of having an extra module for generating reports that could rot, and making multiple calls to maven, I would choose the latter as I have said before as it would be more robust. However with the current code this would not work as it only looks at dependencies for `jacoco.exec` files, however if it looked at reactors too this would work. And again does not damage the current behaviour, merely offers the option to do this.

There is also the situation I have found in testing, if the root module from which you execute `mvn` is not referenced by the reactor module in the `<parent>` section, then the root actually seems to come last. If you therefore create a structure where the root module executed is not a "parent" of the modules being tested you can do it all in one go and even without any extra command line invocations or report modules, but only IF you include reactors as well as dependencies. I admit this sorting order in this case does not seem to be defined in http://maven.apache.org/guides/mini/guide-multiple-modules.html so I guess could change.

johnoliver:
@Godin @marchof Actually I have taken a bit of a closer look, and this is probably not a good idea. Forget about it.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
marchof: Fixed.

jwloka:
Summarizing the discussion it seems that we're on track to provide multi-module reports in an acceptable way with this PR pretty soon. They're are still a few concerns about an extra "report module" which must depend on all modules to be reported. A deployAtEnd-like implementation was suggested und would have been a good alternative but it appears to be not maintainable and has problems with custom packaging types and lifecycle forks.

@marchof  They're only a few TODOs left. Do you want my help with anything specific?

marchof:
@jwloka Sure! Any ideas how to resolve the TODOs are welcome, e.g. as PRs to this branch.

marchof:
@Godin @jwloka May I ask you to review this PR? We can now create the exact same coverage report with Maven than we had with Ant before.

jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>-ea ${argLine}</argLine>
```
Godin: @marchof configuration for `maven-surefire-plugin` seems completely useless here.

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -53,52 +36,65 @@
 	 *            default-value="UTF-8"
 	 */
 	String outputEncoding;
+
+	/**
+	 * Name of the root node HTML report pages.
+	 * 
+	 * @parameter default-value="${project.name}"
+	 */
+	String title;
```
Godin: `@since` missing

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -53,52 +36,65 @@
...
+
+	/**
+	 * Footer text used in HTML report pages.
+	 * 
+	 * @parameter
+	 */
+	String footer;
```
Godin: `@since` missing

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Locale;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.project.MavenProject;
+import org.jacoco.report.IReportGroupVisitor;
+
+/**
+ * <p>
+ * Creates a structured code coverage report from multiple projects (HTML, XML,
```
Godin: Maybe change to
```
Creates a structured code coverage report (HTML, XML and CSV) from multiple projects within reactor.
```
?

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,180 @@
...
+
+	/**
+	 * Output directory for the reports. Note that this parameter is only
+	 * relevant if the goal is run from the command line or from the default
+	 * build lifecycle. If the goal is run indirectly as part of a site
+	 * generation, the output directory configured in the Maven Site Plugin is
+	 * used instead.
+	 * 
+	 * @parameter 
+	 *            default-value="${project.reporting.outputDirectory}/jacoco-aggregate"
```
Godin: I think that for consistency we can safely use `jacoco` folder here - module with aggregated report is not supposed to have other jacoco report.

marchof: Not sure about this: Please see test "it-site". It creates both reports. Maybe this is another problem.

org.jacoco.tests.coverage/pom.xml
```diff
@@ -0,0 +1,130 @@
...
+      <groupId>${project.groupId}</groupId>
+      <artifactId>org.jacoco.examples.test</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <properties>
+    <maven.deploy.skip>true</maven.deploy.skip>
+    <jacoco.includes>**/*.class</jacoco.includes>
```
Godin: `jacoco.includes` seems useless

marchof: @Godin Actually is required here otherwise the shaded agent JAR becomes part of the report or the org.jacoco.agent module (analyzer recursively inspects jar files by default).

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -41,13 +41,15 @@
...
 	 * @parameter default-value="${project.name}"
+	 * @since 0.7.7
 	 */
 	String title;
 
 	/**
 	 * Footer text used in HTML report pages.
 	 * 
 	 * @parameter
+	 * @since 0.7.7
```
Godin: @marchof and so this looks like a separate enhancement, which IMO deserves a separate mention in changelog.

org.jacoco.tests.coverage/pom.xml
Godin: @marchof good point! and so I think would be nice to document this here.

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
marchof: @Godin Good idea. Haven't touched changelog yet for this PR (will conflict with master anyways).

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
Godin: @marchof I see. BTW behavior of site generation in regards to new `report-aggregate` mojo is an interesting question :wink:

Godin: and BTW there is no assertions on `report-integration` in `it-site`

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,180 @@
...
+	@Override
+	protected String getOutputDirectory() {
+		return outputDirectory.getAbsolutePath();
+	}
+
+	@Override
+	public void setReportOutputDirectory(final File reportOutputDirectory) {
+		if (reportOutputDirectory != null
+				&& !reportOutputDirectory.getAbsolutePath().endsWith(
+						"jacoco-aggregate")) {
```
Godin: @marchof this condition looks weird / hackish for me.

marchof: @Godin Same for me, I simply took the pattern from the existing report goal.

jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+
+    <artifactId>it-report-aggregate</artifactId>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>child1</module>
+        <module>child1-test</module>
+        <module>child2</module>
+        <module>report</module>
+    </modules>
```
Godin: minor - indentation mismatch

jacoco-maven-plugin.test/it/it-report-aggregate/verify.bsh
```diff
@@ -0,0 +1,41 @@
...
+
+File reportChild1test = new File( basedir, "report/target/site/jacoco-aggregate/child1-test/index.html" );
+if ( reportChild1test.isFile() ) {
+    throw new RuntimeException( "Report for child1-test should not be created." );
+}
+
+File reportChild2 = new File( basedir, "report/target/site/jacoco-aggregate/child2/index.html" );
+if ( !reportChild2.isFile() ) {
+    throw new RuntimeException( "Report for child2 was not created." );
+}
```
Godin: minor - missing new line

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -53,52 +36,67 @@
...
+	 * 
+	 * @parameter default-value="${project.name}"
+	 * @since 0.7.7
+	 */
+	String title;
+
+	/**
+	 * Footer text used in HTML report pages.
+	 * 
+	 * @parameter
```
Godin: I think would be nice to allow definition of values for `title` and `footer` when `report` mojo executed from command-line.

org.jacoco.tests/pom.xml
Godin: @manandbytes In fact I think that we should merge both "doc" and "coverage" into this module for distribution since they all share the same purpose and same nature. We can deal with this separately, however I also have feeling that your statement about classifiers reveals something in implementation of this new `report-aggregate`.

Godin:
@marchof I'm wondering about following scenario:
* module A contains class A, but no tests for it
* module B contains class B, which depends on class A, and test for B indirectly covers A

it is IMO questionable whether aggregate report should show class A as covered or not.

marchof:
@Godin This question of "cross module coverage" probably depends on the scenario: If B are integration tests users probably want to see coverage on A. If B is plain unit tests I would prefer to only show code coverage on B. Actually our own build implements this by limiting the instrumentation scope of the agent. This could also be implemented in reporting by loading execution data separately for each bundle. I propose to discuss this requirement in future in a separate issue.

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
marchof: @Godin So you're thinking of something like this?

    @parameter property="jacoco.reportTitle"
    @parameter property="jacoco.reportFooter"

org.jacoco.tests/pom.xml
marchof: @Godin I was thinking of a new module org.jacoco.build.reports where we create all the aggregate reports (JavaDoc, JUnit, Coverage). Let's handle this in a separate PR
Indeed you're right, we might need to consolidate multiple dependencies to the same project with the `report-aggregate` goal. Again I would handle this in a separate PR when we learn how the new goal works in real life projects.

Godin: @marchof Actually the more I think about this story of exclusion of dependencies, the more I tend to think that `report-aggregate` can simply be an enhancement of `report` by ability to select dependencies.

marchof: @Godin There are some differences in confuguration e.g. regarding exec file locations which would be an incompatible change of the existing goal. Also this would mean that we create a report group in any case. To move on I would go with the new goal and maybe consolidate later as we learn more about actual usage.

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
Godin: @marchof I withdraw this request for now - let's deal with this separately, since some other important properties are not available for customization from command-line and those properties are not that important.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -45,6 +45,8 @@
  * </ul>
  * 
  * @goal report-aggregate
+ * @requiresProject true
+ * @threadSafe
```
Godin: Question to myself - do we need `@requiresDependencyResolution test` here?

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -0,0 +1,182 @@
...
+ * execution data but should not be part of the report itself:
+ * </p>
+ * 
+ * <ul>
+ * <li><code>compile</code>: Project source and execution data is included in
+ * the report.</li>
+ * <li><code>test</code>: Only execution data is considered for the report.</li>
+ * </ul>
+ * 
+ * @goal report-aggregate
```
Godin: Question to myself - do we need `@requiresDependencyResolution test` or `@requiresDependencyCollection test` or none here?

Godin:
@marchof I had some time to carefully study some official Apache Maven reporting plugins, which have aggregation feature (maven-surefire-report-plugin, maven-pmd-plugin, maven-javadoc-plugin), as well as non official plugins (cobertura-maven-plugin, taglist-maven-plugin) - http://blog.the-future-group.com/2014/01/28/maven-aggregate-reports-multi-module-projects/

And seems that finally understood principle - key point is "Inheritance v. Aggregation" ( https://maven.apache.org/pom.html#Inheritance_v ) and in respect to reporting it is pretty well described in http://stackoverflow.com/a/21590120/244993 : if aggregator project is not parent, then it will be the last project in the build, and mentioned above plugins take advantage of this to guarantee availability of required materials for report generation. So that what is described on page https://github.com/jacoco/jacoco/wiki/MavenMultiModule#strategy-aggregator seems to be about situation, when aggregator project is used also as parent project. And indeed in such case his goals should be executed before others to build it, since parent is required for other projects.

To me this looks as complementary feature/solution to the one implemented here, which seems able to address needs and concerns mentioned earlier on this subject by @velo , @johnoliver and others (generation of report for whole project without need to maintain list of dependencies). Maybe @rfscholte , @jwloka , @mfriedenhagen can correct me, if I misunderstood something or if there are some pitfalls (other than combining aggregator and parent) in this approach?

And in any case, if you don't have objections @marcof, I'm going to squash-merge this PR into master since it looks good and meets our needs.

marchof:
@Godin Thx for the pointers! I added them to the wiki page.
For now I would prefer to have the report-aggregate goal available. In future we can think about enhancing the report goal.

+1 for squash-merge

marchof:
@Godin Thanks!

mblub:
@marchof  @Godin 

This is fantastic, worked great for me!

Used 0.7.7-SNAPSHOT as of today (nexus snapshot version 0.7.7-20160419.203207-13, which presumably includes commit c181f60ce08ec9b0a6f59a2391c33c41bca8f1c0).

Thanks for adding this valuable item.  When do you plan to produce a JaCoCo release that includes this commit?

marchof:
@mblub We do some cleanup and will probably release end of this month.

scottcarey:
"If B are integration tests users probably want to see coverage on A. If B is plain unit tests I would prefer to only show code coverage on B."

Note that there are cases where a unit test MUST be in a different module, due to corner case interactions between test scope and non-test scope transitive dependencies.

Godin:
@scottcarey as you can see in GitHub Web UI on the right hand-side - this ticket is closed and released as part of version 0.7.7. So please direct questions or suggestions about this subject, if there are any, to our mailing list ( https://groups.google.com/forum/#!forum/jacoco ) instead of commenting here.

Thank you for understanding.

PR #389, opened 2016-03-22
Test the fix on multi thread environment

velo:


marchof:
At least for 2 test runs the build succeeded and the report looks as expected.

velo:
hrmmm, back to the planning board... I will need to download the sources and take a look at ITs more carefully

PR #390, opened 2016-03-26
Report

velo:
So, I made an alternative fix for the report-aggregation goal...

### DIFF
--- a/jacoco-maven-plugin.test/.project
+++ b/jacoco-maven-plugin.test/.project
@@ -6,12 +6,18 @@
 	</projects>
 	<buildSpec>
 		<buildCommand>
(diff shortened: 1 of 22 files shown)


marchof:
@velo Thanks for this alternative approach! As said before I don't like the idea to maintain a solution which is not recommended by Maven developers, see this comment:

https://github.com/jacoco/jacoco/pull/388#issuecomment-200512340

Also @johnoliver proposed an improvement to #388 which allows working without report projects (but with extra goal on command line): https://github.com/jacoco/jacoco/pull/388#issuecomment-201760506

marchof:
We will follow the approach in #388.

PR #397, opened 2016-04-15
Improve EOF handling for exec files

marchof:
As proposed here we should not ignore EOFExceptions in case exec files are truncated somewhere in the middle: https://groups.google.com/d/msg/jacoco/qby-WuQp194/iFnx2fAHBAAJ

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.fail;
 
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.EOFException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


bjkail:
The change to ExecutionDataReader.readBlock (and the ExecutionDataWriter constants) break the API in a non-internal package, so I suggest to revert those changes.  ExecutionDataReader.read can safely cast to byte after checking for -1.

Godin:
Sharp eye, thank you @bjkail

Indeed - cast to byte after check for `-1` is safe - quoting Javadoc as a proof:
> Reads the next byte of data from the input stream. The value byte is returned as an int in the range 0 to 255. If no byte is available because the end of the stream has been reached, the value -1 is returned.

I reworked code a bit to restore API compatibility. Please check, if all fine, then I will squash commits and push it into master.

Godin:
Also maybe we should reclassify this as "bug fix" but not as "non-functional change", at least this is a behavioural change since truncated files won't be silently accepted as it was before.

bjkail:
Looks good to me now, thanks.

marchof:
@bjkail @Godin Agree on all comments, Thx! I updated the commit.

PR #399, opened 2016-04-16
Fix compilation with JDK 9 >= b111

Godin:
Currently compilation of tests fails with JDK 9 >= b111.

@marchof do you have an ideas of how we can resolve this?

I can imagine following options:

1. start using mocking framework such as Mockito
2. conditional compilation of different mocks for pre JDK 9 and post JDK 9 using separate module
3. remove this test

### DIFF
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/CoverageTransformerTest.java
@@ -200,7 +200,7 @@ public class CoverageTransformerTest {
 	@Test
 	public void testTransformFiltered2() throws IllegalClassFormatException {
 		CoverageTransformer t = createTransformer();
-		assertNull(t.transform(null, "org.jacoco.Sample", null,
+		assertNull(t.transform((ClassLoader) null, "org.jacoco.Sample", null,
 				protectionDomain, new byte[0]));
 	}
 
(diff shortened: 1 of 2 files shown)


marchof:
@Godin Doesn't your commit fix the issue?

Godin:
@marchof unfortunately no - it breaks compilation with JDK < 9, because new method uses new `java.lang.reflect.Module` from JDK 9 as a parameter

marchof:
I see. There are two options two fix this:

1. Using java.lang.reflect.Proxy
2. Introducing Mockito

I would prefer the Mockito option. In this case I would also remove manual mocking at other places in a separate PR. 
If you agree I can modify the test case today.

Godin:
@marchof the latest stable release of Mockito 1.10.19 requires Java 5, however seems that upcoming version 2.x might require Java 6 due to dependency on byte-buddy. Not sure if this is blocker for us, but latest stable version might have troubles with JDK 8, upcoming 2.x promises to work fine with JDK 8, and no statements about JDK 9.

Also not sure that Mockito is the best choice in long run - there are some interesting comparisons with JMockit for example, which however also requires Java 6, but added support for JDK 9 recently.

In any case addition of dependency on mocking framework will impact maintainability, e.g. questions whether they indirectly pull asm as transitive dependency or not, and whether they work or not with our range of JDKs, are interesting ones.

From here I would prefer to handle story of usage of mocking framework separately.

And proxy will do the job just perfectly for this ticket. Don't know how I forgot about such approach, but I knew that you always have bright ideas :wink: I updated PR and it passes build for all JDKs, please check.

marchof:
Looks good to me: +1

(beside the test itself could need improvement, but that's another story)

Godin:
@marchof yeah, test looks crazy

PR #403, opened 2016-04-20
Consolidate aggregate reports in JaCoCo build

Godin:
With PR #388 we created a new bundle `org.jacoco.tests.coverage` for the aggregate report only. But we have more aggregate artifacts which are only bundled with the all-in-one ZIP:

* JUnit test results
* JavaDoc
* Mojo Docs

All should be integrated in one bundle. Either we move the coverage report back to `org.jacoco.doc` or move all to a new bundle `org.jacoco.build.reports`.

@Godin any opinion about this?


Godin:
@marchof indeed seems that this was the case. Can I ask you to double-check my changes?

Also while working on this I noticed one things, which bothers me a bit: previously with Ant-based generation of report it was possible to rebuild module "org.jacoco.doc" separately, now with #388 you can't rebuild it separately to get aggregated coverage report.

marchof:
Is rebuilding `org.jacoco.doc` separately a use case _we_ actually have? Or is that more of a _general_ limitation that we impose on JaCoCo users with the dependency based design of the `report-aggregate` report? On the other hand how can we reliably create an aggregated coverage report without executing the tests?

org.jacoco.doc/pom.xml
```diff
@@ -127,7 +188,7 @@
             <configuration>
               <reportOutputDirectory>${project.build.directory}/apidocs</reportOutputDirectory>
               <includeDependencySources>true</includeDependencySources>
-              <excludePackageNames>*.internal,org.jacoco.ant</excludePackageNames>
+              <excludePackageNames>*.internal,org.jacoco.ant,org.jacoco.maven</excludePackageNames>
```
marchof: We also should exclude `org.jacoco.examples` to not make it part of our API documentation.

Godin:
@marchof this is more of a _general limitation_ , which is easy to hit - complete rebuild of project is a bit inconvenient during tuning of aggregate report due to long try-fail cycle, as it was for me with `org.jacoco.doc`.

> how can we reliably create an aggregated coverage report without executing the tests?

How it was with Ant? :wink: During rebuild of `org.jacoco.doc` tests of its dependencies were already executed - all data (sources, classes, execs) already available.

Also probably more realistic and interesting scenario is:
```
A<-B<-C
 \ | /
 Report
```
rebuild of B, C and Report as
```
mvn --projects B --also-make-dependents clean install
```
will lead to report with data for "B" and "C", but not for "A".

marchof:
Hm, this means we're now talking about a general limitation of the dependency based approach of our new `report-agregate` goal?

Also I'm not sure whether I understand you example: The reactor will be B and A. From my limited understanding the report is not built at all?

Godin:
@marchof quoting `mvn --help` :
> --also-make-dependents If project list is specified, also build projects that depend on projects on the list

only C and Report depend on B.

org.jacoco.doc/pom.xml
Godin: Fixed

marchof:
@Godin Sorry, now I see, it works the other way round. So do you think re-using the dependency for aggregate reports is a bad idea? We could create a separate list property, but this would force users to maintain the lists redundantly. Or do you see any other solutions?

marchof:
+1 for Merge

Godin:
@marchof for report we need:
* class files: dependencies
* source files: maven-javadoc-plugin also needs them for javadoc generation and takes them from local repository - see `NOTE: If your configuration includes a dependency but that dependency's source artifact is unavailable, the javadoc plugin will fail` at page https://maven.apache.org/plugins/maven-javadoc-plugin/examples/aggregate-dependency-sources.html (i.e. it requires usage of maven-source-plugin)
* execution data: here we have a problem, but I can imagine workaround (which I believe was in one of contributed PRs btw) - allow to specify mask to search for exec files, e.g. "../**/target/jacoco.exec", but I'm not sure if this is the "Maven way". And this can be done for sources to get rid of the need to use maven-source-plugin. But then at the end this will not be at all different from Ant setup, which we had before :confused:

marchof:
@Godin I'm confused now: Are we talking about the way how we create the code coverage report for JaCoCo or are you questioning the implementation of the `report-aggregate` goal as such? In the latter case we should reopen #388 and maybe rollback the merge to master??

Godin:
@marchof I just dumped some thoughts in response to yours
> Or do you see any other solutions?

Excuse me if it created confusion.
I don't know if limitation of #388, which we discovered here, is critical or not.

marchof:
@Godin Having aggregated reports with the limitations regarding partial builds is still better than not having aggregated builds. Given the fact that it took us years to provide a solution I would move on with what we have.

PR #401, opened 2016-05-02
".resources" directory causes problems for github pages

marchof:
https://github.com/github/maven-plugins/issues/112

### Steps to reproduce

JaCoCo version: 0.7.6.201602180812
Operating system: Windows/gitshell 
Tool integration: Maven (github-site plugin)

### Expected behaviour

Reports display correctly on github site. 

### Actual behaviour

The .resources directory is not deployed. 

I have raised an issue with the github/maven-plugins project 
https://github.com/github/maven-plugins/issues/112
however the problem may well not be solvable by them. 

The root cause is that some software, arguably correctly, overlooked a directory starting with a dot. 

Your resources directory should not be a hidden directory. 

Please consider renaming it to resources.

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -27,6 +27,9 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/388">#388</a>).</li>
   <li>New parameters <code>title</code> and <code>footer</code> for Maven
       reporting goals allow customization of generated reports.</li>
+  <li>Renamed "dot" resources in generated HTML reports to become more web
+      hosting friendly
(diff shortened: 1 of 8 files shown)


marchof:
@Godin Here is a possible solution...

Godin:
@marchof LGTM :thumbsup: and this fixes problem with our own experimental deployment to GitHub pages - before this change all images at http://jacoco.github.io/jacoco/trunk/coverage/ were broken

timp:
:+1: thanks for speedy response.

agabrys:
Is there any chance to release 0.7.6.1 version which contains only this patch?

Godin:
@agabrys please wait for 0.7.7 or use snapshot version

PR #405, opened 2016-05-04
Fix build on Windows

marchof:
We should update the versions of our build plugins. I came accross this issue while trying to build JaCoCo on Windows:

https://issues.apache.org/jira/browse/MNG-5792

### DIFF
--- a/jacoco-maven-plugin.test/it/it-report-aggregate/verify.bsh
+++ b/jacoco-maven-plugin.test/it/it-report-aggregate/verify.bsh
@@ -13,15 +13,15 @@ import org.codehaus.plexus.util.*;
 
 String buildLog = FileUtils.fileRead( new File( basedir, "build.log" ) );
 
-if ( buildLog.indexOf( "/child1/target/jacoco.exec" ) < 0 ) {
+if ( buildLog.indexOf( "/child1/target/jacoco.exec".replace('/', File.separatorChar) ) < 0 ) {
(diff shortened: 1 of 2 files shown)


marchof:
@Godin Can you please review?

marchof:
@Godin Java 6 Build fails as it looks like Maven is also executed with Java 6. Some of the new Plug-ins use Java 6 Bytecode. Unlike Java 5 Build, which uses Java 7 to run Maven.

Godin:
@marchof ok, I'll check compatibilities

Godin:
@marchof Okay, I can reproduce original issue with build on Windows - Maven 3.3.1 indeed fails, while Maven 3.2.5 succeeds. Also seems that "it-report-aggregate" is platform-dependent and doesn't work on Windows. And you right - we can't blindly update plugins, if we want to preserve ability to start build using Java 6. I'll remediate step-by-step.

Godin:
@marchof I've put aside upgrade of all plugins and made only changes to fix build on Windows. Now it works on my side, but maybe you can check if it works for you? or you can just trust me :wink:

marchof:
@Godin Change looks good to me. I don't have windows machine at hand, so I'm afraid I have to trust you.

Godin:
@marchof for the safety net I'm adding CI build on Windows using AppVeyor - https://ci.appveyor.com/project/JaCoCo/JaCoCo

PR #400, opened 2016-05-13
Provide more context information when unable to read input during analysis

Godin:
Add location to `IOException` thrown from

1. `zip.getNextEntry()` in `analyzeZip`
2. `new GZIPInputStream(input)` in `analyzeGzip`
2. `Pack200Streams.unpack(input)` in `analyzePack200`


Godin:

marchof:
@Godin Looks good to me!

P.S.: Nice test cases for broken file formats. Looks like we need a (broken) file format mocking library ;-)

PR #406, opened 2016-05-14
Support Java 9 class file format

Godin:
https://bugs.openjdk.java.net/browse/JDK-8150011

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
@@ -1,0 +1,115 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+import org.objectweb.asm.Opcodes;
+
+/**
+ * Patching for Java 9 classes, so that ASM can read them.
+ */
+public final class Java9Support {
+
+	/**
+	 * Version of the Java 9 class file format.
+	 */
+	public static final int V1_9 = Opcodes.V1_8 + 1;
+
+	private Java9Support() {
+	}
+
+	/**
+	 * Reads all bytes from an input stream into a byte array.
+	 *
+	 * @param is
+	 *             the input stream to read from
+	 * @return a byte array containing all the bytes from the stream
+	 * @throws IOException
+	 *             if an I/O error occurs
+	 */
+	public static byte[] readFully(final InputStream is)
+			throws IOException {
+		if (is == null) {
+			throw new IllegalArgumentException();
+		}
+		final byte[] buf = new byte[1024];
+		final ByteArrayOutputStream out = new ByteArrayOutputStream();
+		while (true) {
+			int r = is.read(buf);
+			if (r == -1) {
+				break;
+			}
+			out.write(buf, 0, r);
+		}
+		return out.toByteArray();
+	}
+
+	private static void putShort(byte[] b, int index, int s) {
+		b[index] = (byte) (s >>> 8);
+		b[index + 1] = (byte) s;
+	}
+
+	private static short readShort(byte[] b, int index) {
+		return (short) (((b[index] & 0xFF) << 8) | (b[index + 1] & 0xFF));
+	}
+
+	/**
+	 * Determines whether class definition contains {@link #V1_9} version.
+	 *
+	 * @param buffer
+	 *             definition of the class
+	 * @return <code>true</code> if class definition contains Java 9 version
+	 */
+	public static boolean isPatchRequired(byte[] buffer) {
+		return readShort(buffer, 6) == V1_9;
+	}
+
+	/**
+	 * Returns new definition of class with version {@link Opcodes#V1_8},
+	 * if it has version {@link #V1_9}.
+	 *
+	 * @param buffer
+	 *             definition of the class
+	 * @return new definition of the class
+	 */
+	public static byte[] downgradeIfRequired(byte[] buffer) {
+		return isPatchRequired(buffer) ? downgrade(buffer) : buffer;
+	}
+
+	/**
+	 * Replaces version in the definition of class on {@link Opcodes#V1_8}.
+	 *
+	 * @param b
+	 *             definition of the class
+	 * @return new definition of the class
+	 */
+	public static byte[] downgrade(byte[] b) {
+		byte[] result = new byte[b.length];
+		System.arraycopy(b, 0, result, 0, b.length);
+		putShort(result, 6, Opcodes.V1_8);
+		return result;
+	}
+
+	/**
+	 * Replaces version in the definition of class on {@link #V1_9}.
+	 *
+	 * @param b
+	 *             definition of the class
+	 */
+	public static void upgrade(byte[] b) {
+		putShort(b, 6, V1_9);
+	}
+
+}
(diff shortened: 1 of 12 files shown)


Godin:
@marchof could you please review? Note that this is not sufficient to be able to build JaCoCo with compilation into new class files, because maven-shade-plugin also requires update. However this allows JaCoCo to work properly with new class files - I added integration test (instrumentation with agent + report generation on new class files) and tested with local dev build of latest OpenJDK 9.

org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
```diff
@@ -0,0 +1,98 @@
...
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.Opcodes;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * Patching for Java 9 classes, so that ASM can read them.
+ */
+public final class Java9Support {
```
marchof: I see several JavaDoc warnings on this class. Even if the code is temporary I would prefer if our code comes without warnings.

org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
```diff
@@ -0,0 +1,98 @@
...
+	/**
+	 * Copy of {@link ClassReader#readClass(InputStream, boolean)}.
+	 */
+	public static byte[] readClass(final InputStream is, boolean close)
+			throws IOException {
+		if (is == null) {
+			throw new IOException("Class not found");
+		}
+		try {
+			byte[] b = new byte[is.available()];
```
marchof: This code has quite some complexity. We should either add tests or use more high level APIs. E.g. the input can be easily copied into a ByteArrayOutputStream which then most of the copy magic for us (needs a little bit more memory allocations though).

org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
Godin: @marchof I suppose that those warnings are generated by Eclipse? I don't see warnings in IntelliJ , even when explicitly executing all built-in inspections for Javadocs. So could you please elaborate what needs to be fixed?

marchof: Basically we require to have JavaDoc on all public Java classes and members.
* Javadoc: Missing comment for public declaration	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 25	Java Problem
* Javadoc: The method readClass(InputStream, boolean) from the type ClassReader is not visible	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 31	Java Problem
* Javadoc: Missing tag for return type	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 33	Java Problem
* Javadoc: Missing tag for parameter is	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 33	Java Problem
* Javadoc: Missing tag for parameter close	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 33	Java Problem
* Javadoc: Missing tag for declared exception IOException	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 34	Java Problem
* Javadoc: Missing comment for public declaration	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 79	Java Problem
* Javadoc: Missing comment for public declaration	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 83	Java Problem
* Javadoc: Missing comment for public declaration	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 87	Java Problem
* Javadoc: Missing comment for public declaration	Java9Support.java	/org.jacoco.core/src/org/jacoco/core/internal	line 94	Java Problem

marchof:
I wonder whether we should drop `Analyzer.analyzeClass(ClassReader)` from the public API? This is a asymmetry to Instrumenter anyways.

marchof:
@Godin Beside my comments I'm ok with this PR!

Godin:
@marchof I simplified code and added Javadocs.

About asymmetry: indeed it bothers. This is the only place where we expose ClassReader, but not the only place where we expose ASM APIs. Removal - is an API change. I don't think that it is critical for experimental support of Java 9 class file format. So IMO we should consider separately question about marking it as deprecated and further removal. WDYT?

marchof:
@Godin I agree to remove this API in a separate request. I just came across this as it won't work for Java 9.

bjkail:
FYI (sorry if this is the wrong forum for this comment), we override analyzeClass(ClassReader) in order to filter which classes should be analyzed: for historical reasons, module boundaries in our product are not cleanly separated by archive, so we use package filters here.  It would be ideal if there were still some way for us to do that efficiently before you remove that method (we can also skip the ICoverageVisitor.visitCoverage event, but that's less efficient since Analyzer's ASM processing still occurs).

Godin:
@bjkail No need to apologize. In fact thank you for this valuable information.

As far as I understood you're doing something like
```
Analyzer analyzer = new Analyzer(...);
ClassReader classReader = new ClassReader(...);
if (notFiltered(reader)) {
  analyzer.analyzeClass(reader);
}
```
if so, wondering whether you can replace it by
```
Analyzer analyzer = new Analyzer(...);
ClassReader classReader = new ClassReader(...);
if (notFiltered(reader)) {
  analyzer.analyzeClass(reader.b, "");
}
```
?

bjkail:
Well, more like this:

    Analyzer analyzer = new Analyzer(...) {
      @Override
      public void analyzeClass(ClassReader r) {
        if (notFiltered(r.getClassName())) {
          super.analyzeClass(r);
        }
      }
    };
    
    // Use Analyzer's existing recursive archive processing logic.
    analyzer.analyzeAll(...);

Regardless, we could instead override `analyzeClass(InputStream, String)` to do as you suggest at the cost of a duplicate ClassReader constructor call, but I suppose that's relatively lightweight when compared to overriding visitExecution.  Well, the Analyzer javadoc doesn't actually guarantee that analyzeAll will invoke the analyzeClass methods, but it is nice to be able to reuse Analyzer's recursive archive processing, so it would be nice if the javadoc did make that guarantee.

marchof:
For the filtering API story we will need proper call-backs anyways.

Godin:
JDK 9 EA build 119 is available for testing, so let's go ahead with merging of those changes. And let's handle raised points separately from this PR, if it is needed or would be needed.

And for the record without those changes execution of JaCoCo with JDK 9 EA b 119 fails with message like
```
Exception in thread "main" java.lang.reflect.InvocationTargetException
     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(java.base@9-ea/Native Method)
     at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(java.base@9-ea/NativeMethodAccessorImpl.java:62)
     at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@9-ea/DelegatingMethodAccessorImpl.java:43)
     at java.lang.reflect.Method.invoke(java.base@9-ea/Method.java:531)
     at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(java.instrument@9-ea/InstrumentationImpl.java:396)
     at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(java.instrument@9-ea/InstrumentationImpl.java:408)
Caused by: java.lang.RuntimeException: Class java/util/UUID could not be instrumented.
     at org.jacoco.agent.rt.internal_9472e99.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:139)
     at org.jacoco.agent.rt.internal_9472e99.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:100)
     at org.jacoco.agent.rt.internal_9472e99.PreMain.createRuntime(PreMain.java:55)
     at org.jacoco.agent.rt.internal_9472e99.PreMain.premain(PreMain.java:47)
     ... 6 more
Caused by: java.lang.NoSuchFieldException: $jacocoAccess
     at java.lang.Class.getField(java.base@9-ea/Class.java:1881)
     at org.jacoco.agent.rt.internal_9472e99.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:137)
     ... 9 more
```
which could be improved - something like "cannot read class".

PR #413, opened 2016-05-20
Include the current project in aggregation as well as dependencies.

timp:
For non-multi projects this combines the integration and unit tests.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
@@ -103,6 +103,9 @@ public class ReportAggregateMojo extends AbstractReportMojo {
 	void loadExecutionData(final ReportSupport support) throws IOException {
 		final FileFilter filter = new FileFilter(dataFileIncludes,
 				dataFileExcludes);
+		for (final File execFile : filter.getFiles(getProject().getBasedir())) {
+			support.loadExecutionData(execFile);
+		}
 		for (final MavenProject dependency : findDependencies(
 				Artifact.SCOPE_COMPILE, Artifact.SCOPE_TEST)) {
 			for (final File execFile : filter.getFiles(dependency.getBasedir())) {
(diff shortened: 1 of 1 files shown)


timp:
Updated to only aggregate if more than one exec file.

timp:
See http://timp.github.io/juice/ for use case. 
Also tested with IT tests skipped, giving UT report only.

marchof:
Not sure whether we should mix this use case into the report-aggregate goal, aggregate was designed for multiple modules:

* The report creates a separate group for each module which doesn't make sense for single module
* Real life project will have dependencies which would then show-up in the report anyways

timp:
@marchof 
See  http://timp.github.io/juice/project-reports.html the grouping looks fine . 

There will often be no change for multi module projects, if they are designed, as the plugin currently assumes, with no tests at the reporting level. If there are tests at the reporting level will they be included, which is an improvement. 

This both fixes a current broken snapshot and introduces new needed functionality: I tried for a long time to find another way to report upon the combined coverage of unit and integration tests using the merge functionality, with no success.

marchof:
@timp Nope, it creates a group for every project:

http://timp.github.io/juice/jacoco-aggregate/index.html

Doesn't make sense for single project reports.

timp:
Hi @marchof thanks for sticking with this. 
I have realised that I have mis-understood your use of aggregation. I assumed it meant aggregating the coverage of sub-modules but I now understand that it is aggregation of _dependency_ coverage (as well?). 

I am now applying JaCoCo to a larger multi-module project, so far with no success. Could you point to an example project where the aggregation goal works as designed? 

I am afraid I do not understand why http://timp.github.io/juice/jacoco-aggregate/index.html does not make sense for single project reports. 
For webapps it is very common to produce two coverage reports, one for unit tests and one for integration tests and it is the combination of these which is of most interest.

marchof:
@timp Please see documentation: http://eclemma.org/jacoco/trunk/doc/report-aggregate-mojo.html

> Creates a structured code coverage report (HTML, XML, and CSV) from multiple projects within reactor. The report is created from all modules this project depends on.

The primary purpose of the `report-aggregate` is to create a structured report for multiple modules with execution data combined also from multiple reports. The existing reporting goals only work on single modules.

Also with the dataFileIncludes parameter you can collect executions data from different tests. This might be a feature we can add to the single module reporting goals in future. As of today you can use the merge goal to create a single module report for different tests.

Here is an example for the `report-aggregate` goal: https://github.com/jacoco/jacoco/tree/master/jacoco-maven-plugin.test/it/it-report-aggregate

Godin:
@timp take a look on our integration tests for Maven Plugin - https://github.com/jacoco/jacoco/tree/master/jacoco-maven-plugin.test/it/it-report-aggregate also JaCoCo build itself uses "aggregate" goal - https://github.com/jacoco/jacoco/blob/master/org.jacoco.doc/pom.xml#L120

timp:
Thanks for the pointers, sorry for the noise.

Issue #194, opened 2014-02-18
Revoke support for Maven 2.2.*


mfriedenhagen:
As decided on the Maven developer list, Maven 2.2.\* has reached [end-of-life](http://mail-archives.apache.org/mod_mbox/maven-dev/201402.mbox/%3CCA%2BnPnMzPNhkBqbj-Uz045Gt4j8%3DCiXBFCwnNGB0NZ38j-X%2BUvA%40mail.gmail.com%3E).

We may choose between 2 strategies:
1. Just do not fix any Maven 2.2.\* related issues and do not test jacoco-maven-plugin with this version anymore.
2. Go forward and make Maven 3.\* a prerequisite for building jacoco _and_ running the `jacoco-maven-plugin` _and_ remove any workarounds and documentation for running jacoco with Maven 2.2.*

@Godin, @marchof: your oppinions?

marchof:
Should be asked on the mailing list whether there are any objections.
If we keep Maven 2.x we should least run integration tests.

Godin:
Should be noted that support of Maven 2.x has been removed in SonarQube 5.0.

Godin:
Another note is that Apache Maven Plugins are massively start to require 3.0 - http://mail-archives.apache.org/mod_mbox/maven-dev/201503.mbox/%3C20150320213920.6AE4B17423@minotaur.apache.org%3E

khmarbaise:
Sorry to say that, but this is not true. See here: https://builds.apache.org/view/All/job/dist-tool-plugin/site/dist-tool-prerequisites.html. At the moment only  two plugins (maven-shade-plugin, maven-scm-publish-plugin) need Maven 3.X...and two plugins need Java 1.6 instead of Java 1.5 (maven-checkstyle-plugin, maven-pmd-plugin). (if you need Maven 2.2.1 support take a look here: http://maven.apache.org/maven-2.x-eol.html). What you are referencing is the plan we have...

khmarbaise:
I would vote for leaving Maven 2 behind but would suggest to make a version dump for example to 1.0.0 to make clear something important has happened (Maven 3.X requirement).

Godin:
Indeed, I was referring to plan:
> The next step on our roadmap is to lift all plugin versions to 3.0.0 to make clear those plugins will only work with Maven 3.0+ furthermore the Java minimum requirement will be lifted to JDK 1.6 as well.

to make our own plan ;) but thanks for correction @khmarbaise

[...]

PR #412, opened 2016-05-20
Use Maven Plugin Tools Java Annotations

Godin:
Currently with JDK 9 EA creation of javadoc for jacoco-maven-plugin fails with messages like `unknown tag: requiresProject`.

This problem should vanish, if we'll use annotations - http://maven.apache.org/plugin-tools/maven-plugin-plugin/examples/using-annotations.html And anyway this is a good modernization of code.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/DumpMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/DumpMojo.java
@@ -18,6 +18,9 @@ import java.io.IOException;
 import java.net.InetAddress;
 
 import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
 import org.jacoco.core.tools.ExecDumpClient;
 import org.jacoco.core.tools.ExecFileLoader;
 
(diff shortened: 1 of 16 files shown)


marchof:
@Godin This would mean dropping Maven 2.x (see #194)?

Godin:
@marchof no, it is not connected, I just checked - those changes don't affect compatibility with Maven. I suppose that you asked because of misleading statement on page http://maven.apache.org/plugin-tools/maven-plugin-plugin/examples/using-annotations.html :

> Since version 3.0, you can use Java5 annotations to generate the plugin descriptor file.

it is about about version of Maven Plugin Tools, not about version of Maven - https://issues.apache.org/jira/browse/MPLUGIN-189

And so btw, could you please review changes?

Note that while report generation don't fail with JDK 9 EA anymore, style of generated javadoc is broken, but this is another story.

khmarbaise:
First i have created an appropriate [JIRA issue](https://issues.apache.org/jira/browse/MPLUGIN-301) for the misleading text. Furthermore is leaving Maven 2.X behind really an issue ?

marchof:
@khmarbaise Thx for clarification! Maven 2.x isn't really an issue -- but a separate one: #194

jacoco-maven-plugin/src/org/jacoco/maven/DumpMojo.java
```diff
@@ -33,66 +36,81 @@
...
  * 
- * @goal dump
- * @phase post-integration-test
- * @threadSafe
  * @since 0.6.4
  */
+@Mojo(
+		name = "dump",
+		defaultPhase = LifecyclePhase.POST_INTEGRATION_TEST,
+		requiresProject = false,
```
marchof: This different from what we had before. But intentional, right?

marchof:
@Godin The checked-in Eclipse settings still have auto-format on save configured (And I actuall really like that feature). So annotations will be formatted differently on next save with Eclipse (one line).

Beside that change looks good to me!

jacoco-maven-plugin/src/org/jacoco/maven/DumpMojo.java
Godin: @marchof not intentional, my mistake actually - I thought that if javadoc `@requiresReports` is missing, then it is `false` by default, but in fact it is opposite. I'll fix this.

Godin:
@marchof strange - I used Eclipse Code Formatter plugin for IntelliJ IDEA

Godin:
@marchof reformatted, thx for review, will merge.

PR #300, opened 2016-05-22
Build with JDK 8 should not require additional arguments

marchof:
Currently "-Dbytecode.version=1.8" should be specified to run build using JDK 8, whereas would be better to use just "mvn clean install".

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,7 +11,7 @@ install:
 
 build_script:
   # Maven 3.3.9 requires Java >= 7, but generation of Javadocs requires Java <= 6 (https://github.com/jacoco/jacoco/issues/110)
-  - mvn -V -B -e verify -Pno-java8-validation -Djdk.version=1.6 --toolchains=.travis\appveyor-toolchains.xml
+  - mvn -V -B -e verify -Djdk.version=1.6 --toolchains=.travis\appveyor-toolchains.xml
 
 artifacts:
   - path: jacoco\target\*.zip
--- a/org.jacoco.core.test/.classpath
+++ b/org.jacoco.core.test/.classpath
@@ -10,7 +10,7 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry excluding="org/jacoco/core/test/validation/java8/*.java" including="**/*.java" kind="src" output="target/classes" path="src">
+	<classpathentry kind="src" output="target/classes" path="src">
 		<attributes>
 			<attribute name="optional" value="true"/>
 			<attribute name="maven.pomderived" value="true"/>
--- a/org.jacoco.core.test/pom.xml
+++ b/org.jacoco.core.test/pom.xml
@@ -40,24 +40,67 @@
   
   <profiles>
     <profile>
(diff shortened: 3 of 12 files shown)


marchof:
@Godin Here is my proposal to solve this. It should work for multiple source file versions and JDK versions.

Godin:
@marchof IMO ideally these tests should be moved into separate module, but this solution works, so fine to go ahead with it.

Also for JDK 9 it is a little bit more complex because `-target 1.5` not available in it, but this is another story as it wasn't released.

I'll remove activation of removed profile "no-java8-validation" in script for Appveyor and will merge after that.

timp:
Building from master with 
```
$ mvn -version
Apache Maven 3.3.9
Maven home: /usr/share/maven
Java version: 1.8.0_91, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en_GB, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-22-generic", arch: "amd64", family: "unix"
```
using 
```
mvn clean install
```
hangs with
```
testUnknownHost:
[jacoco:dump] Connecting to nosuchhost/92.242.132.15:6300
[jacoco:dump] Connection timed out
[jacoco:dump] Connecting to nosuchhost/92.242.132.15:6300
[jacoco:dump] Connection timed out
[jacoco:dump] Connecting to nosuchhost/92.242.132.15:6300
```
(I killed process eventually)

marchof:
@timp Please consider using our mailing list for questions. This issue is closed and unrelated, thx!

Obviously "nosuchhost" resolves to an IP address on your system. Please check your DNS settings.

timp:
Sorry, again, for the noise. This is caused by my ISP, British Telecom, see https://community.bt.com/t5/Connected-Devices-Other/92-242-132-15-Barefruit-Ltd/td-p/1062668

Godin:
@timp first of all, as Marc said - this is unrelated to original issue described in this ticket, which is already closed, so better to be discussed separately, preferably in a mailing list.

You problem is caused by your ISP, so excuse me, but it is completely unclear (at least for me) what is expected from us? Provide you a new ISP? It sounds like "build is failing, because of broken hard drive". 😆 

Jokes aside: while content in your link is hard to understand, it contains statements `You can opt out using this link ...`. Did you tried this?

Looking deeper on this integration test - it actually works a kind of as expected:
in case of connection problem JaCoCo Ant DumpTask tries to establish connection up to 10 retries and for each retry shows reason of failed connection. Under normal conditions reason of failed connection will be inability to resolve hostname using DNS. Under conditions of your ISP - my best guess that behavior is the same as in case
```
$ cat /etc/hosts | grep nosuchhost
127.0.0.2 nosuchhost
$ iptables -A INPUT -p tcp -d 127.0.0.2 --dport 6300 -j DROP
```
so that hostname resolution succeeds, but connection fails without TCP response. Usage of `REJECT` instead of `DROP` would have caused response. And thus test is not stuck - its long running time is dictated by a native socket timeout (non Java) in a native code, which is OS-dependent.

So that at the end IMO it is highly questionable whether we should do something about this or not.

PR #415, opened 2016-05-24
Property dataFileIncludes in report-aggregate is not changeable by pom due to known maven bug

Godin:
JaCoCo version: 0.7.7-SNAPSHOT
Operating system: Windows
Tool integration: Maven

The property dataFileIncludes in the report-aggregate can't be set by the pom file due to a known bug in maven:
[see here](https://issues.apache.org/jira/browse/MNG-5440?jql=project%20%3D%20MNG%20AND%20resolution%20%3D%20Unresolved%20AND%20text%20~%20%22%40Parameter%22%20ORDER%20BY%20priority%20DESC)

To fix this by a workaround, the default-value has to been removed from the parameter and access to it should be done by a getter which will set the default if the value is always null

### DIFF
--- a/jacoco-maven-plugin.test/it/it-report-aggregate-customization/child1/pom.xml
+++ b/jacoco-maven-plugin.test/it/it-report-aggregate-customization/child1/pom.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+   All rights reserved. This program and the accompanying materials
+   are made available under the terms of the Eclipse Public License v1.0
+   which accompanies this distribution, and is available at
+   http://www.eclipse.org/legal/epl-v10.html
+
+   Contributors:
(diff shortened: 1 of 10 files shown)


Godin:
@kubachmann thank you for finding this.

marchof:
@Godin Ok to merge.

I only wonder whether we should have a test for this?

Godin:
@marchof I didn't wanted to duplicate existing test and didn't managed to imagine how to update it to cover both default and custom values, so initially gave up on test. Now updated PR with new test with all customizations for `report-aggregate`.

marchof:
@Godin Thx for taking the time to actually creating a test case! I would simplify the setup though and completely remove module child 2 here. This probably nor required to demonstrate the issue?

Godin:
@marchof indeed - module `child2` is not required for the non-regression test for this case, however this test verifies all customizations of properties - that custom values for `dataFileExcludes` and `outputDirectory` are taken into account

marchof:
@Godin Of course you're right. Now I see the use case for child2. Sorry.

marchof:
So, looks good to me!

PR #414, opened 2016-05-24
maven goal report-aggregate doesn't use it's own measurement 

marchof:
JaCoCo version: 0.7.7-SNAPSHOT
Operating system: Windows
Tool integration: Maven

The maven goal report-aggregate doesn't read the exec file in the executing project.
When an integration-test module IT has collected the coverage data and stored the result to IT/target/jacoco.exec, the report generation by report-aggregate reads all the jacoco.exec files of all dependant projects in the reactor but not the file in the IT module itself.

Since a module isn't dependant by itself in maven, in the method findDependencies there should be the statement
`result.add(this.project);`

Another solution could be, to redesign this goal to walk not thru the depends but thru the submodules like aggregated reports of other tools, so the report will be on the top level of the generated site.


marchof:
@kubachmann @Godin Please review.

kubachmann:
Works for me. Result is as expected.

marchof:
@Godin Can you please also check and merge if it looks ok for you?

Godin:
@marchof I'm doing this right now 😉

Godin:
LGTM :thumbsup:

PR #177, opened 2016-05-27
Frames are re-calculated for large methods

Godin:
JaCoCo cannot re-calculate stackmap frames as this is an expensive operation and would require knowing the type hierarchy (which is not available). Also ASM default implementation for the type hierarchy actually loads(!) the classes.

Therefore in JaCoCo special care has been taken to incrementally update the stackmap frames during instrumentation.

It has been reported that the stackmap frame are re-calculated anyways is special situations, stack trace:

```
java.lang.ClassNotFoundException: javax.servlet.jsp.tagext.BodyContent
         at org.jacoco.asm.ClassWriter.getCommonSuperClass(Unknown Source)
         at org.jacoco.asm.ClassWriter.a(Unknown Source)
         at org.jacoco.asm.Frame.a(Unknown Source)
         at org.jacoco.asm.Frame.a(Unknown Source)
         at org.jacoco.asm.MethodWriter.visitMaxs(Unknown Source)
         at org.jacoco.asm.ClassReader.a(Unknown Source)
         at org.jacoco.asm.ClassReader.b(Unknown Source)
         at org.jacoco.asm.ClassReader.accept(Unknown Source)
         at org.jacoco.asm.ClassReader.accept(Unknown Source)
         at org.jacoco.asm.ClassWriter.toByteArray(Unknown Source)
         at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:79)
         at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:139) 
```

Looking at ASM source code it turned out that ASM triggers a frame re-calculation if the methods exceeds a certain size. See MethodWriter.resizeInstructions(). Shit.

Corresponding ASM issue: http://forge.ow2.org/tracker/?func=detail&aid=317551&group_id=23&atid=100023

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
@@ -11,7 +11,7 @@
  *******************************************************************************/
 package org.jacoco.core.test.validation;
 
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
 
 import org.jacoco.core.instr.Instrumenter;
 import org.jacoco.core.runtime.IRuntime;
@@ -70,7 +70,7 @@ public class ResizeInstructionsTest {
 		mv.visitEnd();
 		cw.visitEnd();
 		final byte[] original = cw.toByteArray();
-		assertEquals(true, computedCommonSuperClass);
+		assertTrue(computedCommonSuperClass);
 
 		load(className, original);
 		instrumenter.instrument(original, className);
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
@@ -113,8 +113,8 @@ public class ResizeInstructionsTest {
 	}
 
 	/**
-	 * Adds code that triggers invocation of
-	 * {@link org.objectweb.asm.MethodWriter#resizeInstructions()} during
+	 * Adds code that triggers usage of
+	 * {@link org.objectweb.asm.MethodWriter#INSERTED_FRAMES} during
 	 * instrumentation.
 	 */
 	private static void addCauseOfResizeInstructions(final MethodVisitor mv) {
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
@@ -17,6 +17,7 @@ import org.jacoco.core.instr.Instrumenter;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.RuntimeData;
 import org.jacoco.core.runtime.SystemPropertiesRuntime;
+import org.jacoco.core.test.TargetLoader;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -71,23 +72,11 @@ public class ResizeInstructionsTest {
 		cw.visitEnd();
 		final byte[] original = cw.toByteArray();
 		assertTrue(computedCommonSuperClass);
+		new TargetLoader().add(className, original);
 
-		load(className, original);
-		instrumenter.instrument(original, className);
-	}
-
-	private static void load(final String className, final byte[] bytes)
-			throws ClassNotFoundException {
-		new ClassLoader() {
-			@Override
-			protected Class<?> loadClass(final String name,
-					final boolean resolve) throws ClassNotFoundException {
-				if (name.equals(className)) {
-					return defineClass(name, bytes, 0, bytes.length);
-				}
-				return super.loadClass(name, resolve);
-			}
-		}.loadClass(className);
+		final byte[] instrumented = instrumenter.instrument(original,
+				className);
+		new TargetLoader().add(className, instrumented);
 	}
 
 	/**
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation;
+
+import static org.junit.Assert.assertEquals;
+
+import org.jacoco.core.instr.Instrumenter;
+import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.RuntimeData;
+import org.jacoco.core.runtime.SystemPropertiesRuntime;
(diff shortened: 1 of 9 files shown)


Godin:
@marchof reproducer attached.

It reproduces problem with released ASM 5.0.4 and with r1773, which is base for patch  http://forge.ow2.org/tracker/download.php/23/100023/317551/3773/bugfix317551.patch from http://forge.ow2.org/tracker/?func=detail&aid=317551&group_id=23&atid=100023 But unfortunately even construction of a class for reproducer fails on r1773 with patch:
```
java.lang.NullPointerException
        at org.objectweb.asm.Frame.push(Frame.java:691)
        at org.objectweb.asm.Frame.execute(Frame.java:978)
        at org.objectweb.asm.CurrentFrame.execute(CurrentFrame.java:50)
        at org.objectweb.asm.MethodWriter.visitInsn(MethodWriter.java:746)
        at org.objectweb.asm.ClassReader.readCode(ClassReader.java:1363)
        at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1032)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:708)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:521)
        at org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:1005)
        at org.jacoco.core.test.validation.ResizeInstructionsTest.createClass(ResizeInstructionsTest.java:115)
```

So that if reproducer looks correct for you, then I suppose we can report back that patch is not working.

marchof:
@Godin Now I remember that I came to similar point ;-)

http://forge.ow2.org/tracker/?func=detail&aid=317555&group_id=23&atid=100023

I somehow failed to attach my reproducer at that time (be aware: ow2 tools are tricky)!

Godin:
@marchof :laughing: okay, so that patch actually should be applied on top of r1776 , let me try

Godin:
@marchof same exception as before, which is btw different from the one you faced and contains traces of a patch - `org.objectweb.asm.CurrentFrame` is new file from patch

Godin:
@marchof hmmm, however it works, when generated `Short.MAX_VALUE - 7` NOPs instead of `Short.MAX_VALUE`

Godin:
and `Short.MAX_VALUE - 3` causes similar NPE during instrumentation, but not during construction of class:
```
Caused by: java.lang.NullPointerException
	at org.objectweb.asm.Frame.push(Frame.java:691)
	at org.objectweb.asm.Frame.push(Frame.java:710)
	at org.objectweb.asm.Frame.execute(Frame.java:1308)
	at org.objectweb.asm.CurrentFrame.execute(CurrentFrame.java:50)
	at org.objectweb.asm.MethodWriter.visitMethodInsn(MethodWriter.java:917)
	at org.objectweb.asm.ClassReader.readCode(ClassReader.java:1492)
	at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1032)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:708)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:521)
	at org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:1005)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:85)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:108)
```

Godin:
which is understandable - lower values do not cause resizing

marchof:
Just to double check: Have you tried to load the generated class into a JVM to make sure it is valid? ASM totally relies on valid usage of its APIs and valid input files. Most invalid usages of ASM will result in internal errors.

Godin:
@marchof see commit with added loading - works fine, when class is generated with ASM 5.0.4

marchof:
So this seems to be an issue with the patch which should be reported to ASM.

Godin:
@marchof NPE happens in a new code for resizing during processing of a first instruction of a method - `ICONST_0`, so that I have strong feeling that there is something fishy in implementation of new `CurrentFrame` and its instantiation in `MethodWriter` as it was added exactly for the case of first instruction:
```
// This should happen only once, for the implicit first frame
// (which is explicitly visited in ClassReader if the
// EXPAND_ASM_INSNS option is used).
currentBlock.frame = new CurrentFrame();
```

Godin:
I'm wondering - does it mean that case of resizing is not covered by tests in ASM? They don't use JaCoCo! :wink:

Godin:
@marchof reported - http://forge.ow2.org/tracker/index.php?func=detail&aid=317630&group_id=23&atid=100023

sudhanshusr:
@peter-fu Hi peter, you found any solution to exclude Routes from instrumentaion.? I am also facing the same problem please let me know if you get any lead. thanks

peter-fu:
No, I spitted routes instead.

https://github.com/sbt/jacoco4sbt/issues/44 is still open, so

Godin:
Meanwhile maybe we should override `getCommonSuperClass` to provide more meaningful error message? And maybe even completely forbidding recalculation of frames and unexpected classloading.

@marchof WDYT?

marchof:
@Godin I wouldn't introduce a exception here until we have a fixed ASM version:

1. At least for "app classpath only" the current setup works
2. We already add context information to any exception during instrumentation

Godin:
For the record: http://forge.ow2.org/tracker/?func=detail&aid=317630&group_id=23&atid=100023 provides new patch, which passes our reproducer.

Godin:
And fix went into current trunk of ASM, so that we just need to wait their release.

Godin:
@marchof could you please review? Similarly to https://github.com/jacoco/jacoco/pull/462#issuecomment-269626527 test uses a bit of bytecode generation.

bjkail:
@Godin I don't understand the `assertEquals(true, computedCommonSuperClass);`.  Per [marchof's comment](https://github.com/jacoco/jacoco/pull/177#issuecomment-221664022), we want to assert that the new ASM does not call `getCommonSuperclass` due to an internal `resizeInstructions`.

Godin:
@bjkail construction of original class happens with `COMPUTE_FRAMES` and this assertion ensures that method is called during construction of original class, i.e. that reproducer makes sense. After that during instrumentation there is no `ClassNotFoundException`, i.e. that upgrade of ASM library solves the issue and there is no computation of frames during instrumentation.

marchof:
@Godin Not sure how this test asserts that `getCommonSuperClass()` is not called during instrumentation? `java/lang/Integer` should load without any problems.

I think we should now add an assertion in Instrumenter that forbids execution of `getCommonSuperClass()` at all.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
```diff
@@ -0,0 +1,131 @@
...
+		final byte[] original = cw.toByteArray();
+		assertEquals(true, computedCommonSuperClass);
+
+		load(className, original);
+		instrumenter.instrument(original, className);
+	}
+
+	private static void load(final String className, final byte[] bytes)
+			throws ClassNotFoundException {
+		new ClassLoader() {
```
marchof: Use existing `TargetLoader` instead?

org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
```diff
@@ -0,0 +1,131 @@
...
+		final MethodVisitor mv = cw.visitMethod(Opcodes.ACC_PUBLIC, "m", "()V",
+				null, null);
+		mv.visitCode();
+		addCauseOfResizeInstructions(mv);
+		addCauseOfGetCommonSuperClass(mv);
+		mv.visitMaxs(1, 1);
+		mv.visitEnd();
+		cw.visitEnd();
+		final byte[] original = cw.toByteArray();
+		assertEquals(true, computedCommonSuperClass);
```
marchof: I prefer `assertTrue(computedCommonSuperClass)` and I think we use it consistently in other tests.

Godin:
@marchof 

> Not sure how this test asserts that getCommonSuperClass() is not called during instrumentation? java/lang/Integer should load without any problems.

Default implementation of `String getCommonSuperClass(final String type1, final String type2)` that is used during instrumentation loads both `type1` and `type2`, so while loading of `java/lang/Integer` is ok, loading of created class `Example` will fail with `ClassNotFoundException` if this method is called.  You can observe this in a build where test was added or by downgrading ASM to 0.5.1.

> I think we should now add an assertion in Instrumenter that forbids execution of getCommonSuperClass() at all.

Indeed. I'll add this.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
Godin: You right. Fixed.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
Godin: Done.

marchof:
@Godin Somehow "Add your review" and then checking "Approve" does not seem to work currently. Anything I overlooked here?

Godin:
@marchof don't know why you wasn't able to approve. I can bypass this check...

PR #418, opened 2016-05-29
Test whether there are child execution data files to report upon

timp:
Ensure that the aggregate report is only created if there are data files present in dependent children.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
@@ -86,7 +86,7 @@ public class ReportAggregateMojo extends AbstractReportMojo {
 
 	@Override
 	boolean canGenerateReportRegardingDataFiles() {
-		return true;
+		return executionDataFilesCount() > 0;
 	}
 
 	@Override
@@ -107,6 +107,22 @@ public class ReportAggregateMojo extends AbstractReportMojo {
 		for (final MavenProject dependency : findDependencies(
 				Artifact.SCOPE_COMPILE, Artifact.SCOPE_TEST)) {
 			loadExecutionData(support, filter, dependency.getBasedir());
+		}
+	}
+
+	private int executionDataFilesCount() {
+		final FileFilter filter = new FileFilter(dataFileIncludes,
+				dataFileExcludes);
+		try {
+			int count = 0;
+			for (final MavenProject dependency : findDependencies(
+					Artifact.SCOPE_COMPILE, Artifact.SCOPE_TEST)) {
+				count += filter.getFiles(dependency.getBasedir()).size();
+			}
+			return count;
+		} catch (IOException e) {
+			throw new RuntimeException(
+					"Unable to determine number of executing data files", e);
 		}
 	}
 
--- a/jacoco-maven-plugin.test/it/it-no-aggregate-report/invoker.properties
+++ b/jacoco-maven-plugin.test/it/it-no-aggregate-report/invoker.properties
@@ -1,0 +1,4 @@
+invoker.goals = clean install site
+
+# maven-site-plugin 3.0 requires at least Maven 2.2.0
+invoker.maven.version = 2.2.0+
--- a/jacoco-maven-plugin.test/it/it-no-aggregate-report/pom.xml
+++ b/jacoco-maven-plugin.test/it/it-no-aggregate-report/pom.xml
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+   All rights reserved. This program and the accompanying materials
+   are made available under the terms of the Eclipse Public License v1.0
+   which accompanies this distribution, and is available at
+   http://www.eclipse.org/legal/epl-v10.html
+
+   Contributors:
+      Marc R. Hoffmann, Jan Wloka - initial API and implementation
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>jacoco</groupId>
+    <artifactId>setup-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>no-aggregate</artifactId>
+  <packaging>jar</packaging>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>@project.groupId@</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>prepare-agent</id>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>report</id>
+            <goals>
+              <goal>report</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+  <reporting>
+    <plugins>
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </reporting>
+</project>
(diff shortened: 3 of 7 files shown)


marchof:
What is the use case for this?

As discussed earlier, this case is probably wrong usage of the report-aggregate goal. This goal should only be configured for a module which is responsible for generating the aggregate report. If the goal is e.g. inherited "by chance" from a parent POM I would consider this as wrong configuration.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -110,6 +110,22 @@ void loadExecutionData(final ReportSupport support) throws IOException {
 		}
 	}
 
+	private int executionDataFilesCount() {
+		final FileFilter filter = new FileFilter(dataFileIncludes,
+				dataFileExcludes);
+		try {
+			int count = 0;
+			for (final MavenProject dependency : findDependencies(
```
marchof: Exec file in report project itself is missung here, see #414.

timp:
As pointed out earlier by @Godin  the use case is that maven site goal introduces an execution of the report-aggregate goal, so producing an empty aggregate report in normal projects.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
timp: Yes,  understood. We discussed and the report-aggregate goal is not currently intended to include the reported upon project. #414 appears to reverse this.  Should #414 be  merged this will need to change.

marchof: #414 **is** merged.

marchof:
@Godin @timp But only if it is configured right? Which is invalid setup from my point of view for non-reporting modules.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
Godin: @timp #413 = exec file + project sources , #414 = only exec file , #414 was merged

timp:
@marchof so if there is no mention of the jacoco plugin within the reporting section then this behaviour will not occur? 
I have 
```
    <reporting>
      <plugins>
        <plugin>
          <groupId>org.jacoco</groupId>
          <version>0.7.7-SNAPSHOT</version>
        </plugin>
      </plugin
    </reporting
```
This PR may help reduce confusion.

Godin:
@marchof as far as I understood inclusion happens implicitly as soon as jacoco-maven-plugin declared as a reporting plugin for site.

@timp it would be clearer, if this PR will have integration test for the case it tries to cover

Godin:
@marchof could be noted that `ReportAggregateMojo` participates in site generation, because inherits from `AbstractReportMojo`, which inherits from `AbstractMavenReport`, which implies
> The basis for a Maven report which can be generated both as part of a site generation or as a direct standalone goal invocation.

This is desired behaviour for `ReportMojo ` as per 8b73c3c455431e8f5400595dcb193b1427cbce48.

If  `ReportAggregateMojo` should not participate in site generation, then we can simply break this inheritance chain.

For the time being I hadn't time to study this PR, so it is unclear to me what are the consequences of this PR on various cases.

timp:
I have pulled latest master and passed all current tests. I am looking at adding an integration test, but may not have time.

timp:
@godin A test added.

Godin:
@marchof @timp guys, I had some time to do a wider look on interoperability with maven-site-plugin:
changing our reference example as
```diff
diff --git c/jacoco-maven-plugin.test/it/it-report-aggregate/child2/pom.xml i/jacoco-maven-plugin.test/it/it-report-aggregate/child2/pom.xml
index e340ca2..b7f3e98 100644
--- c/jacoco-maven-plugin.test/it/it-report-aggregate/child2/pom.xml
+++ i/jacoco-maven-plugin.test/it/it-report-aggregate/child2/pom.xml
@@ -21,5 +21,12 @@
 
   <artifactId>child2</artifactId>
   <packaging>jar</packaging>
-  
+
+  <dependencies>
+    <dependency>
+      <groupId>jacoco</groupId>
+      <artifactId>child1</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
 </project>
diff --git c/jacoco-maven-plugin.test/it/it-report-aggregate/invoker.properties i/jacoco-maven-plugin.test/it/it-report-aggregate/invoker.properties
new file mode 100644
index 0000000..df5d7d2
--- /dev/null
+++ i/jacoco-maven-plugin.test/it/it-report-aggregate/invoker.properties
@@ -0,0 +1 @@
+invoker.goals = clean install site
diff --git c/jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml i/jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml
index 71b7a19..074b697 100644
--- c/jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml
+++ i/jacoco-maven-plugin.test/it/it-report-aggregate/pom.xml
@@ -45,4 +45,14 @@
       </plugin>
     </plugins>
   </build>
+
+  <reporting>
+    <plugins>
+      <!-- https://maven.apache.org/plugins/maven-site-plugin/examples/configuring-reports.html#Selecting_Reports_from_a_Plugin:_Configuring_Report_Sets -->
+      <plugin>
+        <groupId>org.jacoco</groupId>
+        <artifactId>jacoco-maven-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </reporting>
 </project>
```
Note that this adds dependency from `child2` to `child1`.
And currently would cause
 1. creation of usual report and link in generated site only in modules `child1` and `child2`
 2. creation of empty aggregate report and link in generated site in root and module `child1`
 3. creation of aggregate report and link in generated site in modules `child1-test` (coverage of `child1` by tests in `child1` and `child1-test`), `child2` (coverage of `child1` by tests in `child1` and `child2`) and `report` (coverage of `child1` and `child2` by tests in `child1`, `child1-test`, `child2` and `report`)

After merge of this PR the only change - no more empty aggregate report and link in root and module `child1`. And actually mismatch between `executionDataFilesCount` and `loadExecutionData` in regards to current project is correct one, otherwise `child1` would have empty aggregate report.

However what is bothering - is a report generated in `child2`, because it doesn't contain sources of `child2` and this concern was described in #413.

Should be noted that implicit execution of all reporting mojos - is expected behavior as per https://maven.apache.org/plugins/maven-site-plugin/examples/configuring-reports.html#Selecting_Reports_from_a_Plugin:_Configuring_Report_Sets :
> By default, when you add a plugin in reporting section, every reporting goal available in the plugin is rendered once.

and it also states 

> Notice: don't forget to define reportSets for plugins which provide aggregator reports, like maven-javadoc-plugin, maven-jxr-plugin or maven-checkstyle-plugin or you'll have these aggregate reports run by default in addition to non-aggregate reports.

Hence we can:
 1. leave everything as is and refer to a statement above about `reportSets`
 2. completely forbid execution of `ReportAggregate` as reporting mojo by breaking inheritance from `AbstractMavenReport` or by unconditionally returning `false` from `canGenerateReport`
 3. merge this PR after addition of comment about mismatch, and probably merge of integration tests to simplify them
 4. merge this PR and polished #413
 5. 4th option actually almost completely removes the differences between `ReportMojo` and `ReportAggregateMojo`, so that maybe last one can be integrated into first one with addition of parameter `includeDependencies` or something like that. This would also imply that aggregation during site generation would always be requiring configuration.

4th option looks good to me and doesn't prevent to do 5th in future.

@marchof WDYT?

marchof:
@Godin @timp Thanks for the example and the analysis!

I wasn't aware of this Maven "feature" and we probably should do something about this to prevent unwanted aggregate reports. In a multi module project where every module has its own tests and most modules have dependencies to each to others your would get (redundant) aggregate reports all over the place. Even with this PR as exec files are present.

So the questions is not about **empty** aggregates report but about whether it makes sense to create  separate aggregate reports at all for every module.

Godin:
@marchof you right - redundant aggregate reports in a multi-module project is not nice. What means that we should go ahead with option 1 or 2 ? Option 1 looks good to me since it is simplest solution and anyway `reportSets` **must be defined explicitly** for standard Maven plugins like `maven-javadoc-plugin`.

marchof:
@Godin If we go for option 1 and add documentation? Or will this cause trouble for existing users?

Godin:
@marchof what do you mean by "trouble"? if existing users don't have explicit `reportSets`, i.e. they have declaration as
```
  <reporting>
    <plugins>
      <plugin>
        <groupId>org.jacoco</groupId>
      </plugin>
    </plugins>
  </reporting>
```
then during upgrade they will need to change it on
```
  <reporting>
    <plugins>
      <plugin>
        <groupId>org.jacoco</groupId>
        <reportSets>
          <reportSet>
            <reports>
              <report>report</report>
            </reports>
          <reportSet>
        </reportSets>
      </plugin>
    </plugins>
  </reporting>
```
otherwise they will be getting aggregate reports everywhere.

marchof:
@Godin I'm o.k. with option 1 for now. Do you add a comment to the goal documentation?

PR #420, opened 2016-06-03
Add note about explicit selection of non-aggregate reports

Godin:
As discussed in #418

### DIFF
--- a/org.jacoco.doc/docroot/doc/maven.html
+++ b/org.jacoco.doc/docroot/doc/maven.html
@@ -103,6 +103,33 @@
 mvn help:describe -Dplugin=org.jacoco:jacoco-maven-plugin -Ddetail
 </pre>
 
+<p>
+  Usage of plugin together with <tt>maven-site-plugin</tt> without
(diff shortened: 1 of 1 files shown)


Godin:
@marchof could you please verify? Note that I decided to update "usage" section rather than specific "report-aggregate" goal, think this is better.

org.jacoco.doc/docroot/doc/maven.html
```diff
@@ -103,6 +103,33 @@
 mvn help:describe -Dplugin=org.jacoco:jacoco-maven-plugin -Ddetail
 </pre>
 
+<p>
+  Usage of plugin in reporting section without
+  <a href="https://maven.apache.org/plugins/maven-site-plugin/examples/configuring-reports.html#Selecting_Reports_from_a_Plugin:_Configuring_Report_Sets">explicit selection of reports</a>
+  might lead to generation of redundant aggregate reports.
```
marchof: Only when using the site goal, right?

marchof:
@Godin Beside mention of site looks good to me!

org.jacoco.doc/docroot/doc/maven.html
Godin: @marchof right, fixed

PR #322, opened 2016-06-10
Add user property for parameter "dataFile" of "report" mojo

Godin:
See https://groups.google.com/d/msg/jacoco/XtIKmcuqbo8/LUJIIPLjvLYJ


Godin:
@widgetpl this is not that easy in case of properties whose values are not primitive, e.g. `rules` of `check` mojo, or even lists such as `includes` and `excludes` of `report` mojo, while it is easy in case of `dataFile` of `report` mojo. So instead of trying to climb a mountain using too general statement "all of them", let's do a baby steps and solve various cases separately, and especially that as of today there is not that much parameters, which cannot be specified from command-line.

widgetpl:
@Godin yes You are right, I am requiring too much and for me dataFile is enough for now as i had problem just with this parameter and when i wrote about rest of them i thought that it might be useful in future for someone ;)

Godin:
@widgetpl thank you for your understanding!

Godin:

Funny that fix is one line, while changes for integration test much bigger 😄

PR #423, opened 2016-06-10
Fix 0.7.7 regression: broken links to stylesheet and javascript in documentation

Godin:
In #401 we forgot to adapt some files in documentation to new location of resources:

image


Godin:

marchof:
@Godin Thanks for spotting this! I wonder whether we can add a link checking tool to our build to avoid such regressions.

Godin:
@marchof do you have an idea of such tool?

marchof:
@Godin Our own www.eclemma.org site generator does this. Not sure whether also a maven plug-in can help here, e.g.: https://maven.apache.org/plugins/maven-linkcheck-plugin/

PR #421, opened 2016-06-11
Rename dot folder in documentation 

Godin:
With #401 we renamed the `.resources` folder from JaCoCo reports to be more hosting friendly. We should do the same for our own documentation. Currently we have:

    /org.jacoco.doc/docroot/doc/.resources/


Godin:
I think that in contrast with #401 we can safely simply omit dot here.

marchof:
@Godin LGTM, thanks!

PR #426, opened 2016-06-17
Instrumentation with creation of "companion" classes instead of addition of members

Godin:


org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,197 @@
...
+				}
+				companions.put(classLoader, companion);
+			} else {
+				cls = companion.cls.get();
+				// if class loader is alive, then class is also alive
+				assert cls != null;
+			}
+
+			if (companion.usedFields == FIELDS_PER_CLASS) {
+				throw new UnsupportedOperationException();
```
bjkail: This seems like an unfortunate limitation.  Can additional companion classes be generated as needed when the limit is reached?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,197 @@
...
+			}
+			field = "p" + companion.usedFields;
```
bjkail: A getFieldName method shared with createCompanionClass would be nice.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,197 @@
...
+			companion.usedFields++;
+		}
+
+		final boolean[] probes = runtimeData
+				.getExecutionData(classId, className, probeCount).getProbes();
+		try {
+			cls.getField(field).set(null, probes);
```
bjkail: I believe Class.getField is O(n), which is unfortunate with large FIELDS_PER_CLASS.  Perhaps smaller FIELDS_PER_CLASS would help, especially if multiple companion classes can be generated, or perhaps an initialize(int fieldId, boolean[]) method with a tableswitch, which would be somewhat wasteful one-use bytecode.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+	public int generateDataAccessor(final long classid, final String classname,
+			final int probecount, final MethodVisitor mv) {
+		if (instrumented.size() == Companions.FIELDS_PER_CLASS) {
+			throw new UnsupportedOperationException();
+		}
+
+		final String fieldName = "p" + Long.toString(classid);
+
+		if (instrumented.add(classname)) {
+			cw.visitField(
```
bjkail: The ASM javadoc is specific about the order of method calls, and I always assumed MethodVisitor.visitEnd needed to be called before subsequent visitField/visitMethod, so this seems somewhat fragile.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+					Opcodes.ACC_PUBLIC | Opcodes.ACC_STATIC
+							| Opcodes.ACC_TRANSIENT | Opcodes.ACC_FINAL,
+					fieldName, InstrSupport.DATAFIELD_DESC, null, null);
+
+			generator.generateDataAccessor(classid, classname, probecount,
+					this.mv);
+
+			this.mv.visitFieldInsn(Opcodes.PUTSTATIC, companionName, fieldName,
+					InstrSupport.DATAFIELD_DESC);
```
bjkail: I guess this will overflow maximum instructions per method before FIELDS_PER_CLASS is reached?  Again, I wonder if it would be good to generate multiple companion classes.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	/**
+	 * Returns the bytecode of generated "companion" class.
+	 *
+	 * @return bytecode of generated "companion" class
+	 */
+	public byte[] getClassDefinition() {
+		mv.visitInsn(Opcodes.RETURN);
+		mv.visitMaxs(4, 0);
```
bjkail: Doesn't maxStack vary depending on generator.generateDataAccessor?  Should this use ClassWriter.COMPUTE_MAXS?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail you right that `Class.getField` is O(n) , what also explains degradation of performance, which was observed when `usedFields` was stored inside of "companion class". Implemented both `tableswitch` and generation of multiple classes.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail done

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail done, number of fields per class was lowered from 65000 to 8000 due to size of "init" method, however IMO initial value  was hardly reachable on practice

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
Godin: @bjkail in this case `generator` is always `OfflineInstrumentationAccessGenerator` and it requires stack size of `4`

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
Godin: @bjkail indeed, so lowered limit to 5000 and implemented generation of multiple classes

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
Godin: @bjkail this can be fixed, however: ASM Javadoc defines order for methods of `ClassVisitor` and `MethodVisitor`, but not between them - quoting http://download.forge.objectweb.org/asm/asm4-guide.pdf :
> Note that it is not necessary to finish one method in order to start visiting another one. In fact MethodVisitor instances are completely independent and can be used in any order (as long as cv.visitEnd() has not been called)

bjkail: That's useful to know, thanks.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
bjkail: I see.  You might as well use the return value of generator.generateDataAccessor in case it changes, though?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,253 @@
...
+			companionName = companion.getName();
+			fieldId = companion.usedFields;
+			companion.usedFields++;
+		}
+
+		final boolean[] probes = runtimeData
+				.getExecutionData(classId, className, probeCount).getProbes();
+		try {
+			// TODO(Godin): can cache "getMethod", if we'll use same strategy as
+			// for access of runtime data - via "equals" method
```
bjkail: Yes, that seems good.  To avoid GC problems, the generated class can use clinit to create/store an instance of itself in a static final, and then Companion can hold a WeakReference to that instance.  Not sure if it's worth avoiding even a single getField, but if so, you could add a method to return the instance.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,253 @@
...
+			throw new RuntimeException(e);
+		}
+	}
+
+	public static final String COMPANION_NAME = InstrSupport.DATAFIELD_NAME;
+
+	private static final String INIT_METHOD_NAME = InstrSupport.INITMETHOD_NAME;
+
+	/**
+	 * https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.11
```
bjkail: This JVMS reference is interesting, but it no longer justifies the current value, so a comment on how this value was chosen would be useful, even if arbitrary or to reduce the class memory size.  Or is it entirely chosen by the maximum method size needed by OfflineInstrumentationCompanionAccessGenerator, or is it arbitrarily lower?  If you're trying to maximize this value, I guess the dynamically generated classes could be larger.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
Godin: @bjkail I've copied code of `OfflineInstrumentationAccessGenerator` into `OfflineInstrumentationCompanionAccessGenerator`. This seems simpler.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail you right and this is a good question - which value we should choose and based on what?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail you perfectly described an idea of this comment. However for now I don't see evidences that optimization is required here, so probably we can defer this idea of optimization.

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
bjkail: Duplicating the logic seems worse to me.  If you don't want the complexity of using the return value from generateDataAccessor, then perhaps add a MAX_STACK constant to OfflineInstrumentationAccessGenerator and use it in both places?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
bjkail: Then I'd suggest removing the TODO.  If you want to keep a comment, then I would suggest mentioning there is no evidence yet that it's worthwhile and perhaps include details on how it could be implemented.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
```diff
@@ -200,6 +203,12 @@ public int instrumentAll(final InputStream input,
 
 	private int instrumentZip(final InputStream input,
 			final OutputStream output, final String name) throws IOException {
+		if (generators.getLast() instanceof
+				OfflineInstrumentationCompanionAccessGenerator) {
+			generators
+					.addLast(new
+							OfflineInstrumentationCompanionAccessGenerator());
```
bjkail: instrumentAll was previously thread-safe, but it no longer is due to this addLast when used with OfflineInstrumentationCompanionAccessGenerator, which is also not thread-safe.  This should probably be documented somehow for both classes.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
```diff
@@ -216,6 +225,20 @@ private int instrumentZip(final InputStream input,
 			}
 			zipout.closeEntry();
```
bjkail: I guess this while loop also needs to check Companions.FIELDS_PER_CLASS?

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
bjkail: The actual value seems less important than explaining how the current value was chosen.    Perhaps something like this?

> This JVM specification requires this value to be less than 65535.  Additionally, the value must be low enough that we do not exceed the maximum method size for the initialize method in dynamically generated companion classes or the clinit method in offline generated companion classes.  This value can be arbitrarily small since multiple companion classes are generated as needed, so we choose a smaller value to avoid dynamically generating very large companion classes in case few classes are actually instrumented in a particular class loader.

Or perhaps put different constants with different comments in Companions vs OfflineInstrumentationCompanionAccessGenerator so that the comments can be a bit more specific/clearer.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
bjkail: Perhaps OfflineInstrumentationCompanionAccessGenerator should have an end/close method and a constructor that accepts an IClassFileWriter with `writeClassFile(String name, byte[] bytes)`.  That way, the logic for when to write the companion classes could be encapsulated in OfflineInstrumentationCompanionAccessGenerator.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
Godin: @bjkail done

org.jacoco.core/src/org/jacoco/core/runtime/OfflineInstrumentationCompanionAccessGenerator.java
Godin: @bjkail done

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
Godin: @bjkail done

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
Godin: @bjkail I'm not sure that we were giving guarantee that `instrumentAll` is thread safe, or that we demanded implementation of `IExecutionDataAccessorGenerator` to be thread safe. I personally tend to consider everything as non thread safe unless stated otherwise.

One alternative to preserve thread safety here - is to create a new instance of `Instrumenter` inside instead of usage of `LinkedList`. But this will mean that `Instrumenter` should be final, thus again will be possible to say that we are breaking something. However again not sure that we were giving guarantee that clients can extend `Instrumenter`.

Another alternative is to introduce private methods `instrument*` with generator as parameter. But not sure if it's worth it, especially that `OfflineInstrumentationCompanionAccessGenerator` is not thread safe and requires call at the end of instrumentation.

In any case I want to believe that we have an ability to change API contracts when we have no other choice or when complexity of preservation is high, at least by doing increment of version up to `0.8.0`.

bjkail: If the general policy is objects are thread-unsafe unless otherwise specified, then doing nothing is fine with me.  There's no API "break" since client code will need to opt into using OfflineInstrumentationCompanionAccessGenerator anyway.

Creating a new Instrumenter per nested archive seems cleaner than using a LinkedList.  If you want to allow clients to extend Instrumenter, you could add an Instrumenter(Instrumenter) copy constructor and a newInstrumenter(Instrumenter) method that clients could override.  Again, nothing would change for current clients since that method would be unused unless the client opted in to using OfflineInstrumentationCompanionAccessGenerator.  If you'd rather keep the LinkedList mechanism, then perhaps move the companionGenerator local variable to the outer block rather than doing `generators.getLast() instanceof ...` twice?

bjkail:
Overall, it seems like the branch is coming along nicely.  Have you considered renaming/exposing the Companions class for clients that implement their own class loader hook rather than using the agent?  That seems useful.  Perhaps a RuntimeInstrumenter with an instrument(ClassLoader, String, byte[])?

Godin:
@bjkail well, I'm not worried that we'll be able to polish APIs. My main worry as before - is a functional aspect. So did some testing of real-life projects and realized that current implementation for "offline" mode in Maven doesn't play nicely with incremental-compilation. More specifically: first execution of `mvn test` works, but second one fails, because classes already instrumented. In fact this issue exist even without "companion" classes, but IMO idea of "companion" classes raises additional questions of how this can be handled. Also wondering if this problem arises in Ant, Gradle, etc.

Godin:
Thinking more about such use-case:

Seems easy to fix without "companion" classes - "offline" mode can just skip already instrumented classes.

A kind of similar approach for case of one "companion" class per instrumented class. What returns us back to a question - is it a really bad strategy? Given that rough estimation of memory consumption (creation of "companion classes" till exhaustion of 128M PermGen/Metaspace):
```
JDK 1.6 - 139803 classes, so approximately 960 B per class
JDK 1.7 - 175994 classes, 762 B per class
JDK 1.8 - 185591 classes, 723 B per class
```
Would be nice to have some measurements/feedback on this from a real big projects, or create synthetic but realistic benchmarks.

Similar approach for case of many "companion" classes in a worst-case degrades to the case of one per class, but pollutes memory at runtime, so probably we'll need to "reconstruct" "companion" classes.

Godin:
@marchof will be great to hear comments about this PR from you, of course only if you have enough time.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
Godin: @bjkail btw, I think that `Instrumenter` is non thread safe because of `setRemoveSignatures`

bjkail: Well, yes, but I think it's obvious that setRemoveSignatures mutates the instance and should be used to initialize the instance, but I think it's less obvious that instrumentAll is thread-unsafe.  Anyway, since the generator is already thread-unsafe, I think it's fine if Instrumenter becomes thread-unsafe when using that generator.

bjkail:
It seems to me that classes will always have sufficiently high JVM overhead that we want to minimize them, particularly due to PermGen prior to Java 8.  Well, if you think bulk companion classes are causing too many more problems that cannot be corrected, then I guess you can revisit per-class companion classes.  Perhaps there should be an option in any case to disable extra classes of any kind in case they cause problems (either extra memory or unexpected classes causing issues for other ClassFileTransformer or Instrumentation.getAll/InitiatedClasses).

I'm still interested in exploring invokedynamic for Java 7 classes once this basic support for companion classes is complete (in whatever form) since that approach would only require a single, small, fixed-size bootstrap companion class.

Godin:
@bjkail I added strategy, which utilizes `invokedynamic` - see `InvokeDynamicProbeArrayStrategy`. And here are some notes: bootstrap method performs request of probe array and since bootstrap method invoked for **each** `invokedynamic` instruction to avoid multiple requests of probe array for the same class all its methods should use one and same instruction, what implies addition of initialization method with this instruction, and thus bootstrap method also can be added into the class. Other ways to utilize `invokedynamic` look suboptimal to me for now.

Godin:
@bjkail regarding your statement
> It seems to me that classes will always have sufficiently high JVM overhead that we want to minimize them, particularly due to PermGen prior to Java 8.

I would prefer to have more formal measurements than feelings. Any ideas of how to obtain them? Also for example JVM based languages such as JRuby, Groovy, Scala as far as I know were generating quite a good amount of classes and yet existed prior to appearance of MetaSpace in Java 8. And I suppose that our classes are much smaller.

> I guess you can revisit per-class companion classes. 

Problem is that instrumented classes in case of "offline" mode are statically linked to "companion" classes, so that any re-building of a set of "companion" classes will also require re-instrumentation of all classes.

bjkail:
> I would prefer to have more formal measurements than feelings. Any ideas of how to obtain them? Also for example JVM based languages such as JRuby, Groovy, Scala as far as I know were generating quite a good amount of classes and yet existed prior to appearance of MetaSpace and other optimizations in Java 8. And I suppose that our classes are much smaller.

The only issue for me is backwards compatibility.  People should expect running with JaCoCo will increase memory consumption on a per-class basis, but the prototype adds additional per-class overhead of ~700 bytes for metaspace + Companion + WeakReference + HashMap$Entry.  Even if we accurately measure that overhead, we can't know if it will be acceptable.  We can document what we find, and we can give users an option to disable if it's unacceptable (either because it's too much overhead or because extra classes cause issues for their environment just like unexpected members can).

> Problem is that instrumented classes in case of "offline" mode are statically linked to "companion" classes, so that any re-building of a set of "companion" classes will also require re-instrumentation of all classes.

Sorry, I'm not familiar enough with the Ant/Maven code, so I'm missing the concern and can't offer suggestions.  Does the Ant/Maven code somehow avoid reprocessing all classes today?  In what case does the companion class need to be rebuilt when we're not just reinstrumenting all classes anyway?

> I added strategy, which utilizes invokedynamic - see InvokeDynamicProbeArrayStrategy. And here are some notes: bootstrap method performs request of probe array and since bootstrap method invoked for each invokedynamic instruction to avoid multiple requests of probe array for the same class all its methods should use one and same instruction, what implies addition of initialization method with this instruction, and thus bootstrap method also can be added into the class. Other ways to utilize invokedynamic look suboptimal to me for now.

We can do better if storeInstance calls visitInvokeDynamic directly (ASM will merge all calls to a single CONSTANT_InvokeDynamic, so there is no benefit to $jacocoInit) and passes the classid, classname, and number of probes as arguments to the bootstrap method, which can forward those arguments to the access generator.  That will require refactoring since the current code expects those to be generation-time constants (e.g., add IExecutionDataAccessorGenerator2 with generateDataAccessor(MethodWriter) that RuntimeData.generateArgumentArray is already on the stack).  If we do it that way, then bootstrap method itself no longer has hard-coded values (they are encoded in the CONSTANT_InvokeDynamic), so the bootstrap method can be shared by all classes, so you can generate a single companion class per ClassLoader that contains this bootstrap method, which means classes no longer need to have any additional members.  This reduces the overhead to a CallSite per class and one class+method per ClassLoader, but the improvement will be limited to Java 7+ class files.  That might be too restrictive, which is why this might only be supplementary to the current prototype.

Godin:
> We can do better if storeInstance calls visitInvokeDynamic directly (ASM will merge all calls to a single CONSTANT_InvokeDynamic, so there is no benefit to $jacocoInit)

Sorry, but this sounds unbelievable - you can try:
```diff
diff --git c/org.jacoco.core/src/org/jacoco/core/internal/instr/InvokeDynamicProbeArrayStrategy.java i/org.jacoco.core/src/org/jacoco/core/internal/instr/InvokeDynamicProbeArrayStrategy.java
index 142487c..28f1943 100644
--- c/org.jacoco.core/src/org/jacoco/core/internal/instr/InvokeDynamicProbeArrayStrategy.java
+++ i/org.jacoco.core/src/org/jacoco/core/internal/instr/InvokeDynamicProbeArrayStrategy.java
@@ -44,9 +44,9 @@ class InvokeDynamicProbeArrayStrategy implements IProbeArrayStrategy {
        }
 
        public int storeInstance(final MethodVisitor mv, final int variable) {
-           mv.visitMethodInsn(Opcodes.INVOKESTATIC, className,
-                           InstrSupport.INITMETHOD_NAME, InstrSupport.INITMETHOD_DESC,
-                           false);
+         mv.visitInvokeDynamicInsn("$jacocoGetProbeArray", "()[Z",
+                         new Handle(Opcodes.H_INVOKESTATIC, className,
+                                      BOOTSTRAP_METHOD_NAME, BOOTSTRAP_METHOD_DESC, false));
                return 1;
        }
```
and test will fail, because there will be two `invokedynamic` instructions in two methods and bootstrap method invoked for each.

Godin:
Point of having one instruction per class - is to avoid multiple round trips to the `RuntimeData` as it is now. I can craft a benchmark, but pretty sure that multiple instructions would have bigger cost than single and than current null check for static field.

bjkail:
Sorry, you're right: the JVM resolves the CallSite per invokedynamic instruction, not per CONSTANT_InvokeDynamic, so the RuntimeData round trip would be required the first time each method is invoked.  The only real reason to use invokedynamic is to avoid adding members, so if we can't avoid that, then there doesn't seem to be a good reason to use it.

Re: overhead, it's the difference between $jacocoInit's RuntimeData round trip per class vs the bootstrap method and RuntimeData round trip per method.  If you test a class with one method, it's likely comparable (just bootstrap method overhead).  If you test a class with multiple methods and invoke them all once, then invokedynamic will likely lose.  If you include the fixed per class and per method overhead of Instrumenter and invoke the methods multiple times, then the overhead might be negligible.

Well, you're the one doing all the work, and I'm just offering suggestions, so feel free to ignore this tangent and go back to the companion classes work.  As mentioned, there probably needs to be a basic approach in place that can handle all classes not just Java 7+ before considering invokedynamic.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
```diff
@@ -0,0 +1,267 @@
...
+	/**
+	 * {@link ClassLoader#defineClass(String, byte[], int, int)}
+	 */
+	private static final Method DEFINE_CLASS;
+
+	static {
+		try {
+			DEFINE_CLASS = ClassLoader.class.getDeclaredMethod("defineClass",
+					String.class, byte[].class, Integer.TYPE, Integer.TYPE);
+			DEFINE_CLASS.setAccessible(true);
```
Godin: Most likely `setAccessible` won't work in Java 9.

bjkail: Why is that?  My understanding is that `setAccessible` will restrict access to unexported packages, but `java.lang` will necessarily be exported.

Godin: @bjkail with JDK 9 EA b148
```
import java.lang.reflect.Method;

public class Example {
    Method DEFINE_CLASS;
    try {
      DEFINE_CLASS = ClassLoader.class.getDeclaredMethod("defineClass", String.class, byte[].class, Integer.TYPE, Integer.TYPE);
      DEFINE_CLASS.setAccessible(true);
    } catch (NoSuchMethodException e) {
      throw new RuntimeException(e);
    }
  }
}
```
gives me
```
Exception in thread "main" java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @6bdf28bb
        at java.base/jdk.internal.reflect.Reflection.throwInaccessibleObjectException(Reflection.java:427)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:201)
        at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:192)
        at java.base/java.lang.reflect.Method.setAccessible(Method.java:186)
        at Example.main(Example.java:8)
```
quoting https://groups.google.com/d/msg/jacoco-dev/g9P4xeXDx8A/9yClieKwCQAJ 
> With the updated proposal in the JSR, this refresh changes setAccessible further so that it cannot be used to break into non-public types, or non-public elements of public types, in exported packages.

bjkail: Interesting.  I'm behind on tracking the module work.  I wonder if the javaagent can be used to get around the restriction?  I recall a thread on one of the mailing lists, and [#ReflectiveAccessByInstrumentationAgents](http://openjdk.java.net/projects/jigsaw/spec/issues/#ReflectiveAccessByInstrumentationAgents) and [Instrumentation.redefineModule](http://download.java.net/java/jdk9/docs/api/java/lang/instrument/Instrumentation.html#redefineModule-java.lang.reflect.Module-java.util.Set-java.util.Map-java.util.Map-java.util.Set-java.util.Map-) seem promising, even if I don't fully understand how they would work.  I seem to recall javaagent do not have their own Module, so perhaps one would have to be dynamically created that contains code to call setAccessible, and then the Module would be specified in extraOpens for "java.lang".  Or perhaps it's worth asking Rory or the mailing lists for advice?

Godin: @bjkail Well most likely you right and [#ReflectiveAccessByInstrumentationAgents](http://openjdk.java.net/projects/jigsaw/spec/issues/#ReflectiveAccessByInstrumentationAgents) ( [`Instrumentation.redefineModule`](http://download.java.net/java/jdk9/docs/api/java/lang/instrument/Instrumentation.html#redefineModule-java.lang.reflect.Module-java.util.Set-java.util.Map-java.util.Map-java.util.Set-java.util.Map-) ) is a way to go. To be honest - I also didn't followed story behind this closely enough to give more details, because we don't use reflection in master. In any case fact that things change on this subject in JDK 9, increases risks for incorporation of such use of reflection.

> I seem to recall javaagent do not have their own Module, so perhaps one would have to be dynamically created that contains code to call setAccessible, and then the Module would be specified in extraOpens for "java.lang".

I assume that simply [`Class.getModule`](http://download.java.net/java/jdk9/docs/api/java/lang/Class.html#getModule--) will do the job.

I don't know what to ask from Rory - I would prefer to dig deeper into a subject and/or try to use to get some concrete questions, but have doubts about my free time in short-term. Also suspect that as an answer you'll be redirected to one of OpenJDK mailing lists :wink: However don't hesitate to ask.

Godin:
For the record on this subject: usage of `assert`ions leads to generation of synthetic field `$assertionsDisabled`.

bjkail:
@Godin You're right, but if a user adds an `assert`, the field will always be present in the class file, so if they have other (fragile) code that does not skip synthetic members, it will fail always while the `assert` is present rather than only failing when testing with JaCoCo.

org.jacoco.core/src/org/jacoco/core/internal/instr/Companions.java
bjkail: There has been recent [discussion](http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-March/011557.html) around adding a [Lookup.defineClass](http://download.java.net/java/jigsaw/docs/api/java/lang/invoke/MethodHandles.Lookup.html#defineClass-byte:A-) method.

Godin: @bjkail I'm aware of it 😉  but thanks for notification :thumbsup:

PR #434, opened 2016-07-26
Fix instrumentation to not violate Java Virtual Machine Specification regarding initialization of final fields

Godin:
Currently `FieldProbeArrayStrategy` creates `static final` field `$jacocoData`, value for which is set by bytecode instruction `putstatic` in method `$jacocoInit`, but according to [JVMS](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-6.html#jvms-6.5.putstatic):

>  if the field is final the instruction must occur in the \<clinit\> method of the current class. Otherwise, an IllegalAccessError is thrown.

Starting with OpenJDK 9 EA b127 this condition is checked for class files with version `>=53` - http://hg.openjdk.java.net/jdk9/jdk9/hotspot/rev/c558d46c1af2#l11.76 Unfortunately we compile classes into 52 even with JDK 9, because maven-shade-plugin can't handle 53, and that's why we don't get failure in core validation tests, but later in integration tests for Ant on one of JDK classes:
```
Exception in thread "main" java.lang.IllegalAccessError: Update to static final field java.sql.Timestamp.$jacocoData attempted from a different method (java.sql.Timestamp) than the initializer method <clinit> 
    at java.sql.Timestamp.$jacocoInit(java.sql@9-ea/Timestamp.java)
    at java.sql.Timestamp.<clinit>(java.sql@9-ea/Timestamp.java)
    at org.jacoco.ant.TestTarget.main(TestTarget.java:36)
```

There is also a report in our mailing list - https://groups.google.com/d/msg/jacoco/jHw39ZUjUNY/L-gHeMeBBQAJ

See https://bugs.openjdk.java.net/browse/JDK-8157181 which is associated with this check and which also states:
> Methods that do not satisfy condition violate the assumptions of the compilers. Compiling such methods results in different behavior of compiled and interpreted code.

and which discusses removal of dependency on class file version.

Also might be that this check will be backported - see linked tickets, e.g. https://bugs.openjdk.java.net/browse/JDK-8161989

So that we should try to find a way to compile classes (or at least test targets) into bytecode version 53 ( #411 ).

In any case - field must not be `final`. But unfortunately this will make `FieldProbeArrayStrategy` unusable for interfaces with methods, because [JVMS](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.5-200-A.3) states that

> Fields of interfaces must have their ACC_PUBLIC, ACC_STATIC, and ACC_FINAL flags set


bjkail:
For interfaces, can we do similar to FieldProbeArrayStrategy except generate clinit (ala your commit 8ebd1d1) but just don't PUTSTATIC in $jacocoInit?  i.e., if a static method is somehow called before the PUTSTATIC in clinit (somehow? in case something like [Marc's BadCycles](https://groups.google.com/d/msg/jacoco-dev/TsUt2DjZ8Hk/Uz3c5vdiFgAJ) is possible?), then we fetch using the slow runtime indirection.

Godin:
@bjkail that's exactly what I was prototyping here = clinit + slow path, and there is no more `$jacocoinit` for interfaces, unless I overlooked something (we definitely should compile test classes into bytecode version 53). btw, seems that I managed to create an example similar to BadCycles, but with interfaces.

Godin:
@bjkail I was wrong about the example.

Godin:
@marchof while this request is not yet ready for code review and merge, because requires some cleanup and polishing of code, could you please have a look on idea in general to be sure that we are on a good track here and something wasn't overlooked by me? For example - maybe you know an example of "bad cycles" with interfaces?

I personally don't see other alternatives, not counting idea of "companion classes", which is anyway not ready for prime time, while this one looks quite good and really close to finalization.

Godin:
@marchof IMO now this PR completely ready to be merged, so could you please review?

bjkail:
FWIW, the core changes look good to me.

Godin:
@bjkail if my understanding of "FWIW" is correct, then - be sure your opinion matters! :wink: @marchof is probably just a bit busy being at JCrete :laughing:

marchof:
Tried to create a cyclic initialization example for interfaces. Results in a JVM crash: https://gist.github.com/marchof/c0ac04a2abb4b163bf410e8018a72cb7

Will report to OpenJDK.

bjkail:
@Godin I assume that once OpenJDK fixes the bug that marchof found that the behavior for cyclic initialization of interfaces will have the same behavior as classes, so I guess it's best to add the slow path logic.

Godin:
@marchof you can also mention that this crashes on the latest JDK 9 EA b131.

@bjkail probably yes.

marchof:
@Godin Unfortunatelly OpenJDK-Bugs are "write once". Internal ID is JI-9042792.

Godin:
@marchof let me send an email ;)

Godin:
As per comments in https://bugs.openjdk.java.net/browse/JDK-8163969 - crash is a regression introduced in 8u40. So @marchof you can try such example on 8u33 for example, I quickly tried and it seems to work fine with strategy implemented here. So I'm still not convinced that we'll need slow path 😉

Godin:
Elaborating a bit more: when I thought about "bad cycles" previously (if my memory is correct), then "bad cycles" for classes are possible, because they can have constructor, but not possible for interfaces, because they can't.

bjkail:
@Godin The "bad cycle" happens when a method can be invoked before clinit.  If I add a clinit to JvmCrash.Target, it appears someMethod is invoked before the clinit runs, so it seems like the same issue would occur for the InterfaceFieldProbeArrayStrategy clinit?
```
/**
 * This snippet crashes with
 *   - Java(TM) SE Runtime Environment (8.0_101-b13) (build 1.8.0_101-b13)
 */
public class JvmCrash {

        interface Base {
                static final Object CONST = new Target(){}.someMethod();

                default void important() {
                        // Super interfaces with default methods get initialized (JLS 12.4.1)
                }
        }

        interface Target extends Base {
                static final Object CONST2 = new Object();
                default Object someMethod() {
                        new Throwable(String.valueOf(CONST2)).printStackTrace();
                        return null;
                }
        }

                new Target() {}.someMethod();
        }

}
```

Godin:
@bjkail Probably I'm too tired and a bit rushed. Even without addition of `CONST2` [bad-interface-cycle.zip](https://github.com/jacoco/jacoco/files/416084/bad-interface-cycle.zip)  runs fine without JaCoCo on JDK8u25 and fails with NPE with JaCoCo version from this PR. Most likely I made a mistake when tried to add it to core test cases.

Probability to face such case seems low since crash went thru multiple JDK updates. But as now we have test case, I'm fine to re-add slow path as there is no other options. However not sure how to add test case itself since it crashes.

bjkail:
@Godin I did think the ATHROW in JvmCrash should insert a probe, which should cause an NPE with this PR, but I didn't have an environment set up to prove it :-).

For white-box testing, perhaps create a specialized ClassVisitor/ClassWriter that returns a no-op MethodVisitor for visitMethod("<clinit>") so that the clinit is never actually inserted for the bad_cycles test?  For black-box testing, perhaps parse the JVM vendor/version and whitelist the known good versions (i.e., prior to b40 and update later to allow the versions with the fix), and you could fallback to the white-box approach otherwise.

The non-interface version of the bad_cycles test added in this PR is good to have, but I think it should be moved to another test class that uses FieldProbeArrayStrategy instead.

Godin:
@bjkail Even without `ATHROW` there should be a probe.

But there is something fishy, what makes me crazy right now, which maybe explains why at first I believed that do have an example of "bad cycles" with interfaces and then not - with current implementation on the same [bad-interface-cycle.zip](https://github.com/jacoco/jacoco/files/416084/bad-interface-cycle.zip) :
- 1.8.0_25-b17 - NPE
- 1.8.0_45-b14 - **no** NPE
- 1.8.0_102-b14 - **no** NPE

Godin:
Maybe that reason is the same as for crash... don't think I'm ready for bisection of JDK forest today  😪

Godin:
Both crash and disappearance of NPE are consequences of [JDK-8043275](https://bugs.openjdk.java.net/browse/JDK-8043275) - 
[bisect-log.txt](https://github.com/jacoco/jacoco/files/416708/bisect-log.txt)

Godin:
And here is an example of "bad cycle" with interfaces, which provokes NPE with current implementation of this PR on 1.8.0_25-b17 - 1.8.0_102-b14 :
```java
public class BadInterfaceCycle {

  interface Base {
    static final Object CONST = new Child() {
      {
        System.out.println("base clinit");
      }
    }.childDefaultMethod();

    default void baseDefaultMethod() {
    }
  }

  interface Child extends Base {
    static final Object C = new Object() {
      {
        System.out.println("child clinit");
      }
    };

    static void childStaticMethod() {
      System.out.println("child static method");
    }

    default Object childDefaultMethod() {
      System.out.println("child default method");
      return null;
    }
  }

    Child.childStaticMethod();
  }

}
```
and prints:
```
base clinit
child default method
child clinit
child static method
```

Godin:
:laughing: 9-ea+131 prints:
```
child clinit
child static method
```

Godin:
Reported as JI-9042842

Godin:
"slow path" implemented, so @marchof could you please review?

bjkail:
InstrumentingLoader uses 4-space indentation rather than tabs.  Otherwise, looks good to me.

Godin:
@bjkail reformatted `InstrumentingLoader`, thx.

marchof:
@Godin Thanks a lot for figuring all this out and finally resolving it! And sorry for my silence. I just started working through the change set and will drop some comments at the corresponding places.

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -71,6 +94,60 @@ private InstrSupport() {
 			| Opcodes.ACC_PRIVATE | Opcodes.ACC_STATIC;
 
 	/**
+	 * Name of the interface initialization method.
+	 *
+	 * According to Java Virtual Machine Specification <a href=
+	 * "https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.9-200"
```
marchof: ">" missing here (the whole quote becomes rendered as a link)

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
```diff
@@ -46,9 +47,10 @@
 	 *            callback to create the code that retrieves the reference to
 	 *            the probe array
 	 */
-	ProbeInserter(final int access, final String desc, final MethodVisitor mv,
+	ProbeInserter(final int access, final String name, final String desc, final MethodVisitor mv,
```
marchof: For consistency we should also add JavaDoc for name parameter and document clinit member.

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
```diff
@@ -0,0 +1,154 @@
...
+import org.jacoco.core.runtime.IExecutionDataAccessorGenerator;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+/**
+ * This strategy for Java 8 interfaces adds a static method requesting the probe
+ * array from the runtime, a static field to hold the probe array and adds code
+ * for its initialization into class initialization method.
```
marchof: Now as InterfaceFieldProbeArrayStrategy became stateful I would add a hint about instance usage: "A implementation instance of this interface must only be used to process the type it has been created for and must only be used once."

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
```diff
@@ -0,0 +1,154 @@
...
+					InstrSupport.INITMETHOD_NAME, InstrSupport.INITMETHOD_DESC,
+					true);
+			mv.visitVarInsn(Opcodes.ASTORE, variable);
+			return 1;
+		}
+	}
+
+	public void addMembers(final ClassVisitor cv, final int probeCount) {
+		createDataField(cv);
+		createGetMethod(cv, probeCount);
```
marchof: For consistency with other implementations shouldn't we name this `createInitMethod`, create method Name is `$jacocoInit`

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation.targets;
+
+public class BadCycleInterface {
+
+	public interface Base {
+		static final Object BASE_CONST = new Child() {
+			{
+				System.out.println("base clinit"); // $line-1$
```
marchof: Personally I don't like tests which print to sys out. This are now the only places we have in our test base.

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof `$jacocoInit` not always initializes field anymore, that's why I named creation method differently here, however seems that nowhere we state that it should do so - in documentation about implementation we state that it is used to initialize local variable with probe array. So indeed - why not rename.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
Godin: @marchof output was quite useful here to understand behavior when executed manually, but indeed now can be removed or replaced.

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeCounter.java
```diff
@@ -30,7 +30,7 @@
 	@Override
 	public MethodProbesVisitor visitMethod(final int access, final String name,
 			final String desc, final String signature, final String[] exceptions) {
-		if (!"<clinit>".equals(name)) {
+		if (!InstrSupport.CLINIT_NAME.equals(name)) {
```
marchof: Oops, here we probably have an bug since we support Java 8: We should only count non abstract methods in interfaces:

    if (!InstrSupport.CLINIT_NAME.equals(name) && (access & Opcodes.ACC_ABSTRACT) == 0) {
        methods = true;
    }

Also we should clarify hasMethod Javadoc: `true` if the class has other non-abstract methods than a static initializer

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
```diff
@@ -0,0 +1,154 @@
...
+		mv.visitLabel(alreadyInitialized);
+		mv.visitInsn(Opcodes.ARETURN);
+
+		mv.visitMaxs(Math.max(size, 2), 0); // Maximum local stack size is 2
+		mv.visitEnd();
+	}
+
+	private void createClinitMethod(final ClassVisitor cv,
+			final int probeCount) {
+		final MethodVisitor mv = cv.visitMethod(InstrSupport.INITMETHOD_ACC,
```
marchof: INITMETHOD_ACC works here, as the JVM ignores everything beside the ACC_STATIC flag. What about a separate CLINIT_ACC constant?

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.InputStream;
+
+import org.jacoco.core.data.ExecutionDataStore;
+import org.jacoco.core.data.SessionInfoStore;
+import org.jacoco.core.instr.Instrumenter;
+import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.RuntimeData;
+import org.jacoco.core.runtime.SystemPropertiesRuntime;
+
+public final class InstrumentingLoader extends ClassLoader {
```
marchof: This is probably subject of a separate PR: Shouldn't we use this loader directly in `ValidationTestBase` and drop `ValidationTestBase`?

marchof:
@Godin Ok, review done from my side, none of my 8 comments is a blocker.

P.S: Can you please check whether you see 8 comments? I only see them in the "Files changed" page, not on "Conversation".

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
marchof: I can imagine. Just hear myself preaching: "Remove sys.out or replace with assertion" ;-)

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof also thought about this

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
Godin: @marchof I suppose that you wanted to say "drop `BadCycleTestBase`". `ValidationTestBase` was quite hard to use in presence of multiple dependent classes, especially in presence of anonymous classes. And still hard to use, when multiple classes should be analyzed. Let's deal with this separately.

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeCounter.java
Godin: @marchof Both points are valid, however not a regression in this change as such. Also not so critical bug: previously and now - just addition of useless members.

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof not sure I understand where you suggest to place this - in `IProbeArrayStrategy` or in concrete implementation ? In the second case

> of this interface

sounds weird.

> process the type it has been created for

is valid for all implementations of `IProbeArrayStrategy` except `NoneProbeArrayStrategy`, because they all receive `className` and `classId`, so that even if others are stateless, the all not reusable and not just stateful `InterfaceFieldProbeArrayStrategy`.

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeCounter.java
Godin: @marchof in fact this even doesn't lead to creation of members, because factory checks number of probes before number of methods.

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
Godin: @marchof ticket created - #440

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: @marchof fixed

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
Godin: @marchof done

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof note added into `ProbeArrayStrategyFactory`

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof renamed

org.jacoco.core/src/org/jacoco/core/internal/instr/InterfaceFieldProbeArrayStrategy.java
Godin: @marchof added

Godin:
@marchof addressed all 8 comments in one way or another.

P.S. sometimes GitHub Web UI requires page reload and not just switching between tabs

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeCounter.java
Godin: @marchof oops - it does in case of clinit+abstract methods, so ticket created - #441

bjkail: @Godin If an interface has `<clinit>` but no default methods, then the class will have probes, so the current code will unnecessarily create a field, right?  I agree it's not a regression and could be fixed separately.

bjkail: Oops, web UI didn't reload as you said ;-).  The opened issue looks good to me.

Godin: @bjkail yup, already figure this out - https://github.com/jacoco/jacoco/pull/434#discussion_r74848553 :wink:

marchof:
@Godin Merged! Many thanks to @Godin and @bjkail for all the investigation which was possible to resolve this - even two JDK bugs were created for this ;-)

Good to know that in the meantime we have enough knowledge in the team to work on all tricky implementation details of JaCoCo::Core.

don-vip:
We use the new 0.7.8 snapshot in JOSM that includes this PR and can confirm it works! We have a few errors caused by java.lang.IncompatibleClassChangeError, I wonder if it's a Jacoco or a Groovy issue?

The stacktrace is 

    java.lang.IncompatibleClassChangeError: Method org.openstreetmap.josm.gui.mappaint.mapcss.Condition.createKeyCondition(Ljava/lang/String;ZLorg/openstreetmap/josm/gui/mappaint/mapcss/Condition$KeyMatchType;Lorg/openstreetmap/josm/gui/mappai
    	at org.openstreetmap.josm.gui.mappaint.mapcss.Condition$createKeyCondition.call(Unknown Source)
    	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
    	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
    	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:149)
    	at org.openstreetmap.josm.gui.mappaint.mapcss.KeyConditionTest.applies_1(KeyConditionTest.groovy:81)
    	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(java.base@9-ea/Native Method)
    	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(java.base@9-ea/NativeMethodAccessorImpl.java:62)
    	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(java.base@9-ea/DelegatingMethodAccessorImpl.java:43)

See https://josm.openstreetmap.de/jenkins/job/Java-EarlyAccess-JOSM/jdk=JDK9/lastCompletedBuild/testReport/org.openstreetmap.josm.gui.mappaint.mapcss/KeyConditionTest/applies_1/

Godin:
@don-vip There is no traces of JaCoCo in the provided stack trace, while there are clearly traces of Groovy, so very likely it is caused by Groovy. If this happens even without JaCoCo, then for sure it is caused by Groovy.

PR #411, opened 2016-07-29
Java 9 Support: Profile for 1.9?

Godin:
Question: Do we need to add a profile for Java 9? Like for the other java versions:

    <profile>
      <id>jdk18</id>
      <activation>
        <property>
          <name>jdk.version</name>
          <value>1.8</value>
        </property>
      </activation>
      <properties>
        <jvm.args>-XX:-FailOverToOldVerifier -Xverify:all</jvm.args>
      </properties>
    </profile>

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -52,6 +52,9 @@ case "$JDK" in
 9-ea)
   install_jdk $JDK9_EA_URL
   ;;
+9-ea-stable)
+  install_jdk $JDK9_EA_STABLE_URL
+  ;;
 esac
 
 # Do not use "~/.mavenrc" set by Travis (https://github.com/travis-ci/travis-ci/issues/3893),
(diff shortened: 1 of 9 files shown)


Godin:
@marchof could you please review? once this will be merged, I'll rebase #434 , so that it will be fully tested

marchof:
@Godin  LGTM, one remarks though: Unlike for other class file versions this is a quite limited test, only for core.test classes. With our other test builds we use the overall JaCoCo code base for the particular class file version.

Godin:
@marchof correct, however taking into account complications described above - as a first baby step this is better than nothing as it was prior to this change.

Also could be noted that during execution of our tests, we include only our classes into instrumentation, while inclusion of JDK classes would allow to test instrumentation at least on JDK classes, which I suppose are compiled into bytecode version 53.

PR #321, opened 2016-07-31
Exceptions cause missed branches in previous lines

Godin:
If a for loop is followed by line with a implicit exception the execution of one of the branches for the loop condition is reported as missed (1 of 2 branches missed):

	private void implicitExceptionAfterForLoop() {
		for (int i = 0; i < 3; i++) { // $line-implicitExceptionAfterForLoop.for$
			nop(); // $line-implicitExceptionAfterForLoop.loopbody$
		}
		ex(); // $line-implicitExceptionAfterForLoop.exception$
	}

This only happens for class files compiled with the JDK compilers, it works for class files compiled with Eclipse.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
@@ -103,6 +103,13 @@ public class ExceptionsTest extends ValidationTestBase {
 		assertLine("explicitExceptionFinally.finallyBlock",
 				ICounter.FULLY_COVERED);
 
+		// TODO(Godin): why fully covered? it looks strange, but the same in
+		// already presented cases above:
+		assertLine("implicitExceptionAfterBranch.if", ICounter.FULLY_COVERED, 1,
+				1);
+		assertLine("implicitExceptionAfterBranch.ifbody", ICounter.NOT_COVERED);
+		assertLine("implicitExceptionAfterBranch.exception",
+				ICounter.NOT_COVERED);
 	}
 
 }
--- a/org.jacoco.core/src/org/jacoco/core/internal/flow/LabelInfo.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/flow/LabelInfo.java
@@ -137,6 +137,9 @@ public final class LabelInfo {
 	 *            label to test
 	 * @return <code>true</code> if a probe should be inserted before
 	 */
+	// TODO(Godin): IMO misleading method name since it is not always used,
+	// e.g. not in
+	// org.jacoco.core.internal.flow.MethodProbesAdapter.visitJumpInsn()
 	public static boolean needsProbe(final Label label) {
 		final LabelInfo info = get(label);
 		return info != null && info.successor
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target03.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target03.java
@@ -44,6 +44,10 @@ public class Target03 implements Runnable {
 			implicitExceptionFinally();
 		} catch (StubException e) {
 		}
+		try {
+			implicitExceptionAfterBranch();
+		} catch (StubException e) {
+		}
 	}
 
 	private void implicitException() {
@@ -126,6 +130,13 @@ public class Target03 implements Runnable {
 		} finally { // $line-explicitExceptionFinally.finally$
 			nop(); // $line-explicitExceptionFinally.finallyBlock$
 		}
+	}
+
+	private void implicitExceptionAfterBranch() {
+		if (f()) { // $line-implicitExceptionAfterBranch.if$
+			return; // $line-implicitExceptionAfterBranch.ifbody$
+		}
+		ex(); // $line-implicitExceptionAfterBranch.exception$
 	}
 
 	public static void main(String[] args) {
--- a/org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesAdapter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesAdapter.java
@@ -108,7 +108,8 @@ public final class MethodProbesAdapter extends MethodVisitor {
 
 	@Override
 	public void visitJumpInsn(final int opcode, final Label label) {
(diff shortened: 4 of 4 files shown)


marchof:
@Godin It is. The situation is tricky here as beside the label for the new line we have another label for the control flow.

Godin:
@marchof do you mean that fix, which I attached to this thread, is incomplete? if so, wondering which test should be added to reveal this?

Godin:
So @marchof , what do you think about the fix proposed here?

marchof:
@Godin Can you please do the following sanity check: What is the total number of probes e.g. in org.jacoco.core without and with this patch? Should be a small increase only.

If we fix this in any case we need to increment exec file version :-/

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -103,6 +103,13 @@ public void testCoverageResult() {
 		assertLine("explicitExceptionFinally.finallyBlock",
 				ICounter.FULLY_COVERED);
 
+		// TODO(Godin): why fully covered? it looks strange, but the same in
```
marchof: @Godin Fully covered only in respect to instructions. But 1 branch missed, 1 branch covered. Makes sense?

Godin:
@marchof yes - I know that version should be bumped and this is unavoidable evil, with dramatic consequences for some of consumers of exec files, such as Jenkins Plugin, last time when we did this there were serious impacts. And that's why I'd like to be sure that we have a fix prior to exploring what needs to be done to actually promote this fix.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
Godin: @marchof indeed, this makes sense

marchof:
@Godin Maybe we prepare the fix and let it sit here until we have another reason for a breaking change.

Sidenote: For the filtering story I would consider reworking the whole probe story from the event API to the tree API. In the meantime we materialize a tree anyways and this might reduce complexity dramatically. On the other hand this might cause slight changes to probe positions/enumeration.

Godin:
@marchof I was thinking more about strategy: prepare the fix and let it sit here until we help downstream projects to deal with multiple JaCoCo versions - on example of Jenkins plugin this needs to be done in case of any change of exec-file version independently of underlying reasons for it, becase can't be expected that all projects hosted on a single Jenkins instance will upgrade version of JaCoCo synchronously, meaning that either users wouldn't use new JaCoCo version, either won't have their reports in Jenkins.

For `org.jacoco.core-0.7.7.201606060606.jar`: without this change 1756 probes, 1820 with.

So do you think that we have a fix here? If so, I'll add item about help for downstream projects into my todo-list.

marchof:
@Godin Yep, I think this is a valid fix. A increase of 3,5% in probes probably reflects exactly the cases we missed before. So we only need to fix your TODO comments.

Regarding exec file compatibility: I always tried to make clear that exec files are a temporary storage facility that make no guarantees about compatibility. It's an internal, undocumented binary format. If you want to integrate against a stable and documented file format, the XML report is currently the only format we can offer. We might add new formats in future, or even include analysis in the agent when data is dumped. But maintaining code that supports multiple exec file versions does not look very scalable to me. Beside exec file versions another common problem of our users is that we also need exact same class files available during analysis.

Godin:
@marchof yup, I have no doubts that content of exec-file should be considered as internal implementation details without guarantees about compatibility, even if I added some oil into the fire of its usage by relying on it in first version of SonarQube plugin several years ago when JaCoCo was just born. But past is past, author of Jenkins Plugin would also love to revert time back and base plugin on XML - https://github.com/jenkinsci/jacoco-plugin/pull/55#issuecomment-111468735 , but unfortunately in a real world that's not that easy, someone should anyway pay back and deal with current state of the things.

Having said all this, I'm not saying that we should provide this layer of compatibility. In fact it is not that hard to do on their side - few packaging tricks allow SonarQube plugin to deal with two versions of exec-file at the same time.

Godin:
@marchof Forgot to say in previous comment: even if at the root Jenkins / SonarQube plugins abuse reading exec-file, people would blindly blame JaCoCo. That's why IMO it is in our interests to help upstream projects to find one or another way thru this.

bjkail:
We implemented loading multiple versions of JaCoCo in our internal tooling so we can handle multiple versions of .exec files because it's difficult to coordinate upgrades for many teams at the same time.  Loading multiple versions of JaCoCo generally works ok for us, but it doesn't work as well when features are needed from later JaCoCo versions (e.g., bug fixes or Java 9 support) while maintaining .exec file compatibility.  It would be ideal for us if JaCoCo were somehow factored so that we could vary the module that deals with .exec file format but stay at the latest for all other code.  I guess that would require JaCoCo to have a stable internal interface to the code that handles the .exec file (or at least fallback gracefully when interacting with an older version of it), so I'm not sure that's practical or maintainable, but I thought I'd share our experience.

Godin:
Looks like that there is one more case that doesn't work properly even with those changes! Reproducer shows the problem with ECJ as well as with javac - no branches are shown as covered for the switch statement:
```
private static void implicitExceptionInSwitchCase() {
	switch (i2()) { // $line-implicitExceptionInSwitchCase.switch$
		case 2:
			ex(); // $line-implicitExceptionInSwitchCase.exception$
			break;
		default:
			nop();
			break;
	}
}
```

Godin:
For the clarity: we had the same behavior for all these cases in version 0.7.4 before implementation of #310 in version 0.7.5, so this is not about regression - simply additional cases that were missed in #310.

org.jacoco.core/src/org/jacoco/core/internal/flow/LabelInfo.java
```diff
@@ -137,6 +137,9 @@ public static boolean isMethodInvocationLine(final Label label) {
 	 *            label to test
 	 * @return <code>true</code> if a probe should be inserted before
 	 */
+	// TODO(Godin): IMO misleading method name since it is not always used,
```
alexrun: @Godin Do you have any plans on fixing this TODO?

dmurat:
For my own education, can you please confirm if incomplete coverage of Groovy assert statements is also a consequence of this issue?

PR #442, opened 2016-08-16
Add simple filtering to generated html report

Otanikotani:
In projects I work on it is important to be able to see only specific packages in the JaCoCo HTML report. Hence I added a very simple input field to the report to filter packages\classes by name. Here is an example of how it looks\works: http://prnt.sc/c6be4i

Please note that Total row is a lie after filtering ;)

P.S. I also added a cdn style from Materialize project to make report look more beautiful but I think it is not to be included there (this is how it looks with minimal effort: http://prnt.sc/c6bhso)


Godin:
Hi @Otanikotani ,

could you please clarify use case?

> In projects I work on it is important to be able to see only specific packages in the JaCoCo HTML report.

I believe that without any modifications in JaCoCo it is already possible to exclude/include only files from specific packages into report. Using an Ant Task you can even split them into different groups - see http://www.eclemma.org/jacoco/trunk/doc/ant.html#report

Otanikotani:
Hi @Godin ,

Thank you for sharing about grouping, I didn't know that there is a feature like that. Here are the reasons why I added a filter:

1. Excluding/including packages and classes currently require changes in configuration and creating the report again. Some projects are too big to do it often and I found it easier to be able to filter what I want to see at any time I have access to the html report without remaking it.

2. Primarily I work on writing tests for projects and we use JaCoCo to report (mostly via coverage) the work being done. Now, say there is a package with 100 classes in it (we are talking huge legacy projects here). Lets also say that 20 of those classes have a word "Row" in their names and they do something similar. What I would do as a test writer is I would write tests for these 20 classes and then report the work being done via JaCoCo coverage report. When we do reports we often need to add screenshots of the JaCoCo report to Jira to prove that classes have been covered. Now, the problem is that when I open the package with 100 classes, I cannot make a screenshot of specific 20 rows in the report. But when I filter it with word "Row" the table is exactly those 20 rows I wanted.

It could be that my use-case (making screenshots, sigh) is not very common, I understand. What can be done instead of adding a filter (as in this pull request) is adding a possibility to template the report. Of course, when ant is used I can just inject anything I want into index.html once it is generated (it is getting more difficult with maven) - but I feel like it would be a hacky way of doing it.

Godin:
@Otanikotani speaking about Maven - in [example.zip](https://github.com/jacoco/jacoco/files/432198/example.zip) you'll find example of project for which for which you can
* execute tests:
```
$ mvn clean verify
```
* generate full report:
```
$ mvn jacoco:report
```
* generate report containing `Row1` in name of class file:
```
mvn jacoco:report -Djacoco.includes=\*\*Row1\*\*
```
* generate report containing `Row1` or `Row3` in name of class file:
```
mvn jacoco:report -Djacoco.includes=\*\*Row1\*\*,\*\*Row3\*\*
```

Note that you don't need to change any files and re-run tests to generate reports containing new set of files. I assume (strongly believe) that time to generate report is much smaller and negligible compared to time to execute tests. If this is not the case, then please provide exact measurements for your case.

This way you can attach to Jira ticket whole report containing required classes without need in a screenshot. IMO this fully covers your use case, isn't it?

Godin:
Forgot to say that CSV report is easy to filter out:
```
$ cat target/site/jacoco/jacoco.csv | grep "Row1"
example,default,Row1,3,0,0,0,1,0,1,0,1,0
```
And most likely application of XSLT stylesheet on XML report already gives "templating".

alan125:
I would like to vote for the PR, since I have the same issues as @Otanikotani . The project I'm working on isn't maven based, but ant based. Also, the project I'm working in has some complex configurations that I'm not supposed to touch. I agree that CSV and XML reports can be filtered, but I mostly use the HTML report, and having filtering would be really beneficial.

yeikel:
I think this can be a nice feature just for convenience if the project is large enough and the number of packages and classes is large.

PR #441, opened 2016-08-16
Do not add useless members into Java 8 interfaces that have only clinit and abstract methods

Godin:
`ProbeCounter.hasMethods` should be true for Java 8 interfaces only in case of presence of non-abstract methods.


Godin:

bjkail:
Looks good to me.  As a very minor comment, better grammar for changes.html would be "interfaces that have" ("that" is for discriminating, "which" is for annotating) :-).

Godin:
@bjkail thx, fixed

marchof:
@Godin Thx!

PR #440, opened 2016-08-16
Improve ValidationTestBase

marchof:
* use `InstrumentingLoader`
* get rid of `BadCycleTestBase`


marchof:
@Godin Lots of redundant code deleted... can you please review?

BTW I didn't add a change log as this is internal test code only.

marchof:
@Godin Damn, test failed for Java 9-ea-stable :-/ Looks like the type is not considered by the InstrumentingLoader.

Godin:
@marchof it looks like the same problem for which we have condition in test and which was reported to OpenJDK as JI-9042842 - https://github.com/jacoco/jacoco/pull/434#issuecomment-239648730 , for some reason on JDK 9 EA superinterface is not initialized

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/BadCycleInterfaceTest.java
```diff
@@ -18,26 +18,23 @@
...
 	}
 
 	@Test
 	public void test() throws Exception {
-		loader.loadClass(BadCycleInterface.Child.class.getName())
-				.getMethod("childStaticMethod").invoke(null);
-
-		analyze(BadCycleInterface.Child.class);
-
+		assertLine("baseclinit", ICounter.FULLY_COVERED);
```
Godin: @marchof more specifically - this assertion should be conditional

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
```diff
@@ -16,7 +16,7 @@
 	public interface Base {
 		static final Object BASE_CONST = new Child() {
 			{
-				Stubs.nop("base clinit"); // $line-1$
+				Stubs.nop("base clinit"); // $line-baseclinit$
```
Godin: @marchof I used numbers to denote an actual execution order. So I'm wondering why not leave them as is? As an alternatives - `$line-1-baseclinit$` or explicit comments.

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
```diff
@@ -21,31 +20,53 @@
...
 		runtime.startup(data);
+		this.instrumenter = new Instrumenter(runtime);
+	}
+
+	public InstrumentingLoader(Class<?> target) throws Exception {
+		this(new SystemPropertiesRuntime(), target.getPackage().getName(),
+				target.getClassLoader());
+	}
+
+	public InstrumentingLoader() throws Exception {
```
Godin: @marchof this constructor is not used

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
```diff
@@ -21,31 +20,53 @@
...
+	private final String scope;
+	private final ClassLoader delegate;
 
-	public InstrumentingLoader() throws Exception {
-		data = new RuntimeData();
-		runtime = new SystemPropertiesRuntime();
+	private final RuntimeData data;
+	private final Instrumenter instrumenter;
+
+	public InstrumentingLoader(IRuntime runtime, String scope,
```
Godin: @marchof after removal of unused constructor this one will be used only from the another remaining

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
```diff
@@ -21,31 +20,53 @@
...
+		this.runtime = runtime;
+		this.scope = scope;
+		this.delegate = delegate;
+		this.data = new RuntimeData();
 		runtime.startup(data);
+		this.instrumenter = new Instrumenter(runtime);
+	}
+
+	public InstrumentingLoader(Class<?> target) throws Exception {
+		this(new SystemPropertiesRuntime(), target.getPackage().getName(),
```
Godin: @marchof probably not so important, but without trailing dot this works as regexp `org.*` and not as `org\..*`

org.jacoco.core.test/src/org/jacoco/core/test/TargetLoader.java
```diff
@@ -52,17 +52,24 @@ public TargetLoader() {
 	}
```
Godin: @marchof  method `TargetLoader.add(Class<?> source)` is unused. Moreover seems that after this change there is no need in map `classes` in `TargetLoader` since it will be used only to load a single class, what is in fact stated in his javadoc already :wink:

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
marchof: Right, I removed the too specific default constructor. Still want to keep the constructor where all delegates can be injected to enable this utility for future scenarios (other runtimes, other class loaders).

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
marchof: If execution order is relevant here it should be asserted. Should I try to record and assert execution order?

org.jacoco.core.test/src/org/jacoco/core/test/TargetLoader.java
marchof: Should be consistent now. But kept the classes map otherwise we need to assert that only one class is inserted.

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
marchof: Is this used as a regex? Only usage is `name.startsWith(scope)`, right?

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/BadCycleInterfaceTest.java
```diff
@@ -18,26 +18,21 @@
...
 			// JDK-9042842
-			assertLine("2", ICounter.NOT_COVERED);
-		} else {
-			assertLine("2", ICounter.FULLY_COVERED);
+			assertLine("baseclinit", ICounter.FULLY_COVERED);
 		}
-		assertLine("4", ICounter.FULLY_COVERED);
+		assertLine("childclinit", ICounter.FULLY_COVERED);
+		assertLine("childdefault", ICounter.FULLY_COVERED);
+		assertLine("childstatic", ICounter.FULLY_COVERED);
```
Godin: @marchof 
For JDK 9 EA sequence is
```
childclinit
childstatic
```
for others
```
baseclinit
childdefault
childclinit
childstatic
```

marchof: Ok, thx. Tried to fix the build.

Godin: @marchof and I think that we should keep assertion for JDK 9 EA that lines are not executed - this way we can easily catch the moment when this will be fixed and will be able to remove conditional behavior:

```java
if (System.getProperty("java.version").startsWith("9-ea")) {
  // JDK-9042842
  assertLine("baseclinit", ICounter.NOT_COVERED);
  assertLine("childdefault", ICounter.NOT_COVERED);
} else {
  assertLine("baseclinit", ICounter.FULLY_COVERED);
  assertLine("childdefault", ICounter.FULLY_COVERED);
}
assertLine("childclinit", ICounter.FULLY_COVERED);
assertLine("childstatic", ICounter.FULLY_COVERED);
```

org.jacoco.core.test/src/org/jacoco/core/test/InstrumentingLoader.java
Godin: @marchof I mentioned "regexp" just for demonstration of a difference when scope has trailing dot and when not.

Given `target` in a package `foo` and some class `baz` in package `foobar`:
```
"foobar.baz".startsWith("foo")
```
vs
```
"foobar.baz".startsWith("foo.")
```

But as was mentioned earlier - this is probably minor.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/targets/BadCycleInterface.java
Godin: @marchof I think that full execution order here has an interest only as a demonstration. What really matters here and in `BadCycleClass` - is that some method of a child can be executed before its `clinit` method. And we test exactly this - such situation is properly handled and we do not rely on false assumption that `clinit` executed before everything else. So indeed it seems reasonable to have an assertion about this, but assertion for full sequence is probably overkill.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/BadCycleInterfaceTest.java
```diff
@@ -26,12 +26,15 @@ public BadCycleInterfaceTest() throws Exception {
 
 	@Test
 	public void test() throws Exception {
-		if (!System.getProperty("java.version").startsWith("9-ea")) {
+		if (System.getProperty("java.version").startsWith("9-ea")) {
 			// JDK-9042842
+			assertLine("baseclinit", ICounter.NOT_COVERED);
```
Godin: @marchof excuse me that I provided wrong snippet - actually since base interface is not itialized in JDK 9 EA due to JI-9042842, it doesn't pass thru classloader and hence nor instrumented nor analyzed, resulting in `ICounter.EMPTY`. This time I checked locally, so hopefully this finally will be the last attempt :)

Godin: `ICounter.EMPTY` for "baseclinit" only

marchof:
@Godin Compiles, let's merge! ;-)

Seriously: if you don't see any issues let's merge this one. I work on execution detection in a separate PR.

Godin:
@marchof fine by me. Thank you for this cleanup and simplification!

PR #445, opened 2016-08-22
Add FAQ entry about offline instrumentation

marchof:
It seems to be a common mistake to instrument classes offline but not to add jacocoagent.jar.


marchof:
@Godin Can you please review?

Godin:
@marchof LGTM!

PR #448, opened 2016-09-08
Update comments about "bad cycles" with interfaces

Godin:
Our current example of "bad cycles" with interfaces ( https://github.com/jacoco/jacoco/pull/434#issuecomment-239643789 ) - is more a demonstration of correct change of behavior between JDK8 and JDK9 than a correct example of "bad cycles" with interfaces as stated in [JDK-8164302](https://bugs.openjdk.java.net/browse/JDK-8164302). We should add correct example.

### DIFF
--- a/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/BadCycleInterfaceTest.java
+++ b/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/BadCycleInterfaceTest.java
@@ -26,22 +26,23 @@ public class BadCycleInterfaceTest extends ValidationTestBase {
 
 	@Test
 	public void test() throws Exception {
-		if (System.getProperty("java.version").startsWith("9-ea")) {
-			// JDK-9042842
-			assertLine("baseclinit", ICounter.EMPTY);
-			assertLine("childdefault", ICounter.NOT_COVERED);
-			assertLogEvents("childclinit", "childstaticmethod");
-		} else {
+		if (System.getProperty("java.version").startsWith("1.8")) {
+			// Incorrect interpetation of JVMS 5.5 in JDK 8 causes a default
+			// method to be called before the static initializer of an interface
+			// (see JDK-8098557 and JDK-8164302):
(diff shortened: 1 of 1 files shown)


Godin:
In fact we don't need to wait resolution of [JDK-8163969](https://bugs.openjdk.java.net/browse/JDK-8163969) to add correct test of "bad cycles" with interfaces, because exception provokes crash and we can use scenario without exception.

Note that this new test also reveals difference in initialization order - most likely [JDK-8062114](https://bugs.openjdk.java.net/browse/JDK-8062114).

marchof:
@Godin After fix of JRE I'm not sure whether we still can get "bad cycles". By this I mean a default method invocation *before* clinit has been called at all.

Are you sure in example BadCycleInterface the class stack isn't something like:

    Child.childDefaultMethod()
      Base.clinit()
        Child.clinit()

In this case it wouldn't be a problem for JaCoCo as we would initialize the probes at the very beginning of Child.clinit().

BTW, this is why I introduced the exception -- to see the call hierarchy ;-)

Godin:
@marchof you probably right. After you said this, I remembered that `System.out` wasn't very useful during initial creation of test, since it lacks stack information, and so log events too. Let me check. Anyway existing test shows that we should preserve slow path in `InterfaceFieldProbeArrayStrategy` for support of JDK 8, where issue wasn't fixed.

Godin:
@marchof You were right - new example doesn't provoke usage of slow path, as well as your example from https://github.com/jacoco/jacoco/pull/434#issuecomment-239427471 
So I suppose combination of fixes [JDK-8043275](https://bugs.openjdk.java.net/browse/JDK-8043275) and [JDK-8098557](https://bugs.openjdk.java.net/browse/JDK-8098557) indeed removes possibility of "bad cycles" with interfaces starting with JDK 9.
I reworked this PR to slightly update comments.

marchof:
@Godin Looks got to me! Now it makes sense in respect to the JDK issues/fixes.

BTW, we definitely need a separate build for every Java update version - only 134 jobs required so far ;-))) https://en.wikipedia.org/wiki/Java_version_history

Issue #449, opened 2016-09-11
Wrong coverage rate due to loss of precision


MEZk:
JaCoCo reports wrong coverage rate due to loss of precision.

### Steps to reproduce

JaCoCo version: 0.7.7.201606060606
Operating system: Linux 4.4.0-21-generic #37-Ubuntu SMP x86_64 x86_64 x86_64 GNU/Linux
Tool integration: Maven
maven-jacoco-plugin configuration: [jacoco-maven-plugin-config.txt](https://github.com/jacoco/jacoco/files/465927/jacoco-maven-plugin-config.txt)
### Expected behaviour

I assume that JaCoCo uses the following formulas to calculate final coverage rates:

_instructions coverage rate = ((total instructions count - missed instructions count) / total instructions count) * 100%_
_branches coverage rate = ((total branches count - missed branches count) / total branches count) * 100%_

For example in JaCoCo coverage report I have the following numbers:

_total instructions count  = 60501
missed instructions count = 506_

and
 
_total branches count  = 7652
missed branches count = 78_

So, the instructions coverage rate and branches coverage rate will be:

_instructions coverage rate = ((60501 - 506) / 60501) * 100% = 0.991636501 * 100% = 99.1636501...%
branches coverage rate = ((7652 - 78) / 7652) * 100% = 0.989806586 * 100% =  98.9806586...%_

### Actual behaviour

When I execute _mvn clean verify_ , JaCoCo fails the build due to the fact that coverage rate for branches is 0.98 which is lower than the specified threshold (see [jacoco-maven-plugin-config.txt](https://github.com/jacoco/jacoco/files/465927/jacoco-maven-plugin-config.txt)):
- for instructions: 0.99
- for branches: 0.99

However, in the coverage report I see that:
instructions coverage rate = 99%
branches coverage rate = 99%

### Problem description

There is a wrong coverage rate in the report due to loss of precision. 
jacoco:check phase uses correct values which are calculated with the help of formulas specified above. As a result it is unclear for user to determine where is a problem: reports look OK, but jacoco:check fails the build.

### Proposal 

Use equal precision in both report coverage rates and check phase.

marchof:
You can specify any precision for the check goal, while the report percentage column always display percentage with 2 digit precision ("98%").

The issue here is probably more with rounding: In your example 98.9806586 is rounded up to 99%. That's why you assume a too high coverage ratio for the check.

[...]

PR #452, opened 2016-09-24
Use RoundingMode#FLOOR for percentages in HTML report

Godin:
As per https://github.com/jacoco/jacoco/issues/449#issuecomment-246319827


Godin:

PR #457, opened 2016-10-27
Use jacoco.org domain

marchof:
As we're handing over eclemma.org to Eclipse foundation we should use jacoco.org only in JaCoCo URLs.


marchof:
@Godin Can you please review?

Godin:
@marchof merged after few additional replacements

marchof:
@Thx!

PR #466, opened 2016-12-03
No need to sign JARs using maven-jarsigner-plugin

Godin:
The primary use case for signing JaCoCo artifacts was the EclEmma plugin. Next release of EclEmma under Eclipse Foundation umbrella will use JaCoCo bundles from [Eclipse Orbit](http://download.eclipse.org/tools/orbit/downloads/), which signed by eclipse.org anyway. So that no need to sign anymore during our releases.

Initial discussion: https://groups.google.com/d/msg/jacoco-dev/hTLvWPXqaUg/fKoDA3zqBAAJ

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -48,6 +48,13 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/427">#427</a>).</li>
 </ul>
 
+<h3>Non-functional Changes</h3>
+<ul>
+  <li>Released JaCoCo JARs are not signed any more. Signed versions of JaCoCo are
+      now available from the Eclipse Orbit project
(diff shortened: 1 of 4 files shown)


marchof:
@Godin I found more occurences for signing which probably should be removed:

- jacoco/pom.xml (jarsigner.skip property)
- org.jacoco.agent.rt/pom.xml (jarsigner.skip property)

As discussed on ML: We mention this in changes.html

Godin:
@marchof should be fine now

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -48,6 +48,12 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/427">#427</a>).</li>
 </ul>
 
+<h3>Non-functional Changes</h3>
+<ul>
+  <li>Stop signing release JARs
```
Godin: indeed, fixed

PR #468, opened 2016-12-09
Simplify numbering of versions

Godin:
JARs of releases that we we publish into Maven central repository should have major, minor and patch components in version, but not qualifier, i.e. simply `0.7.8`. Qualifier should remain in MANIFEST files.

Initial discussion: https://groups.google.com/d/msg/jacoco-dev/HNWPzop4ZI4/pvmfWsSSCAAJ


marchof:
@Godin Beside the change log, is there anything else we need to modify for this? Or will this happen as part of the release process?

Godin:
@marchof so far I believe that this affects only release procedure

marchof:
@Godin Please feel free to merge after checks are green.

PR #474, opened 2016-12-12
Runtime initialize non-constant static finals

bjkail:
In my case, `ExecutionDataWriter.FORMAT_VERSION` is the problematic constant.  We workaround by using reflection to access the field so that javac won't inline the value from the version of the library we use to compile.

I reviewed the `org.jacoco.doc/target/apidocs/constant-values.html`, and `JaCoCo.ASM_API_VERSION` seemed like the only other constant that was likely to cause similar problems for anyone.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
@@ -142,7 +142,7 @@ public class ExecutionDataReaderWriterTest {
 		buffer.write(ExecutionDataWriter.BLOCK_HEADER);
 		buffer.write(0xC0);
 		buffer.write(0xC0);
-		final char version = ExecutionDataWriter.FORMAT_VERSION - 1;
+		final char version = (char) (ExecutionDataWriter.FORMAT_VERSION - 1);
 		buffer.write(version >> 8);
 		buffer.write(version & 0xFF);
 		createReader().read();
--- a/org.jacoco.core/src/org/jacoco/core/JaCoCo.java
+++ b/org.jacoco.core/src/org/jacoco/core/JaCoCo.java
@@ -13,8 +13,6 @@ package org.jacoco.core;
 
 import java.util.ResourceBundle;
 
-import org.objectweb.asm.Opcodes;
-
 /**
  * Static Meta information about JaCoCo.
  */
@@ -28,9 +26,6 @@ public final class JaCoCo {
 
 	/** Name of the runtime package of this build */
 	public static final String RUNTIMEPACKAGE;
-
-	/** ASM API version */
-	public static final int ASM_API_VERSION = Opcodes.ASM5;
 
 	static {
 		final ResourceBundle bundle = ResourceBundle
--- a/org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java
+++ b/org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java
@@ -24,7 +24,12 @@ public class ExecutionDataWriter implements ISessionInfoVisitor,
 		IExecutionDataVisitor {
 
 	/** File format version, will be incremented for each incompatible change. */
-	public static final char FORMAT_VERSION = 0x1007;
+	public static final char FORMAT_VERSION;
+
+	static {
+		// Runtime initialize to ensure javac does not inline the value.
+		FORMAT_VERSION = 0x1007;
+	}
 
 	/** Magic number in header for file format identification. */
 	public static final char MAGIC_NUMBER = 0xC0C0;
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ClassInstrumenterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ClassInstrumenterTest.java
@@ -13,7 +13,6 @@ package org.jacoco.core.internal.instr;
 
 import static org.junit.Assert.assertNull;
 
-import org.jacoco.core.JaCoCo;
 import org.junit.Before;
 import org.junit.Test;
 import org.objectweb.asm.ClassVisitor;
@@ -29,7 +28,7 @@ public class ClassInstrumenterTest implements IProbeArrayStrategy {
 	@Before
 	public void setup() {
 		instrumenter = new ClassInstrumenter(this, new ClassVisitor(
-				JaCoCo.ASM_API_VERSION) {
+				InstrSupport.ASM_API_VERSION) {
 		});
 	}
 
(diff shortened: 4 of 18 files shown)


Godin:
@bjkail to confirm my understanding of
> calling code must be recompiled when upgrading to a new version of the
library even if the caller would otherwise not require any changes

you are doing something like
```
if (JaCoCo.FORMAT_VERSION != 0x1007) {
  throw new RuntimeException();
}
```
this condition is always false since `javac` inlines `FORMAT_VERSION`. And from here comes the need of recompilation?

bjkail:
We have a client library that uses JaCoCo, but clients need to be able to choose their own JaCoCo version.  We store execution data on a per-format folder basis, and the client library uses `ExecutionDataWriter.FORMAT_VERSION` so it can store new execution data files in the right place or so it only attempts to load execution data files using `ExecutionDataReader` that it will be able to read.

Godin:
Okay, so very similar to my example: due to inlining client library can claim that uses older format, while runtime dependency changed and it actually uses new format.

org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java
```diff
@@ -24,7 +24,11 @@
 		IExecutionDataVisitor {
 
 	/** File format version, will be incremented for each incompatible change. */
-	public static final char FORMAT_VERSION = 0x1007;
+	public static final char FORMAT_VERSION;
+
+	static {
+		FORMAT_VERSION = 0x1007;
+	}
```
Godin: I can do this by myself, but may I ask you to change back executable bit for this file?
GitHub tells me that old mode was 100644 and new one is 100755:
mode

org.jacoco.core/src/org/jacoco/core/JaCoCo.java
```diff
@@ -30,14 +30,15 @@
 	public static final String RUNTIMEPACKAGE;
 
 	/** ASM API version */
-	public static final int ASM_API_VERSION = Opcodes.ASM5;
+	public static final int ASM_API_VERSION;
```
Godin: I think we should add a comment to prevent accidental change back.

org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java
```diff
@@ -24,7 +24,11 @@
 		IExecutionDataVisitor {
 
 	/** File format version, will be incremented for each incompatible change. */
-	public static final char FORMAT_VERSION = 0x1007;
+	public static final char FORMAT_VERSION;
```
Godin: And comment here too.

org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
```diff
@@ -142,7 +142,7 @@ public void testInvalidVersion() throws IOException {
 		buffer.write(ExecutionDataWriter.BLOCK_HEADER);
 		buffer.write(0xC0);
 		buffer.write(0xC0);
-		final char version = ExecutionDataWriter.FORMAT_VERSION - 1;
+		final char version = (char) (ExecutionDataWriter.FORMAT_VERSION - 1);
```
Godin: Interesting that explicit cast not needed in case of inlined value. Learned something new today :satisfied:

bjkail:
I believe I fixed the issues, thanks.  Sorry for the sloppiness with the author/committer email and file modes.

org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
Godin: There are implicit casts: `char -> int -> subtraction` , in first case value after subtraction is known at compile time and so one more safe implicit cast `int -> char`. In second case value not known and so javac says "possible lossy conversion from int to char".

marchof:
To make it more explicit that this is not a constant value I would prefer a static method instead:

    public static char ExecutionDataWriter.getFormatVersion()

This breaks the API but people using it currently ended up with broken code anyways.

org.jacoco.core/src/org/jacoco/core/JaCoCo.java
```diff
@@ -30,14 +30,16 @@
 	public static final String RUNTIMEPACKAGE;
 
 	/** ASM API version */
-	public static final int ASM_API_VERSION = Opcodes.ASM5;
```
marchof: This should only get compiled into JaCoCo classes and not be used externally. For internal classes inlining is perfectly ok because the compiled version of JaCoCo has specifically been tested for this ASM version.

Instead of changing the initialization scheme here I would rather remove it from the API and move it to some internal classes (e.g. InstrSupport).

bjkail: I've added a separate commit to move ASM_API_VERSION to InstrSupport.

bjkail:
@marchof My code uses `FORMAT_VERSION` via reflection and is not broken, and other code that uses `FORMAT_VERSION` and uses the same JaCoCo version at compilation and runtime is not broken.  Replacing `FORMAT_VERSION` with `getFormatVersion()` means more work for me to support multiple JaCoCo versions, so I would rather drop the pull request than do that.

marchof:
@bjkail Many thanks for moving `ASM_API_VERSION`!

I see your point about compatibility: Another option would be to leave `FORMAT_VERSION` as is and deprecate it. In addition we add a new method `getFormatVersion()`. If I understand your case correctly you need to stick with reflection anyways to support old JaCoCo versions.

But I can also live with your PR as is.

bjkail:
@marchof As soon as a JaCoCo is released with non-constant `FORMAT_VERSION`, I can updated my code to compile against that version, which means javac won't inline the value, so my .class file will contain a reference to `FORMAT_VERSION`, which will resolve the value at runtime regardless of whether the field is actually constant, so I can remove the reflection.  The test I suggested to Evgeny did basically the same: compile a test class that uses `FORMAT_VERSION`, then use TargetLoader to load it alongside different versions of `ExecutionDataWriter` transformed via ASM to have `FORMAT_VERSION` with value 0 as either constant (simulating current JaCoCo) or non-constant (simulating future JaCoCo).

Godin:
@bjkail Not sure I fully understand why "more work". But in any case let's keep this as a field. @marchof other options seems overkill to me. I updated changelog, will merge after pass of builds.

@bjkail A kind of off topic: since you support multiple versions, may I ask out of curiosity - down to which one? Especially that we talk about version of format that last time was changed in 0.7.5 (2015/05/24)?

bjkail:
@Godin By "more work" I meant that I would have to update the current reflection to look for both `getFormatVersion()` and `FORMAT_VERSION` (assuming the former replaced the latter).  It's not hard or complex work, it's just more :-).

I know 0.7.4 is still heavily used (the most recent for 0x1006), but I've seen older versions in use too.  I have not encountered 0.4.x (0x1005) or older.  Older versions are used both due to risk aversion and because upgrading creates a "flag day" for components that want to merge/intersect their execution data (if one component moves then they all need to move).

PR #475, opened 2016-12-15
Add missing test case that is described in documentation

Godin:
This adds test case that is described on page http://www.eclemma.org/jacoco/trunk/doc/flow.html :

> ..  implicit exceptions from other instructions than method invocations (e.g. NullPointerException ...

And removes misleading comments

> Currently no coverage at all, as we don't see when a block aborts somewhere in the middle.

from the cases, where we actually see some coverage, when block aborts in the middle.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
@@ -27,9 +27,15 @@ public class ExceptionsTest extends ValidationTestBase {
 	@Test
 	public void testCoverageResult() {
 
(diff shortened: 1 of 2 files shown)


marchof:
@Godin Not sure why a single build (9-ea-stable) fails. Can you please check and merge?

Godin:
@marchof from time to time `testInvalidCommand(org.jacoco.agent.rt.internal.output.TcpClientOutputTest)` fails in Travis, I don't know why - race condition, timeouts or something like that.

marchof:
@Godin can we trigger a manual build so we can merge this?

Godin:
@marchof yup - I restarted build already

PR #462, opened 2016-12-29
java.lang.ClassFormatError: Short length on BootstrapMethods in class file

Godin:
I am trying to integrate JaCoCo for code coverage of [Eclipse Collections](http://www.eclipse.org/collections/)

However the build fails with the below error:
`java.lang.ClassFormatError: Short length on BootstrapMethods in class file org/eclipse/collections/impl/collector/Collectors2Test`

The [Collectors2Test.java](https://github.com/eclipse/eclipse-collections/blob/0db19cf63e53dafeb59ac46b582fa0a266c600c5/unit-tests/src/test/java/org/eclipse/collections/impl/collector/Collectors2Test.java) is 1748 lines long with very small test methods. So, this should not cause any issues as they are well below 64KB method limit of Java.
_If I break up this class in two separate files then the build goes through fine._  Even after online research I am not able to find a way to fix the Travis build with JaCoCo integration and _without_ splitting the Collectors2Test.java. The build goes through fine in IntelliJ.

The project is setup with Maven build and has surefire plugin. `argLine` is set in relevant modules where JaCoCo test coverage is run.

### GitHub repos and build links:

1. Eclipse Collections master
   [Repo](https://github.com/eclipse/eclipse-collections)
   [Travis build (passing)](https://api.travis-ci.org/jobs/175779436/log.txt?deansi=true)

2. JaCoCo integration with Collectors2Test split in two:
   [Repo](https://github.com/nikhilnanivadekar/eclipse-collections/tree/enableJaCoCo)
   [JaCoCo integration commit](https://github.com/eclipse/eclipse-collections/pull/166/commits)
   [Travis build (passing)](https://api.travis-ci.org/jobs/176257417/log.txt?deansi=true)

3. JaCoCo integration with Collectors2Test not split:
   [Repo](https://github.com/nikhilnanivadekar/eclipse-collections/tree/buildFailureJaCoCo)
   [JaCoCo integration commit](https://github.com/nikhilnanivadekar/eclipse-collections/commit/62c5ca4cf04c6ab692f135bf1ef03c1834d5b92a)
   [Travis build (**failing**)](https://api.travis-ci.org/jobs/176584223/log.txt?deansi=true)

### Steps to reproduce

1. Fork and clone the eclipse collections repo
2. Apply commit 62c5ca4cf04c6ab692f135bf1ef03c1834d5b92a on top of it
3. Enable [Travis](https://travis-ci.org/) integration for your fork.
4. Push the commit from step 2
5. Check the unit test build. It generally takes about 5-6 mins to fail.

### DIFF
--- a/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
+++ b/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
+++ b/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
@@ -23,6 +23,7 @@ import org.jacoco.core.instr.Instrumenter;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.RuntimeData;
 import org.jacoco.core.runtime.SystemPropertiesRuntime;
+import org.jacoco.core.test.TargetLoader;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -90,16 +91,8 @@ public class AsmBugTest {
 	private static int run(final String className, final byte[] bytes)
 			throws ClassNotFoundException, NoSuchMethodException,
 			InvocationTargetException, IllegalAccessException {
-		return (Integer) new ClassLoader() {
-			@Override
-			public Class<?> loadClass(String name)
-					throws ClassNotFoundException {
-				if (className.equals(name)) {
-					return defineClass(name, bytes, 0, bytes.length);
-				}
-				return super.loadClass(name);
-			}
-		}.loadClass(className).getMethod("run").invoke(null);
+		return (Integer) new TargetLoader().add(className, bytes)
+				.getMethod("run").invoke(null);
 	}
 
 	/**
--- a/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
+++ b/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
@@ -1,0 +1,133 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2016 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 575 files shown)


Godin:
I chose to not add exact example with lambda deserialization as it depends on javac and its internals, and actually not so trivial to also trigger resizing in ASM. But added test that performs a bit of bytecode generation instead. As you can see - it fails before ASM upgrade and passes after. So @marchof could you please review this change?

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+		final byte[] instrumented = instrumenter.instrument(original,
+				className);
+		assertEquals(42, run(className, instrumented));
+	}
+
+	private static int run(final String className, final byte[] bytes)
+			throws ClassNotFoundException, NoSuchMethodException,
+			InvocationTargetException, IllegalAccessException {
+		return (Integer) new ClassLoader() {
```
marchof: We can also use `TargetLoader` here.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.objectweb.asm.Opcodes;
+
+/**
+ * Test of ASM bug <a href=
+ * "http://forge.ow2.org/tracker/?func=detail&aid=317748&group_id=23&atid=100023">#317748</a>
+ * that caused
+ * {@code java.lang.ClassFormatError: Short length on BootstrapMethods in class file}
+ * during instrumentation.
+ */
+public class AsmBugTest {
```
marchof: We encountered several ASM bug in the past. So the test name looks a bit too generic for me. Also we tend to name our test cases like what is tested, not what is a bug. I would prefer something like `BootstrapMethodReferenceTest`.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
Godin: Done.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AsmBugTest.java
Godin: Done.

nikhilnanivadekar:
@Godin Thank you very much for fixing this. 
What are the plans for a JaCoCo version release so that I can upgrade it in Eclipse Collections and remove the workaround?

Godin:
@nikhilnanivadekar we can't tell you the exact date yet - there is no strict plan so far.

And as was said in https://github.com/eclipse/eclipse-collections/pull/166#discussion_r88514734 : EclEmma Eclipse Plugin also needs an update, otherwise execution of tests with coverage in Eclipse using EclEmma will be failing without the current workaround. This update will take some time as we need to not only release JaCoCo, but also promote it as well as new ASM version thru Eclipse CQs, followed by a first release of EclEmma under Eclipse umbrella.

All in all my advice for you will be - no need to rush especially that workaround doesn't cause any troubles, a little of patience and stay tuned for announcements about releases :wink:

Godin:
For the record: here is a ticket about update for EclEmma - https://bugs.eclipse.org/bugs/show_bug.cgi?id=510059

PR #211, opened 2017-01-04
Generate OSGi Manifests

Godin:
I removed IDE dependencies with commit ae311b988b180df4f52295c0368b17d58ec8ce86. Now we can create proper MANIFEST.MF files during the build with bnd: http://felix.apache.org/site/apache-felix-maven-bundle-plugin-bnd.html


Godin:

There are some notable differences in resulting files, in particular:
* no generation of OSGi attributes for all `*.test` and removal of their existing `MANIFEST.MF` - anyway they are not part of distribution and hence unused
* removed `MANIFEST.MF` from `jacocoant.jar` (the one that has all dependencies included) that previously had OSGi attributes, while is not OSGi bundle
* all internal packages are exported with `x-internal:=true` directive
* added `uses` directive
* added attribute `Require-Capability`

marchof:
@Godin First of all many thanks for picking this up!

Regarding `jacocoant.jar`: I agree this is not a bundle. But from my point of view every JAR should have a `MANIFEST.MF` at least with basic meta information. Can't we add a `MANIFEST.MF` without any OSGi header like in `jacocoagent.jar`?

I wanted to test the new bundles with EclEmma but I failed to include them with the EclEmma build as the build now picks up the JaCoCo bundles from orbit. Any ideas how I can use local JaCoCo snapshots builds with EclEmma?

Godin:
@marchof 

> Regarding jacocoant.jar: I agree this is not a bundle. But from my point of view every JAR should have a MANIFEST.MF at least with basic meta information. Can't we add a MANIFEST.MF without any OSGi header like in jacocoagent.jar?

Well, removal was easy. I'll try to add basic entries. `Manifest-Version`, `Implementation-Title`, `Implementation-Vendor` and `Implementation-Version` would be enough?

> I wanted to test the new bundles with EclEmma

Nice idea. I did only diff for `MANIFEST.MF` files previously.

> but I failed to include them with the EclEmma build as the build now picks up the JaCoCo bundles from orbit. Any ideas how I can use local JaCoCo snapshots builds with EclEmma?

You can revert https://github.com/eclipse/eclemma/commit/330ec846686c63dfc9710e7619f9f92d714a7c9b locally:
```
git revert 330ec846686c63dfc9710e7619f9f92d714a7c9b
```
apply patch for usage of JaCoCo 0.7.9:
```diff
diff --git i/org.eclipse.eclemma.core/META-INF/MANIFEST.MF w/org.eclipse.eclemma.core/META-INF/MANIFEST.MF
index cc88add..57782d2 100644
--- i/org.eclipse.eclemma.core/META-INF/MANIFEST.MF
+++ w/org.eclipse.eclemma.core/META-INF/MANIFEST.MF
@@ -15,9 +15,9 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.debug.core,
  org.eclipse.jdt.core,
  org.eclipse.jdt.launching,
- org.jacoco.core;bundle-version="[0.7.7,0.7.8)",
- org.jacoco.agent;bundle-version="[0.7.7,0.7.8)",
- org.jacoco.report;bundle-version="[0.7.7,0.7.8)"
+ org.jacoco.core;bundle-version="[0.7.9,0.7.10)",
+ org.jacoco.agent;bundle-version="[0.7.9,0.7.10)",
+ org.jacoco.report;bundle-version="[0.7.9,0.7.10)"
 Bundle-Activator: org.eclipse.eclemma.internal.core.EclEmmaCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: J2SE-1.5
diff --git i/org.eclipse.eclemma.ui/META-INF/MANIFEST.MF w/org.eclipse.eclemma.ui/META-INF/MANIFEST.MF
index 62befeb..6b1a097 100644
--- i/org.eclipse.eclemma.ui/META-INF/MANIFEST.MF
+++ w/org.eclipse.eclemma.ui/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 3.0.0.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.eclemma.core,
- org.jacoco.core;bundle-version="[0.7.7,0.7.8)",
+ org.jacoco.core;bundle-version="[0.7.9,0.7.10)",
  org.eclipse.core.expressions,
  org.eclipse.core.runtime,
  org.eclipse.debug.ui,
```
and
```
mvn clean verify -Djacoco.version=0.7.9-SNAPSHOT
```
will use JaCoCo artifacts from local Maven repository.

Godin:
For the record: appearance of OSGi attributes in `jacocoant.jar` - is a regression that was introduced in version 0.6.1 by #56 , where we also changed it Maven coordinates.

marchof:
@Godin Still looking for a simple method to test EclEmma with local snapshot versions (also for future versions). If we could create a local p2 repo with JaCoCo snapshots we could simply add this repor to the EclEmma target definition. I tried to create a small tycho build for exactly this but did not suceed so far.

Godin:
@marchof while I agree that method described in https://github.com/jacoco/jacoco/pull/211#issuecomment-271180207 is not simple enough, I suggest to do use it now. And come back to the question about simplification of testing of EclEmma with snapshot version of JaCoCo separately, especially that difficulty is caused by changes on EclEmma side and that there are multiple possible solutions that we need to evaluate:
* on EclEmma side we can restore dependencies and `pomDependencies=true` that were removed in https://github.com/eclipse/eclemma/commit/330ec846686c63dfc9710e7619f9f92d714a7c9b , placing them into dedicated separate profile for JaCoCo testing
* also I like your idea about p2 repository

marchof:
@Godin Your first proposal would make testing quite simple:

* Adjust JaCoCo versions in MANFEST-MFs
* Run build with profile

Godin:
@marchof do you remember that during EclipseCon Europe 2016 while testing removal of compatibility layer for older Eclipse versions in EclEmma we were observing some unwanted effects of `pomDependencies=consider`? :stuck_out_tongue_winking_eye:

marchof:
@Godin Then, alternatively we add a profile to JaCoCo which creates a p2 repo.

Godin:
@marchof btw did you noticed that this PR was updated with fix for `jacocoant.jar` (tricky one - see comments inside of `pom.xml`) and update for changelog?

Godin:
@marchof you can easily create p2 repository containing snapshot version of JaCoCo using https://github.com/reficio/p2-maven-plugin Just execute `mvn p2:site` in a directory with following `pom.xml`
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001 XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">

  <groupId>org.example</groupId>
  <artifactId>example</artifactId>
  <packaging>pom</packaging>
  <version>0.1-SNAPSHOT</version>

  <properties>
    <jacoco.version>0.7.9-SNAPSHOT</jacoco.version>
  </properties>

  <build>
    <plugins>
      <plugin>
        <groupId>org.reficio</groupId>
        <artifactId>p2-maven-plugin</artifactId>
        <version>1.1.1</version>
        <configuration>
          <artifacts>
            <artifact><id>org.jacoco:org.jacoco.core:${jacoco.version}</id></artifact>
            <artifact><id>org.jacoco:org.jacoco.report:${jacoco.version}</id></artifact>
            <artifact><id>org.jacoco:org.jacoco.agent:${jacoco.version}</id></artifact>
          </artifacts>
        </configuration>
      </plugin>
    </plugins>
  </build>
</project>
```
result will be in `target/repository`.

marchof:
@Godin Yep, I noticed the change for `jacocoant.jar`! I still want to test whether the bundles actually work in an OSGi environment, i.e. building EclEmma with it. `p2-maven-plugin` works for me, thanks!

PR #477, opened 2017-01-08
Add missing assertions

Godin:
There are some unused line markers in validation tests. IMO either they should be removed as they are misleadingly look like important for test, either assertions should be added.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
@@ -53,6 +53,11 @@ public class ExceptionsTest extends ValidationTestBase {
 			/* partly when ECJ: */ICounter.NOT_COVERED);
 		}
 		assertLine("noExceptionTryCatch.catchBlock", ICounter.NOT_COVERED);
+		if (isJDKCompiler()) {
+			assertLine("noExceptionTryCatch.catchBlockEnd",
+			/* empty when ECJ: */ICounter.FULLY_COVERED);
+		}
+		assertLine("noExceptionTryCatch.afterBlock", ICounter.FULLY_COVERED);
 
 		// 4. Try/Catch Block With Exception Thrown Implicitly
 		assertLine("implicitExceptionTryCatch.beforeBlock",
@@ -65,6 +70,12 @@ public class ExceptionsTest extends ValidationTestBase {
 			/* partly when ECJ: */ICounter.FULLY_COVERED);
 		}
 		assertLine("implicitExceptionTryCatch.catchBlock",
+				ICounter.FULLY_COVERED);
+		if (isJDKCompiler()) {
+			assertLine("implicitExceptionTryCatch.catchBlockEnd",
+			/* empty when ECJ: */ICounter.NOT_COVERED);
+		}
+		assertLine("implicitExceptionTryCatch.afterBlock",
 				ICounter.FULLY_COVERED);
 
 		// 5. Try/Catch Block With Exception Thrown Implicitly After Condition
@@ -85,6 +96,9 @@ public class ExceptionsTest extends ValidationTestBase {
 		assertLine("explicitExceptionTryCatch.catch", ICounter.FULLY_COVERED);
 		assertLine("explicitExceptionTryCatch.catchBlock",
 				ICounter.FULLY_COVERED);
+		assertLine("explicitExceptionTryCatch.catchBlockEnd", ICounter.EMPTY);
+		assertLine("explicitExceptionTryCatch.afterBlock",
+				ICounter.FULLY_COVERED);
 
 		// 7. Finally Block Without Exception Thrown
 		// Finally block is yellow as the exception path is missing.
@@ -95,6 +109,11 @@ public class ExceptionsTest extends ValidationTestBase {
 			/* partly when ECJ: */ICounter.EMPTY);
 		}
 		assertLine("noExceptionFinally.finallyBlock", ICounter.PARTLY_COVERED);
+		if (isJDKCompiler()) {
+			assertLine("noExceptionFinally.finallyBlockEnd",
+			/* not covered when ECJ: */ICounter.FULLY_COVERED);
+		}
+		assertLine("noExceptionFinally.afterBlock", ICounter.FULLY_COVERED);
 
 		// 8. Finally Block With Implicit Exception
 		// Finally block is yellow as the non-exception path is missing.
@@ -109,6 +128,11 @@ public class ExceptionsTest extends ValidationTestBase {
 		}
 		assertLine("implicitExceptionFinally.finallyBlock",
 				ICounter.PARTLY_COVERED);
+		if (isJDKCompiler()) {
+			assertLine("implicitExceptionFinally.finallyBlockEnd",
+			/* fully when ECJ: */ICounter.NOT_COVERED);
+		}
+		assertLine("implicitExceptionFinally.afterBlock", ICounter.NOT_COVERED);
 
 		// 9. Finally Block With Exception Thrown Explicitly
 		assertLine("explicitExceptionFinally.beforeBlock",
@@ -121,6 +145,11 @@ public class ExceptionsTest extends ValidationTestBase {
 		}
 		assertLine("explicitExceptionFinally.finallyBlock",
 				ICounter.FULLY_COVERED);
+		if (isJDKCompiler()) {
+			assertLine("explicitExceptionFinally.finallyBlockEnd",
+			/* fully when ECJ: */ICounter.EMPTY);
+		}
+		assertLine("explicitExceptionFinally.afterBlock", ICounter.EMPTY);
 
 	}
 
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
@@ -48,6 +48,7 @@ public class ExceptionsTest extends ValidationTestBase {
 		// 3. Try/Catch Block Without Exception Thrown
 		assertLine("noExceptionTryCatch.beforeBlock", ICounter.FULLY_COVERED);
 		assertLine("noExceptionTryCatch.tryBlock", ICounter.FULLY_COVERED);
+		assertLine("noExceptionTryCatch.catch", ICounter.NOT_COVERED);
 		assertLine("noExceptionTryCatch.catchBlock", ICounter.NOT_COVERED);
 
 		// 4. Try/Catch Block With Exception Thrown Implicitly
@@ -56,6 +57,7 @@ public class ExceptionsTest extends ValidationTestBase {
 		assertLine("implicitExceptionTryCatch.before", ICounter.FULLY_COVERED);
 		assertLine("implicitExceptionTryCatch.exception", ICounter.NOT_COVERED);
 		assertLine("implicitExceptionTryCatch.after", ICounter.NOT_COVERED);
+		assertLine("implicitExceptionTryCatch.catch", ICounter.FULLY_COVERED);
 		assertLine("implicitExceptionTryCatch.catchBlock",
 				ICounter.FULLY_COVERED);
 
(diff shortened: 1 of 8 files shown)


org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -48,6 +48,7 @@ public void testCoverageResult() {
 		// 3. Try/Catch Block Without Exception Thrown
 		assertLine("noExceptionTryCatch.beforeBlock", ICounter.FULLY_COVERED);
 		assertLine("noExceptionTryCatch.tryBlock", ICounter.FULLY_COVERED);
+		assertLine("noExceptionTryCatch.catch", ICounter.NOT_COVERED);
```
marchof: This assertion fails with the Eclipse compiler (probably that's why I removed it at all):

    Status in line 75: 		} catch (StubException e) { // $line-noExceptionTryCatch.catch$ expected:<[NOT]_COVERED> but was:<[PARTLY]_COVERED>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -56,6 +57,7 @@ public void testCoverageResult() {
 		assertLine("implicitExceptionTryCatch.before", ICounter.FULLY_COVERED);
 		assertLine("implicitExceptionTryCatch.exception", ICounter.NOT_COVERED);
 		assertLine("implicitExceptionTryCatch.after", ICounter.NOT_COVERED);
+		assertLine("implicitExceptionTryCatch.catch", ICounter.FULLY_COVERED);
```
marchof: This assertions also fails with Eclipse:

    Status in line 86: 		} catch (StubException e) { // $line-implicitExceptionTryCatch.catch$ expected:<[FUL]LY_COVERED> but was:<[PART]LY_COVERED>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -74,13 +76,15 @@ public void testCoverageResult() {
...
 		assertLine("explicitExceptionTryCatch.throw", ICounter.FULLY_COVERED);
+		assertLine("explicitExceptionTryCatch.catch", ICounter.FULLY_COVERED);
 		assertLine("explicitExceptionTryCatch.catchBlock",
 				ICounter.FULLY_COVERED);
 
 		// 7. Finally Block Without Exception Thrown
 		// Finally block is yellow as the exception path is missing.
 		assertLine("noExceptionFinally.beforeBlock", ICounter.FULLY_COVERED);
 		assertLine("noExceptionFinally.tryBlock", ICounter.FULLY_COVERED);
+		assertLine("noExceptionFinally.finally", ICounter.EMPTY);
```
marchof: With Eclipse compiler:

    Status in line 116: 		} finally { // $line-noExceptionFinally.finally$ expected:<[EMPTY]> but was:<[PARTLY_COVERED]>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -90,6 +94,7 @@ public void testCoverageResult() {
 		assertLine("implicitExceptionFinally.before", ICounter.FULLY_COVERED);
 		assertLine("implicitExceptionFinally.exception", ICounter.NOT_COVERED);
 		assertLine("implicitExceptionFinally.after", ICounter.NOT_COVERED);
+		assertLine("implicitExceptionFinally.finally", ICounter.EMPTY);
```
marchof: With Eclipse compiler:

    Status in line 127: 		} finally { // $line-implicitExceptionFinally.finally$ expected:<[EMPTY]> but was:<[PARTLY_COVERED]>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -98,6 +103,7 @@ public void testCoverageResult() {
 				ICounter.FULLY_COVERED);
 		assertLine("explicitExceptionFinally.before", ICounter.FULLY_COVERED);
 		assertLine("explicitExceptionFinally.throw", ICounter.FULLY_COVERED);
+		assertLine("implicitExceptionFinally.finally", ICounter.EMPTY);
```
marchof: With Eclipse compiler:

    Status in line 127: 		} finally { // $line-implicitExceptionFinally.finally$ expected:<[EMPTY]> but was:<[PARTLY_COVERED]>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
Godin: Heck, I knew that should check with Eclipse :laughing:

Godin: Might just remove marker, however knowing that there is difference in bytecode generated by ECJ and without looking into it easy to expect "empty" or "not covered", but wondering why "partly"? :confused:

Godin: With `javac 1.8.0_112` after instrumentation:
```
      stack=4, locals=2, args_size=0
         0: invokestatic                   // Method $jacocoInit:()[Z
         3: astore_0
// line 8
         4: invokestatic                   // Method nop:()V
         7: aload_0
         8: iconst_2
         9: iconst_1
        10: bastore
// line 10
        11: invokestatic                   // Method nop:()V
// line 13
        14: aload_0
        15: iconst_3
        16: iconst_1
        17: bastore
        18: goto          33
// line 11
        21: astore_1
        22: aload_0
        23: iconst_4
        24: iconst_1
        25: bastore
// line 12
        26: invokestatic                   // Method nop:()V
        29: aload_0
        30: iconst_5
        31: iconst_1
        32: bastore
// line 14
        33: aload_0
        34: bipush        6
        36: iconst_1
        37: bastore
        38: return
      Exception table:
         from    to  target type
             7    14    21   Class java/lang/RuntimeException
      LocalVariableTable:
           26       7     1     e   Ljava/lang/RuntimeException;
```

With `org.eclipse.jdt.core.compiler:ecj:4.6.1` after instrumentation:
```
      stack=4, locals=1, args_size=0
         0: invokestatic                   // Method $jacocoInit:()[Z
         3: astore_0
// line 8
         4: invokestatic                   // Method nop:()V
         7: aload_0
         8: iconst_2
         9: iconst_1
        10: bastore
// line 10
        11: invokestatic                   // Method nop:()V
// line 11
        14: aload_0
        15: iconst_3
        16: iconst_1
        17: bastore
        18: goto          33
        21: pop
        22: aload_0
        23: iconst_4
        24: iconst_1
        25: bastore
// line 12
        26: invokestatic                   // Method nop:()V
        29: aload_0
        30: iconst_5
        31: iconst_1
        32: bastore
// line 14
        33: aload_0
        34: bipush        6
        36: iconst_1
        37: bastore
        38: return
      Exception table:
         from    to  target type
             7    14    21   Class java/lang/RuntimeException
```

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
Godin: With `javac 1.8.0_112` after instrumentation:
```
      stack=4, locals=2, args_size=0
         0: invokestatic                   // Method $jacocoInit:()[Z
         3: astore_0
// line 8
         4: invokestatic                   // Method nop:()V
         7: aload_0
         8: iconst_2
         9: iconst_1
        10: bastore
// line 10
        11: invokestatic                   // Method nop:()V
        14: aload_0
        15: iconst_3
        16: iconst_1
        17: bastore
// line 12
        18: invokestatic                   // Method nop:()V
// line 13
        21: goto          34
// line 12
        24: astore_1
        25: invokestatic                   // Method nop:()V
        28: aload_1
        29: aload_0
        30: iconst_4
        31: iconst_1
        32: bastore
        33: athrow
// line 14
        34: aload_0
        35: iconst_5
        36: iconst_1
        37: bastore
        38: return
      Exception table:
         from    to  target type
             7    18    24   any
```

With org.eclipse.jdt.core.compiler:ecj:4.6.1 after instrumentation:
```
    Code:
      stack=4, locals=2, args_size=0
         0: invokestatic                   // Method $jacocoInit:()[Z
         3: astore_0
// line 8
         4: invokestatic                   // Method nop:()V
         7: aload_0
         8: iconst_2
         9: iconst_1
        10: bastore
// line 10
        11: invokestatic                   // Method nop:()V
// line 11
        14: goto          31
        17: astore_1
        18: aload_0
        19: iconst_3
        20: iconst_1
        21: bastore
// line 12
        22: invokestatic                   // Method nop:()V
// line 13
        25: aload_1
        26: aload_0
        27: iconst_4
        28: iconst_1
        29: bastore
        30: athrow
// line 12
        31: invokestatic                   // Method nop:()V
// line 14
        34: aload_0
        35: iconst_5
        36: iconst_1
        37: bastore
        38: return
      Exception table:
         from    to  target type
             7    17    17   any
```

marchof:
@Godin I think it make more sense to look at the original class files (without instrumentation) and see how the compiler assignes lines to bytecode. Also not sure about the snippets you posted: We're talking about Target03.class, right? In this class relevant code starts after line 25, then I don't a relation to the line numbers in your snippets.

[Here](https://gist.github.com/marchof/3498f16f17e060e7ca7336afeef91c33) is a decompiled ECJ version of Target03.class. 

As you can see ECJ assigns the `goto` instruction (at the end of the protected block) as well as the `astore` instruction (to keep the exception instance) to the catch line. The `goto` instruction is only executed in the non-exception case, while `astore` is only executed in the exception case. That's why we see "partly covered".

Godin:
> I think it make more sense to look at the original class files (without instrumentation) and see how the compiler assignes lines to bytecode.

Instrumented classes allow to see positions of probes in addition to line numbers. At least in this case it really helped me to understand what's going on.

> Also not sure about the snippets you posted: We're talking about Target03.class, right?

Actually bytecode was decompiled from the below sources that are extracts from `Target03.java` :
```
package example;

class TryCatch {
  static void nop() {
  }

  static void fun() {
    nop();
    try {
      nop();
    } catch (RuntimeException e) {
      nop();
    }
  }
}
```

```
package example;

class TryFinally {
  static void nop() {
  }

  static void fun() {
    nop();
    try {
      nop();
    } finally (RuntimeException e) {
      nop();
    }
  }
}
```
Excuse me for creation of confusion.

> As you can see ECJ assigns the goto instruction (at the end of the protected block) as well as the astore instruction (to keep the exception instance) to the catch line. The goto instruction is only executed in the non-exception case, while astore is only executed in the exception case. That's why we see "partly covered".

Yup - our understanding matches, even if I didn't used `Target03` directly :wink: 

From here: Good to know about this difference, so comment anyway should be added.

Also we can actually keep assertions - enable them only in CI as a protection from regressions at least with/in javac, postponing question about absence of CI with ECJ. WDYT?

marchof:
@Godin How do you want to implement compiler specific checks? Maybe a method in the base class which is controlled with a system property?

    if (isJDKCompiler()) {
        assertLine("noExceptionTryCatch.catch", ICounter.NOT_COVERED);
    }

Godin:
@marchof yup - something controlled by a `java.lang.System.getProperty`

Godin:
Compilation that is triggered by Maven uses `javac` and there is already property that supposed to be defined when building using Maven, and thus is already defined in Travis - `bytecode.version`, so we could simply reuse it.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
Godin: And I made a typo here: `explicit` became `implicit`.

marchof:
@Godin Looks good for me now! Seeing your inline comments about ECJ we might also use a fluent API like this instead of if-statements:

    assertLine("explicitExceptionFinally.finally", //
        withJDK(ICounter.EMPTY), //
        withECJ(ICounter.EMPTY));

Godin:
> we might also use a fluent API

Let's keep this nice idea for future. We don't have an ability to execute assertions in case of ECJ, so I think that for now more natural to keep comments - executable code that cannot be executed requires explanation, otherwise misleading.

PR #480, opened 2017-01-13
MethodAnalyzerTest should not violate contract of ASM API

Godin:
For some experimental work I tried to implement addition of probe at beginning of method and had hard time to understand inconsistent failures in `MethodAnalyzerTest` - while some tests were failing as expected, others were passing.

Turned out that this was caused by the fact that `MethodAnalyzerTest` violates ASM API - according to [JavaDoc]( http://static.javadoc.io/org.ow2.asm/asm/5.2/org/objectweb/asm/MethodVisitor.html) of `MethodVisitor`:
> visitLineNumber methods must be called after the labels passed as arguments have been visited

As consequence test is less realistic, because simulates behavior that is different from the case when class is loaded from bytecode. Should be noted that such violation is not a big problem to generate bytecode. To demonstrate this asymmetry:
```java
package org.jacoco.core.internal.analysis;

import org.objectweb.asm.ClassReader;
import org.objectweb.asm.ClassWriter;
import org.objectweb.asm.Label;
import org.objectweb.asm.Opcodes;
import org.objectweb.asm.tree.ClassNode;
import org.objectweb.asm.tree.MethodNode;
import org.objectweb.asm.util.TraceClassVisitor;

import java.io.PrintWriter;
import java.util.ArrayList;

public class Example {

		final ClassNode classNode = new ClassNode();
		classNode.version = Opcodes.V1_6;
		classNode.access = Opcodes.ACC_PUBLIC;
		classNode.name = "Example";
		classNode.superName = "java/lang/Object";

		final MethodNode mn = new MethodNode();
		mn.name = "m";
		mn.desc = "()V";
		mn.exceptions = new ArrayList<String>();

		final Label label = new Label();
		// mn.visitLabel(label);
		mn.visitLineNumber(1, label);
		mn.visitInsn(Opcodes.RETURN);
		classNode.methods.add(mn);

		classNode.accept(newTraceClassVisitor());

		// label "L0" will not be presented in an output for above case,
		// but will be presented in an output for case below,
		// where class loaded from bytecode:
		final ClassWriter cw = new ClassWriter(
				ClassWriter.COMPUTE_MAXS | ClassWriter.COMPUTE_FRAMES);
		classNode.accept(cw);

		new ClassReader(cw.toByteArray()).accept(newTraceClassVisitor(), 0);
	}

	private static TraceClassVisitor newTraceClassVisitor() {
		return new TraceClassVisitor(new PrintWriter(System.out));
	}

}
```

IMO test should be fixed to avoid hard debugging time in future and to make test more realistic. As for fix: while we can remove direct usage of `MethodNode` by saving and loading bytes as shown above, I think will be better to simply use API properly.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
@@ -26,6 +26,7 @@ import org.objectweb.asm.Label;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.tree.MethodNode;
 import org.objectweb.asm.tree.TryCatchBlockNode;
+import org.objectweb.asm.util.CheckMethodAdapter;
 
 /**
  * Unit tests for {@link MethodAnalyzer}.
@@ -55,9 +56,13 @@ public class MethodAnalyzerTest implements IProbeIdGenerator {
 	// === Scenario: linear Sequence without branches ===
 
 	private void createLinearSequence() {
-		method.visitLineNumber(1001, new Label());
+		final Label l0 = new Label();
+		method.visitLabel(l0);
+		method.visitLineNumber(1001, l0);
 		method.visitInsn(Opcodes.NOP);
-		method.visitLineNumber(1002, new Label());
+		final Label l1 = new Label();
+		method.visitLabel(l1);
+		method.visitLineNumber(1002, l1);
 		method.visitInsn(Opcodes.RETURN);
 	}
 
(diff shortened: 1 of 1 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -43,7 +47,30 @@
 	@Before
 	public void setup() {
 		nextProbeId = 0;
-		method = new MethodNode();
+		method = new MethodNode(InstrSupport.ASM_API_VERSION) {
```
marchof: @Godin Have you considered using ASM's `CheckMethodAdapter` instead? When you emit the method node to a instance of this adapter exactly this false API usage should be detected.

Godin: @marchof Nice catch! Actually I completely forgot about existance of `CheckMethodAdapter`, so will reconsider/rework this - feel free to bloc merging by marking review as "requires changes" :wink:

PR #481, opened 2017-01-15
Add links about integration of JaCoCo into Codecov and Coveralls

Godin:


marchof:
@Godin Wanted to directly merge this but the Appveyor CI build seems to take its time...

Godin:
@marchof According to https://appveyor.statuspage.io/ - they had an incident that was resolved, so expect that build will be processed finally. Don't worry about merge as you've done most important part - review, and I will be able to merge :wink: And in any case - as an administrators we always have an ability to bypass checks and merge anyway.

PR #485, opened 2017-01-27
Consolidate .gitignore files

Godin:


marchof:
@Godin This does not yet work for Eclipse. The issue here is that we use linked resources which show up as outgoing changes.

1. `org.jacoco.agent/src/jacocoagent.jar`is a linked resource which should be excluded.
2. `org.eclipse.eclemma.core/.settings` and `org.eclipse.eclemma.core/.settings` are actual resources which should be continued to be versioned. In all other projects the settings folder is a link only and needs to be excluded.

But I still think this is doable in a single `.gitignore`file. I'll check their syntax.

marchof:
@Godin This should do the trick:

    target/
    .settings/
    !/org.jacoco.core/.settings/
    !/org.jacoco.core.test/.settings/
    /org.jacoco.agent/src/jacocoagent.jar

Godin:
@marchof fixed

PR #486, opened 2017-01-28
prepare-agent should not overwrite existing property into empty string when execution skipped

Godin:
https://groups.google.com/d/msg/jacoco/oWXxpvEz-dw/x1VWy3NPBgAJ


jacoco-maven-plugin/src/org/jacoco/maven/AbstractAgentMojo.java
```diff
@@ -149,8 +149,13 @@ public void executeMojo() {
...
 		final String name = getEffectivePropertyName();
 		final Properties projectProperties = getProject().getProperties();
-		getLog().info(name + " set to empty");
-		projectProperties.setProperty(name, "");
+		final String oldValue = projectProperties.getProperty(name);
+		if (oldValue == null) {
+			getLog().info(name + " set to empty");
+			projectProperties.setProperty(name, "");
+		} else {
+			getLog().info(name + " set to " + oldValue);
```
marchof: Actually we don't set anything here: Either we don't log at all or we log something like:  "Keep [property] as [oldValue]".

Godin: @marchof fixed

PR #488, opened 2017-02-14
Add TotalCoverage output to maven report

mauriceqch:
I have written a small formatter to output test coverage percentage to a text file.
In my case, it was useful for Gitlab CI because it avoided installing xml parsing tools on the server.

Here is the .gitlab-ci.yml I am using :

```
test:
  script: 
    - "mvn test"
    - "mvn jacoco:report"
    - "echo \"$(cat target/site/jacoco/testcoverage.txt)% covered\""
```

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
@@ -8,6 +8,7 @@
  * Contributors:
  *    Evgeny Mandrikov - initial API and implementation
  *    Kyle Lieber - implementation of CheckMojo
+ *    Maurice Quach - implementation of total coverage
  *
  *******************************************************************************/
 package org.jacoco.maven;
@@ -42,6 +43,7 @@ import org.jacoco.report.check.Rule;
 import org.jacoco.report.check.RulesChecker;
 import org.jacoco.report.csv.CSVFormatter;
 import org.jacoco.report.html.HTMLFormatter;
+import org.jacoco.report.totalcoverage.TotalCoverageFormatter;
 import org.jacoco.report.xml.XMLFormatter;
 
 /**
@@ -113,12 +115,21 @@ final class ReportSupport {
 				targetdir)));
 	}
 
+	public void addTotalCoverageFormatter(final File targetfile,
+			final String encoding) throws IOException {
+		final TotalCoverageFormatter totalCoverageFormatter = new TotalCoverageFormatter();
+		totalCoverageFormatter.setOutputEncoding(encoding);
+		formatters.add(totalCoverageFormatter
+				.createVisitor(new FileOutputStream(targetfile)));
+	}
+
 	public void addAllFormatters(final File targetdir, final String encoding,
 			final String footer, final Locale locale) throws IOException {
 		targetdir.mkdirs();
 		addXmlFormatter(new File(targetdir, "jacoco.xml"), encoding);
 		addCsvFormatter(new File(targetdir, "jacoco.csv"), encoding);
 		addHtmlFormatter(targetdir, encoding, footer, locale);
+		addTotalCoverageFormatter(new File(targetdir, "testcoverage.txt"), encoding);
 	}
 
 	public void addRulesChecker(final List<Rule> rules,
--- a/org.jacoco.report.test/src/org/jacoco/report/totalcoverage/TotalCoverageFormatterTest.java
+++ b/org.jacoco.report.test/src/org/jacoco/report/totalcoverage/TotalCoverageFormatterTest.java
@@ -1,0 +1,77 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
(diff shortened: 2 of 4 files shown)


Godin:
@mauriceqch first of all - thank you for your contribution!

However I have feeling that your use case is quite limited and too specific.

> avoided installing xml parsing tools

Given that you have Maven, then JDK is also here and it contains classes to work with XML 😉 
Also there is already CSV, which is super easy to parse.

@marchof WDYT?

marchof:
@Godin I agree. The same information is in the header of the xml report. "installing xml parsing tools" is definitely not required when you have a JDK at hand.

@mauriceqch Can you please elaborate more about the Gitlab CI integration and why a new format is required?

mauriceqch:
@Godin @marchof  Thank you for your answers !

Gitlab CI uses a runner to execute builds. Builds and commands executed are specified in a .gitlab-ci.yml file at the root of the project.
Gitlab then uses a regular expression to retrieve the test coverage.

The pull request provides a file which contains the test coverage percentage and can be printed on the standard output using cat and echo commands. It is then easy to set a regex to retrieve the test coverage percentage in Gitlab. Gitlab then provides a badge which can be displayed on the home page of the project.

[
pipelines_settings_test_coverage
](url)

pipelines_settings_badges

A solution I found on the net was to install a package like xmllint which then allows me to retrieve the text coverage in the HTML or XML using an xpath selector. However, this means that it is necessary to install an additionnal package.
I have also tried using regexes to retrieve the percentage but failed to succeed.

What would you suggest to use the JDK to achieve this goal ?
After thinking about it, i see these two solutions :
* Creating a maven plugin to provide a goal which parses jacoco.xml and outputs the test coverage percentage.
* Pulling a JDK library like saxon and use it from the command line

I agree that the same information is present in the XML and the HTML, however it seems more complicated to me.

Another solution, maybe less obstrusive, would be to print a line indicating coverage percentage, like simplecov does :

```
Dog
  barks

Finished in 0.00058 seconds (files took 0.08804 seconds to load)
1 example, 0 failures

Coverage report generated for RSpec to /tmp/coverage_example/coverage. 6 / 6 LOC (100.0%) covered.
```

Please tell me if I'm missing something !

Godin:
@mauriceqch what about http://www.mojohaus.org/xml-maven-plugin/transform-mojo.html ? :wink:

mauriceqch:
@Godin Thank you very much ! Didn't know about this maven plugin !

I managed to achieve the same result using the transform mojo.
Here is the configuration I used, I hope that it will help other people stumbling on the same problem.

pom.xml
```
<plugin>
	<groupId>org.codehaus.mojo</groupId>
	<artifactId>xml-maven-plugin</artifactId>
	<executions>
		<execution>
			<goals>
				<goal>transform</goal>
			</goals>
		</execution>
	</executions>
	<configuration>
		<catalogs>
			<catalog>catalog.xml</catalog>
		</catalogs>
		<transformationSets>
			<transformationSet>
				<dir>./</dir>
				<includes>target/site/jacoco/jacoco.xml</includes>
				<stylesheet>stylesheet.xsl</stylesheet>
			</transformationSet>
		</transformationSets>
	</configuration>
</plugin>
```

stylesheet.xsl
```
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
	<xsl:output method="xml" indent="yes" omit-xml-declaration="yes" />
	<xsl:template match="/">
		<xsl:value-of select="round((report/counter[@type='INSTRUCTION']/@covered div (report/counter[@type='INSTRUCTION']/@missed + report/counter[@type='INSTRUCTION']/@covered)) * 10000) div 100"/>
	</xsl:template>
</xsl:stylesheet>
```

catalog.xml
```
<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
	<public publicId="-//JACOCO//DTD Report 1.0//EN"
		uri="src/main/java/org/jacoco/report/xml/report.dtd"/>
</catalog>
```

.gitlab-ci.yml
```
test:
  script: 
    - "mvn test"
    - "mvn jacoco:report"
    - "mvn xml:transform"
    - "echo \"$(cat target/generated-resources/xml/xslt/target/site/jacoco/jacoco.xml)% covered\""
```

The sample output file is at target/generated-resources/xml/xslt/target/site/jacoco/jacoco.xml and contains the test coverage percentage. For example :
```
4
```

I had to cut and paste `report.dtd` into my project in order for the maven plugin to access it through the `catalog.xml`. Is there a better way ?

Godin:
@mauriceqch `report.dtd` is packaged into `org.jacoco.report.jar` , but I'm not sure that `xml-maven-plugin` supports loading of dtd from classpath. Seems that they recommend to use local copy - see http://www.mojohaus.org/xml-maven-plugin/examples/catalog.html , however an empty `report.dtd` just as a placeholder seems to work fine. In any case dtd is not that big, so IMO copy is not that big issue.

jripault:
Instead of using maven xml plugin, a simple awk do the job
`awk -F"," '{ instructions += $4 + $5; covered += $5 } END { print covered, "/", instructions, " instructions covered"; print 100*covered/instructions, "% covered" }' target/jacoco/jacoco.csv`

easy to use in the gitlab-ci file

PR #491, opened 2017-02-25
Exclude enum methods "values" and "valueOf" from reports

Godin:
I propose to do a baby step in story of filtering by excluding enum methods `values` and `valueOf` from reports.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
@@ -87,4 +87,48 @@ public class ClassAnalyzerTest {
 		assertEquals(1, coverage.getMethods().size());
 	}
 
+	@Test
+	public void testMethodFilter_EnumValues() {
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+		final MethodProbesVisitor mv = analyzer.visitMethod(0, "values",
+				"()[LFoo;", null, null);
+		assertNull(mv);
+		assertTrue(coverage.getMethods().isEmpty());
+	}
+
+	@Test
+	public void testMethodFilter_EnumNonValues() {
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+		final MethodProbesVisitor mv = analyzer.visitMethod(0, "values", "()V",
+				null, null);
+		mv.visitCode();
+		mv.visitInsn(Opcodes.RETURN);
+		mv.visitEnd();
+		assertEquals(1, coverage.getMethods().size());
+	}
+
+	@Test
+	public void testMethodFilter_EnumValueOf() {
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+		final MethodProbesVisitor mv = analyzer.visitMethod(0, "valueOf",
+				"(Ljava/lang/String;)LFoo;", null, null);
+		assertNull(mv);
+		assertTrue(coverage.getMethods().isEmpty());
+	}
+
+	@Test
+	public void testMethodFilter_EnumNonValueOf() {
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+		final MethodProbesVisitor mv = analyzer.visitMethod(0, "valueOf", "()V",
+				null, null);
+		mv.visitCode();
+		mv.visitInsn(Opcodes.RETURN);
+		mv.visitEnd();
+		assertEquals(1, coverage.getMethods().size());
+	}
+
 }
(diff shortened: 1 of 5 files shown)


org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -82,8 +84,18 @@ public void visitEnd() {
...
+	/**
+	 * @return <code>true</code> if method should not be analyzed
+	 */
 	// TODO: Use filter hook in future
-	private boolean isMethodFiltered(final int access, final String name) {
+	private boolean isMethodFiltered(final String className,
+			final String superClassName, final int access, final String name,
+			final String desc) {
+		if ("java/lang/Enum".equals(superClassName)
+				&& ("values".equals(name) || name.equals("valueOf") && desc
```
marchof: @Godin Shouldn't we also check the signature of the values method? For readability I would prefer braces or a separate statement for each method. Maybe also some inline comments what we're going to filter here.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
```diff
@@ -87,4 +87,24 @@ public void testMethodFilter_Lambda() {
 		assertEquals(1, coverage.getMethods().size());
 	}
 
+	@Test
+	public void testMethodFilter_EnumValues() {
```
marchof: @Godin What about a negative test for different signature?

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
```diff
@@ -87,4 +87,24 @@ public void testMethodFilter_Lambda() {
...
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+		final MethodProbesVisitor mv = analyzer.visitMethod(0, "values", "()V",
+				null, null);
+		assertNull(mv);
+		assertTrue(coverage.getMethods().isEmpty());
+	}
+
+	@Test
+	public void testMethodFilter_EnumValueOf() {
```
marchof: @Godin Same here: Negativ test for different signature.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: Good catch :thumbsup: Don't know why my mind decided that signature check not needed for `values` method, while needed for `valueOf` :laughing:

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
Godin: @marchof added

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
Godin: @marchof added

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof fixed

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -82,8 +83,24 @@ public void visitEnd() {
...
+	 * @return <code>true</code> if method should not be analyzed
+	 */
 	// TODO: Use filter hook in future
-	private boolean isMethodFiltered(final int access, final String name) {
+	private boolean isMethodFiltered(final String className,
+			final String superClassName, final int access, final String name,
+			final String desc) {
+		if ("java/lang/Enum".equals(superClassName)) {
+			// filter out methods that compiler creates for enums
+			if ("values".equals(name)
```
marchof: We mix two different comparisons styles here: "expected".equals(actual) vs. actual.equals("expected"). If would prefer to use the first one only as it is NPE safe.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
```diff
@@ -87,4 +87,38 @@ public void testMethodFilter_Lambda() {
 		assertEquals(1, coverage.getMethods().size());
 	}
 
+	@Test
+	public void testMethodFilter_EnumValues() {
+		analyzer.visit(Opcodes.V1_5, Opcodes.ACC_PUBLIC, "Foo", null,
+				"java/lang/Enum", null);
+
+		MethodProbesVisitor mv = analyzer.visitMethod(0, "values", "()[LFoo;",
```
marchof: Shouldn't that be two test methods? Took me a short moment to realize that this tests two setups.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: Umm, below there is anyway `name.startsWith(...)` that isn't NPE safe. Also according to my knowledge of ASM API - `desc` can't be `null`. For the NPE safety IMO far better to ensure that parameters of methods are not `null` by default and nullability should be stated explicitly.

But ok, I'll swap arguments, hoping that  you won't say that concatenation on left hand side looks awkward 😆

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
Godin: I hesitated to create separate, because then IMO not so easy to catch difference between them. But ok, I'll split.

Issue #15, opened 2012-08-29
Filtering options for coverage analysis


marchof:
JaCoCo should support configurable filtering options for code coverage analysis. This page lists of use cases and requirements: https://github.com/jacoco/jacoco/wiki/FilteringOptions

sham-marol:
Hi,

Do we have the filtering option, where I can exclude part of the code in my class example - a method during code coverage analysis. Does the latest release of jacoco has this feature?

henri-tremblay:
I understand the goal to have a clean API. But indeed, at some point, something should happen. I had the same issue with my personal projects and have waited too much time.

I used to use Clover, but it was not very stable from one version to another and isn't open source.

Cobertura was too intrusive so I skipped to JaCoCo which I'm happy with.... apart from the filtering. And the only really important filtering seems to be able to tag the code.

I understand that people can then trick JaCoco by using too much tags. But every new tool can be used for evil. Anyway, that's not my use-case. I'm having 100% coverage on EasyMock. So, I want to know when I'm under that mark and easily find the hole. Without filtering, I can't. I need to dig all around the place to find it.

///COVERAGE:OFF from mchr3k is exactly what I would need.

How can I help?  I'm willing to provide a pull request that would be accepted. What would that be?

marchof:
@henri-tremblay I agree with you that the API is not the highest priority. Highest priority currently are removing annoying compiler artifacts especially for new Java 7 language constructs. As JaCoCo is fully based on class files source file processing (e.g. for comment parsing) is not in scope.

henri-tremblay:
But if the filtering is done by annotations, it won't be really precise. Java 8 will be a bit better since you can put annotations all over the place.

I haven't looked at mchr3k code yet but I would guess the following algorithm:
# Record all coverage
# Grep in the code for tags
# Remove from coverage the code between on an of

It doesn't really require a formal parsing, only a grep. It will be slower then when not activated but if not activated by default it won't matter

WDYT?

TomRK1089:
@henri-tremblay My understanding is that there is nothing to grep. JaCoCo analyzes compiled bytecode, not source files. Thus, there will be no magical ///COVERAGE:OFF to find.

henri-tremblay:
Hum... I understand. It should still be possible from a Maven task. Another solution would be to have a filter file a bit like FindBugs would do. Since the debug info will provide the line numbers, that should be enough.

addedsparkle:
[...]

PR #495, opened 2017-03-10
Exclude from a report methods annotated by "lombok/Generated"

t1:


t1:
pull request for #15

Godin:
@t1 first of all - thank you for your contribution! :thumbsup:

Referring to https://github.com/jacoco/jacoco/issues/15#issuecomment-285793246: "works out of the box" - is a good approach in general. However: I'm not sure that enough to exclude only annotated methods, in particular I'm thinking about https://projectlombok.org/features/Cleanup.html and maybe there are others, but can't tell for sure due to lack of tests and experience with Lombok on our side. And the message that will be delivered to community by addition of built-in exclusion for Lombok annotation might lead to a question why not all code that Lombok generates is excluded. By saying this, I'm not against such change, just want to make sure that will not be introduced something what will be hard to maintain (including answers on questions in mailing list and so on) in a long run. @marchof WDYT?

P.S. commit message requires change, because #15 is a bigger story - for example also about Java 7 `try-with-resources` and not just about Lombok, but that's minor in lights of question above

t1:
@Godin: You are completely right: This fixes only a subset of the code that Lombok can generate (and of course doesn't fix most of what has to be done for #15). And I must admit that I hadn't thought of this before.

Let's take a look at all [lombok features](https://projectlombok.org/features/index.html):

| Annotation  | Comment |
| ------------- | ------------- |
| [`val`](https://projectlombok.org/features/val.html) | - |
| [`@NonNull`](https://projectlombok.org/features/NonNull.html) | **Adds branch** |
| [`@Cleanup`](https://projectlombok.org/features/Cleanup.html) | **Adds branches** |
| [`@Getter` `@Setter`](https://projectlombok.org/features/GetterSetter.html) | Adds method (see lazy below) |
| [`@ToString`](https://projectlombok.org/features/ToString.html) | Adds method |
| [`@EqualsAndHashCode`]() | Adds methods |
| [`@NoArgsConstructor` `@RequiredArgsConstructor` `@AllArgsConstructor`](https://projectlombok.org/features/Constructor.html) | Adds method (constructor) |
| [`@Data`](https://projectlombok.org/features/Data.html) | Shortcut for other annotations; adds methods |
| [`@Value`](https://projectlombok.org/features/Value.html) | Shortcut for other annotations; adds methods |
| [`@Builder`](https://projectlombok.org/features/Builder.html) | Adds methods |
| [`@SneakyThrows`](https://projectlombok.org/features/SneakyThrows.html) | **Adds branch** |
| [`@Synchronized`](https://projectlombok.org/features/Synchronized.html) | Adds lines, but no branch |
| [`@Getter(lazy=true)`](https://projectlombok.org/features/GetterLazy.html) | **Adds branches** |
| [`@Log` (et.al.)](https://projectlombok.org/features/Log.html) | Adds field with initializer |
| [`var`](https://projectlombok.org/features/experimental/var.html) \[experimental] | - |
| [`@Accessors`](https://projectlombok.org/features/experimental/Accessors.html) \[experimental] | - |
| [`@ExtensionMethod`](https://projectlombok.org/features/experimental/ExtensionMethod.html) \[experimental] | - |
| [`@FieldDefaults`](https://projectlombok.org/features/experimental/FieldDefaults.html) \[experimental] | - |
| [`@Delegate`](https://projectlombok.org/features/experimental/Delegate.html) \[experimental] | Adds methods |
| [`@Wither`](https://projectlombok.org/features/experimental/Wither.html) \[experimental] | Adds methods |
| [`@onX`](https://projectlombok.org/features/experimental/onX.html) \[experimental] | - |
| [`@UtilityClass`](https://projectlombok.org/features/experimental/UtilityClass.html) \[experimental] | Adds method (private constructor) |
| [`@Helper`](https://projectlombok.org/features/experimental/Helper.html) \[experimental] | - |

`-` means that this annotation doesn't generate any code that affects branch or line coverage.

Features that add methods get the `@lombok.Generated` annotation and are skipped by this pull request. Features that add lines without branches are covered, so this might change the line statistics. I assume that this not really an issue.

Leaving us with four Lombok features that add branches: We heavily use Lombok in our team, but we've never used `@Cleanup` (I guess it became obsolete with ARM blocks available). We regularly use `@NonNull` and `@SneakyThrows`, and rarely `@Getter(lazy=true)`. So this is a realistic use case.

When I look at our tests for code with `@NonNull` and `@Getter(lazy=true)`, we try to cover all branches here as well! So we explicitly *want* that code to be covered. I'm not so sure about `@Cleanup` and `@SneakyThrows`, it may be that we don't want to cover that code, as we assume that code generated by Lombok is sufficiently safe, but I guess that's a question of target coverage check limits: If we want 100% coverage, we'd want to cover those, too.

OTOH do you have any idea of how this *could* be fixed? We could try to de-lombok the code ourselves, but changes in Lombok would require an update to JaCoCo. Even finding out which version of Lombok actually produced that code may not be very easy... as well as figuring out, if this code actually *did* come from Lombok and is not in the source code. I'd say this is not feasible.

What do you think? Is it sufficient to cleanly document that only the methods and constructors annotated as `@lombok.Generated` are skipped?

Godin:
@bjkail Hey, glad to see you here! Actually was about to ask if you're interested in this subject and if so maybe you could leave some valuable comments as before 😉  And here they are 😋 

You are right on all points - let's cleanup and harden this PoC. Also wondering WDTY about approach in general? About it we'd better add explanatory comments also. As well as estimate impact on performance of analysis - can imagine some optimizations... but traded them on the correctness and attempt to keep relative simplicity of code for now.

Anyway thank you! we really appreciate your help 👍

Godin:
@t1 first of all - thank you very much for such detailed comment! :thumbsup:
And please excuse us for delays with replies.

> This fixes only a subset of the code that Lombok can generate (and of course doesn't fix most of what has to be done for #15). And I must admit that I hadn't thought of this before.

So reworded commit message to not cause automatic close of #15. Also rebased on top of master.

> OTOH do you have any idea of how this could be fixed? We could try to de-lombok the code ourselves, but changes in Lombok would require an update to JaCoCo. Even finding out which version of Lombok actually produced that code may not be very easy... as well as figuring out, if this code actually did come from Lombok and is not in the source code. I'd say this is not feasible.

Filtering on a level lower than classes/methods, i.e. on a level of bytecode instructions is difficult to implement even in case of bytecode that only Java compiler generates. Each "third" comment of #15 among variety of "plus ones" and "whens" tries to explain exactly this to the users - see for example https://github.com/jacoco/jacoco/issues/15#issuecomment-146559448 or https://github.com/jacoco/jacoco/issues/15#issuecomment-241092025

Attempts to also put on a table instructions (and not just methods) generated by Lombok or JVM-based languages other than Java (Scala, Kotlin, etc) only complicate this story. And exactly from here question arises - whether it is sufficient to exclude only methods for Lombok.

> What do you think? Is it sufficient to cleanly document that only the methods and constructors annotated as @lombok.Generated are skipped?

Based on your description, I'm fine to give it a try.

But anyway would like to hear opinion of @marchof . So @marchof WDYT?

marchof:
@Godin I'm also not a lombok user but from @t1 's description this seems to come with great benefits for typical lombok usage. Therefore I would also vote for adopting this.

Implementation wise we should use the new filtering API which just became available (eaef191) on master ;-)

Godin:
Some additional thoughts from a non-Lombok-user:

Assuming that getters and setters are used by other code, they supposed to be covered indirectly, and if they are not used, then why do they exist? :stuck_out_tongue_winking_eye: so can missing coverage be an indicator of unused code? if so, then maybe the fact that filter can't be disabled might be challenged. Was always wondering about this, but never had an access to experienced Lombok user, @t1 :wink: 

Also in contrast for example to `try-with-resources` statement that introduces completely unreachable and hence untestable code, they don't introduce such.

Also filtering of getters and setters generated by Lombok might cause a logical question - why handwritten getters and setters are not filtered? :wink: Aren't we vendor-neutral, @marchof ? :laughing: 

P.S. in the lights of restrictions that modularity in Java 9 will bring, the fact that Lombok hacks into internals of compiler to do its job looks quite interesting...

marchof:
@Godin When code is handwritten unit tests are required. If code is generated there is no need for unit tests for the generated code (given that someone else tested the generator for us).

Godin:
> Implementation wise we should use the new filtering API which just became available

@marchof maybe, however
* as of today the main purpose of a new internal API - is to filter some instructions of a method, but not all of them
* needs to be measured, but usage of this API in comparison to the current way of filtering of synthetic methods might be slower
* we haven't migrated already existing filtering of methods
* might be a way simpler/faster from an implementation point of view to use old way
* and focus more on current work-in-progress for instruction-level filters ;)

scooper4711:
As a Lombok user, the methods I'm eager to filter from Jacoco (and sonarqube) reports are equals() and hashcode(). They are complex and difficult to get full branch coverage. I trust the generated code and don't need to get int every nook and cranny.

marchof:
@Godin The new API already works for this purpose! You have all information at hand in MethodNode. I agree that there is a faster way than filtering all instructions from the generated method, but this performance issue only applies to actually filtered methods.

I would prefer to integrate this feature with the new API and also move existing filters soon. I can prepare PRs for this.

Godin:
> performance issue only applies to actually filtered methods

Fair enough.

> also move existing filters soon. I can prepare PRs for this.

@marchof I also can do this, while definitely can't review my own changes :stuck_out_tongue_closed_eyes: what about deal to do a review of filter for `try-with-resources` before move of existing filters? :wink: 

> I would prefer to integrate this feature with the new API

Given a fair statement about performance and upcoming refactoring of existing filters, now this looks reasonable and convincing. @t1 will you have time and wish to work on this?

BeneStem:
This comment is "just" for motivation and encouragement purposes. You deserve some credit!

@t1 thank you so much for that amazing pull-request. My dev-team and I at [Otto (GmbH & Co. KG)](https://www.otto.de/) would love to have this feature. We are lombok-powerusers.
Excluding everything annotated by lombok/Generated is exactly what we need!

marchof:
@Godin To get this contribution in I can adjust it to the new API and do the paperwork.

marchof:
@t1 Is there a reason for closing this PR? Any objections that we include your contributions based of the new filtering API?

t1:
Sorry for not replying earlier. I was extremely busy last week.
I've started to work on this again this morning, but I'm not git-savvy enough to understand how I can change the `fix #15` comment in an existing pull request. And it wasn't much work, yet, and redoing it based on the new API seemed easier to me, so I wanted to start from scratch.
@marchof: Do you want to take over? I wouldn't object ;-)

marchof:
@t1 Done #513. Please feel free to comment on this. The new implementation assumes the annotation is runtime invisible as Lombok JavaDoc says retention is CLASS. Can you confirm?

BeneStem:
@marchof i think you are corrent. the annotation is runtime invisible. there is an option to write findbugsannotations next to the generated annotation. the findbugsannotation is runtime visible though :)

Godin:
@t1 @BeneStem guys, do you have a hello-world Maven project demonstrating how to get methods annotated with `lombok.Generated` ? I'd like to do end-to-end test before merging #513

BeneStem:
@Godin If you give me a few minutes I could build a demo with gradle.

Godin:
@BeneStem take your time, example with Gradle will be useful! But anyway I would really love to see how to make it work with Maven too 😉

BeneStem:
@Godin I'll do it with maven than.

BeneStem:
@Godin Ok. You can use this one: https://github.com/BeneStem/iaa-lecture. If you have any questions feel free to ask :)

Godin:
@BeneStem sorry, but first of all I wouldn't call it "minimalistic example" 😉  secondly and most importantly: I don't see annotation `lombok.Generated` in generated classes and this is exactly why I asked for the example - addition of Lombok as dependency is clearly not enough.

BeneStem:
@Godin Lombok should add the annotations without setting any properties. Look at the bottom of this page: https://projectlombok.org/features/configuration.html. There is stated that you have to disable it manually if you don't want it.

Godin:
@BeneStem here is my point:
if I compile
```java
import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;

@MyAnnotation
class Fun {
}

@Target({ElementType.CONSTRUCTOR, ElementType.METHOD, ElementType.FIELD, ElementType.TYPE})
@Retention(RetentionPolicy.CLASS)
@interface MyAnnotation {
}
```
then I can clearly see `MyAnnotation` in class file:
```
javap -v -p Fun | grep MyAnnotation
  #11 = Utf8               LMyAnnotation;
```
but I don't see `Generated` in your classes:
```
javap -v -p -cp target/classes de.nordakademie.iaa.model.Room | grep Generated
```

Godin:
Moreover
```
java -jar ~/.m2/repository/org/projectlombok/lombok/1.16.16/lombok-1.16.16.jar config -g --verbose | grep "lombok.Generated"
## Generate @lombok.Generated on all generated code (default: false).
```
as well as https://github.com/rzwitserloot/lombok/commit/70f77f928ac7a6edf3ef91a41399c9a79561c871 state that generation of this annotation is off by default.

BeneStem:
@Godin You are correct. I added another dependency and activated FindbugsAnnotation to see if this one is set.
I was able to verify that this is setted.

I will try activating the generated annotation now...

BeneStem:
@Godin I tried activating "lombok.extern.findbugs.addGeneratedAnnotation=true" the annotation. But was not able to find it in the class file... But the FB annotation is there.
Are you able to see the annotation after pulling?

Godin:
@BeneStem I managed to configure my own example - was using wrong name of configuration property and the correct one is `lombok.addLombokGeneratedAnnotation`

BeneStem:
@Godin Ahh, thank you. Now it is working in my example as well. Thank you and sorry for the confusion!

BeneStem:
@Godin I stated an issue at lomboks repostiory to document the property.

PR #498, opened 2017-03-16
Include runtime dependencies in aggregated reports

okutane:
It's necessary to also check coverage of runtime dependencies.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
@@ -105,7 +105,7 @@ public class ReportAggregateMojo extends AbstractReportMojo {
 				dataFileExcludes);
 		loadExecutionData(support, filter, getProject().getBasedir());
 		for (final MavenProject dependency : findDependencies(
-				Artifact.SCOPE_COMPILE, Artifact.SCOPE_TEST)) {
+				Artifact.SCOPE_COMPILE, Artifact.SCOPE_RUNTIME, Artifact.SCOPE_TEST)) {
 			loadExecutionData(support, filter, dependency.getBasedir());
 		}
 	}
@@ -128,7 +128,7 @@ public class ReportAggregateMojo extends AbstractReportMojo {
 	void createReport(final IReportGroupVisitor visitor,
 			final ReportSupport support) throws IOException {
 		final IReportGroupVisitor group = visitor.visitGroup(title);
-		for (final MavenProject dependency : findDependencies(Artifact.SCOPE_COMPILE)) {
+		for (final MavenProject dependency : findDependencies(Artifact.SCOPE_COMPILE, Artifact.SCOPE_RUNTIME)) {
 			support.processProject(group, dependency.getArtifactId(),
 					dependency, getIncludes(), getExcludes(), sourceEncoding);
 		}


marchof:
@okutane Can you please describe your use case a bit and how your modules are structured? So far we assumed that there will be a separate module for the aggregate report which explicitely declares dependencies to include into the report.

okutane:
Hello! I've used project structure similar to following.

### Initial structure 

root pom:

```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
...
    <modules>
        <module>api</module>
        <module>core</module>
        <module>api-impl1</module>
        <module>api-impl2</module>
    </modules>

    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <configuration>
                    ...
                    <includes>
                        <include>**/AllTests.class</include>
                    </includes>
                    <argLine>${argLine}</argLine>
                </configuration>
            </plugin>
            <plugin>
                <groupId>org.jacoco</groupId>
                <version>0.7.9</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>prepare-agent</goal>
                        </goals>
                    </execution>
                    <execution>
                        <id>report</id>
                        <phase>test</phase>
                        <goals>
                            <goal>report</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
...
</project>

```

api just declares some interfaces, api-implN implements them in a different ways and core binds everything together (now only by integration tests).

core pom:

```
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

    <parent>...</parent>

    ...
    <dependencies>
        <dependency>
            <groupId>...</groupId>
            <artifactId>api</artifactId>
            <version>...</version>
        </dependency>
    </dependencies>

    <profiles>
        <profile>
            <id>impl1</id>
            <dependencies>
                <dependency>
                    <groupId>...</groupId>
                    <artifactId>api-impl1</artifactId>
                    <version>...</version>
                    <scope>runtime</scope>
                </dependency>
            </dependencies>
        </profile>
    </profiles>
</project>
```

### How I used it
`mvn clean test -P impl1`

Since the tests are only in **core** module I've got everything built, executed and tested, but the coverage report was only for the **core** module itself.

### How I fixed it

Then I switched to report-aggregate and had coverage only for **api**. data for **api-impl1** has been ignored because the scope wasn't either of 'compile' or 'test'. as for the **core** itself, I guess the data has been ignored because aggregate reports don't include data for the current module. Anyway, this all was fixed by introduction of **tests** module which included everything needed as compile dependencies. https://github.com/okutane/sanity/commit/cbcdd0d36964705d30ee7cc06bfd41bd13d26c33

### Pull request

It's OK to me to keep that structure, but I would like to switch api-impl scopes to runtime and still have the coverage in integration tests.

okutane:
@marchof, I guess I can add a few parameters for report-aggregate goal if you don't want the default behavior to be changed.

marchof:
@okutane Thanks for the detailed explanation! In your scenario the "core" module is a good place to also create the coverage report. Also I think it is reasonable to include runtime dependencies for the scope. For the exec file collection I'm not sure though but at least this will not hurt.

 @Godin WDYT? Update documentation and include this PR?

marchof:
@okutane Finalized paperwork for this in PR #502.

PR #500, opened 2017-03-16
Exclude from a report a part of bytecode that compiler generates for a try-with-resources

Godin:
https://github.com/jacoco/jacoco/wiki/filtering-JAVAC.TRYWITH


marchof:
@Godin Count me deeply impressed!!! And I'm very sorry that I couldn't immediately jump into both PRs. So+Mo are travel dates, so I hope to study it in detail then.

Godin:
@marchof no worry, will be cool to get all this on board, but no need to rush - let's do a proper careful review and testing. Slowly, but at least we are progressing on this big subject. And I personally start to see the light at the end of the tunnel, which wasn't the case few months ago.

FYI next week I'm also kind of traveling - will attend DevoxxUS, and hope to not be offline all the time.

bjkail:
@Godin It would be useful if the members of `Pattern` had high-level descriptions of the code being matched and which conditions cause the compiler to generate that pattern.  It would be good to have a test of t-w-r with custom finally block (`try (...) { ... } finally { nop() }`) to be sure it doesn't cause problems for the analyses.  It would also be good to have tests with nested t-w-r nested in the try, catch, and finally blocks (e.g., `try (...) { ... } finally { try (...) { ... } catch (...) { ... } finally { ... } }`, etc.).

Godin:
@bjkail Hey, glad to see you here! Actually was about to ask if you're interested in this subject and if so maybe you could leave some valuable comments as before 😉  And here they are 😋 

You are right on all points - let's cleanup and harden this PoC. Also wondering WDTY about approach in general? About it we'd better add explanatory comments also. As well as estimate impact on performance of analysis - can imagine some optimizations... but traded them on the correctness and attempt to keep relative simplicity of code for now.

Anyway thank you! we really appreciate your help 👍

bjkail:
@Godin The overall approach seems reasonable to me of skipping ignored instructions in `MethodAnalyzer.visitEnd` and matching bytecode sequences in `TryWithResourcesMatcher`.

Last time I thought about t-w-r, I was thinking to require the t-w-r close sequence be replicated at the end of the try block and each catch block (including Throwable) to ensure it was either written by a compiler or fully correctly hand-written.  My motivation was to avoid filtering out a badly hand-written close sequence (e.g., at the end of try block only but not all catch blocks or missing Throwable block, etc.), but the algorithm to do the rigid matching seemed quite complex, especially for custom finally blocks and nested t-w-r (thus my suggested testcases).  However, I am now thinking your approach is fine.  There is some risk of false positive (matching a hand-written close sequence), but the risk is low since the close sequence is relatively complex, and we can leave detecting _badly_ hand-written close sequences (like missing close sequence on some possible code path) to static analyzers like FindBugs.

bjkail:
As an anecdote, last time I investigated t-w-r was because we kept regressing coverage in classes that use t-w-r heavily because drops go unnoticed due to all the uncovered generated code.  I tried forcing all the generate code paths to be executed (our code/tests made it easy to force all the conditions: null resource, exception with null resource, exception with non-null resource, close exception, exception and close exception), but I ultimately gave up when I discovered javac emits completely dead code:
* `if (#primaryExc) != null)` is always false in the try block and is always false in the catch blocks.
* `if (Identifier != null)` can be always false in the try block.  [JDK-7020499](https://bugs.openjdk.java.net/browse/JDK-7020499) improves that for Java 9, but I wish it also had "effectively non-null" analysis like the "effectively final" analysis that would elide the check.

So, I think this will be a great feature to have in JaCoCo :-).

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/TryWithResourcesTest.java
```diff
@@ -0,0 +1,148 @@
...
+	public void two() {
+		assertLine("two.before", ICounter.FULLY_COVERED);
+		// without filter next line covered partly:
+		assertLine("two.try", ICounter.FULLY_COVERED);
+		assertLine("two.open1", ICounter.FULLY_COVERED);
+		assertLine("two.open2", ICounter.FULLY_COVERED);
+		assertLine("two.body", ICounter.FULLY_COVERED);
+		// without filter next line has branches:
+		assertLine("two.close", ICounter.EMPTY);
+		assertLine("three.after", ICounter.FULLY_COVERED);
```
Godin: typo - `"three.after"` instead of `"two.after"`

Godin:
@marchof I made functional fixes here compared to initial version, as well as introduced packages/interfaces as was discussed in https://github.com/jacoco/jacoco/pull/496#issuecomment-288576629 similarly to what was done in https://github.com/jacoco/jacoco/pull/501. As well as made changes to improve performance of this filter - analysis of 17391 classes from rt.jar (≈ 65 Mb) from JDK 1.8.0_121 using following build.xml
```
<project xmlns:jacoco="antlib:org.jacoco.ant" default="report">
    <classpath path="jacocoant.jar"/>
  </taskdef>
  <target name="report">
    <jacoco:report>
      <executiondata>
        <file file="jacoco.exec"/>
      </executiondata>
      <structure name="rt.jar">
        <classfiles>
          <fileset file="rt.jar"/>
        </classfiles>
      </structure>
    </jacoco:report>
  </target>
</project>
```
takes 5 seconds on my laptop, both without this change as well as with it, i.e. overhead is not clearly observable.

So all in all - from a functional point of view IMO all done here and will be really great to receive your comments to merge this into master.

marchof:
@Godin How do you think we should proceed with this series of PRs? My issue is that they all rely on the same new APIs and strategies where I have several comments on like in #501. I propose to clean up #501 first and afterwards base the other PRs on this work.

Alternatively we create a "filtering" branch and integrate all PRs one after the other onto it.

Godin:
@marchof deal - let's clean up #501 first

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
```diff
@@ -0,0 +1,849 @@
...
+import org.objectweb.asm.util.Textifier;
+import org.objectweb.asm.util.TraceClassVisitor;
+import org.objectweb.asm.util.TraceMethodVisitor;
+
+public class TryWithResourcesFilterTest implements IFilterOutput {
+
+	private final MethodNode m = new MethodNode(InstrSupport.ASM_API_VERSION, 0,
+			"name", "()V", null, null);
+
+	private static void asmify(String filename) throws IOException {
```
marchof: Method is unused.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
```diff
@@ -0,0 +1,849 @@
...
+		assertEquals(2, from.size());
+
+		assertEquals(m.instructions.get(5), from.get(0));
+		assertEquals(m.instructions.get(18), to.get(0));
+
+		assertEquals(m.instructions.get(22), from.get(1));
+		assertEquals(m.instructions.get(88), to.get(1));
+	}
+
+	private void print() {
```
marchof: Unit tests shouldn't print to System.out.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
```diff
@@ -0,0 +1,849 @@
...
+import org.objectweb.asm.Label;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+import org.objectweb.asm.util.ASMifier;
+import org.objectweb.asm.util.Textifier;
+import org.objectweb.asm.util.TraceClassVisitor;
+import org.objectweb.asm.util.TraceMethodVisitor;
+
+public class TryWithResourcesFilterTest implements IFilterOutput {
```
marchof: Lot of magic numbers here. Can't we use labels as in #501?

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
```diff
@@ -108,6 +108,7 @@ public void visitEnd() {
 	public void analyzeClass(final ClassReader reader) {
 		final ClassVisitor visitor = createAnalyzingVisitor(
 				CRC64.checksum(reader.b), reader.getClassName());
+		// TODO maybe we can SKIP_FRAMES here?
```
marchof: Didn't we remove this in #501?

Godin:
> At a first glance it looks like we need some better byte code pattern matching abstraction to create maintainable code.

@marchof believe me - current version is a way way better than initial monsters ;)

> Try to wrap my mind about the actual filter implementation now.

Overall principle is that matching performed sequentially (`&&`, `if (!condition) exit`) with a bit of backtracking. And one of key points - is assignment of names, which helped to simplify initial version a lot. Also could be noted and seen on example of two commits here that code/overall algorithm based on those principles is quite recombinable/changeable. Most of the time and headache were spent on generation of test cases and understanding of bytecode and differences in it for different cases.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: Yes, but with explicit statement that it will reappear here :wink: - see https://github.com/jacoco/jacoco/pull/501#discussion_r108471442

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
Godin: not by means of labels, but got rid of magic numbers

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
Godin: was handy in presence of magic numbers :laughing: but not anymore

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
Godin: was handy for creation of tests, removed

marchof:
@Godin Where can I find the patterns that you where matching for? in the test case? Maybe you can dump a informal description in our wiki: https://github.com/jacoco/jacoco/wiki/filtering-JAVAC.TRYWITH

I think I get the idea how the control flow of matching is implemented. My point is that we mix two complex concepts into one piece of code:

1. Description of different byte code sequences for different compilers
2. Pattern matching implementation

To create maintainable code I would prefer to separate some concepts here:

1. Separate filters for different compilers
2. Pure pattern declaration in the filter
3. Generic matcher facility

This should greatly reduce the complexity of the filters. I already started working on this on my early prototype. If you give me a couple of days I can come up with a proposal. Would be helpful though to understand what the specific patterns are you're matching for here (see above).

Godin:
@marchof 

> Where can I find the patterns that you where matching for? in the test case?

Yes - I believe that unit test is exhaustive in terms of patterns.

> Maybe you can dump a informal description in our wiki: https://github.com/jacoco/jacoco/wiki/filtering-JAVAC.TRYWITH

I consider wiki as a temporary storage (recall the https://github.com/jacoco/jacoco/pull/496#issuecomment-288251806) and expect documentation to become part of code (ref to [today's tweet from our friend Mickhael Istria](https://twitter.com/mickaelistria/status/846985263230439424) :wink:) and so indeed - more explanations should be added into this change.

I'll draft some in wiki.

> I think I get the idea how the control flow of matching is implemented.

:thumbsup:

> My point is that we mix two complex concepts into one piece of code:
> 1. Description of different byte code sequences for different compilers
> 2. Pattern matching implementation

That's the purpose of comments in code, isn't it (https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/) ? :wink:

More seriously - implementation workflow looked like:
1. generate bytecode for a test (given that in different cases bytecode can be different and that cases are unknown in advance, was not that easy, but mostly matter of time)
2. understand bytecode by adding comments into test
3. looking at comments understand patterns
4. after that implementation is almost one-to-one maps to comments in tests

But indeed some description/overview of patterns is missing here (see above).

> To create maintainable code I would prefer to separate some concepts here:
> 1. Separate filters for different compilers

They are already separated into different methods. Can extract into different classes with some shared logic.

> 2. Pure pattern declaration in the filter
> 3. Generic matcher facility
>
> This should greatly reduce the complexity of the filters. I already started working on this on my early prototype.

In fact in the middle of implementation I tried to have pure declarative approach, but code became far more bigger and far less readable.

One of the complications with declarative approach here - is that some part of code uses imperative style inherently (see loop that counts number of resources while matching ECJ bytecode and names of variables that are not constants because represent different resources). Possibility to come up with some nice DSL not excluded, but doubt about ROI.

> If you give me a couple of days I can come up with a proposal.

Maybe we can do it vice versa - add descriptions, confirm that this implementation works, merge it and then peacefully work on "generic matching facility"? But no pressure :wink:

marchof:
@Godin I don't fully agree on all points, so I assume your remark about comments indicates that we have some serious code smells here.

Definitely I think the current filter implementation is hard to maintain and has some "write once" characteristics...

But I can agree on the proposed aproach: We make this work first and look for a declarative pattern matching algorithm later.

Godin:
@marchof just for the record: note about comments was not about "code smells", but about complexity of understanding bytecode (i.e. not how code does something, but why), so drafted overview/description on a wiki page - https://github.com/jacoco/jacoco/wiki/filtering-JAVAC.TRYWITH

Godin:
@marchof as we discussed by phone: test of number of instructions using

```java
import java.io.File;

import org.jacoco.core.analysis.Analyzer;
import org.jacoco.core.analysis.CoverageBuilder;
import org.jacoco.core.analysis.IBundleCoverage;
import org.jacoco.core.data.ExecutionDataStore;

public class CountInstructions {
	public static void main(String[] args) throws Exception {
		final CoverageBuilder coverageBuilder = new CoverageBuilder();
		new Analyzer(new ExecutionDataStore(), coverageBuilder)
				.analyzeAll(new File(args[0]));
		final IBundleCoverage bundle = coverageBuilder.getBundle("");
		System.out.println("Instructions: "
				+ bundle.getInstructionCounter().getTotalCount());

	}
}
```

As expected - no changes in 5 and 6, decrease in 7 and 8:

 rt.jar    | w/o filter | w/ filter
 ----------|------------|----------
 JDK 5u22  |  3 320 205 | 3 320 205
 JDK 6u45  |  4 390 154 | 4 390 154
 JDK 7u80  |  4 760 771 | 4 758 117
 JDK 8u121 |  4 700 913 | 4 695 163

Haven't thought about this before: can also use 3.8 Gb of 6865 JARs in my local Maven repository or/an 88135 artifacts from our Maven Central corporate proxy after sorting JARs on pre/post 7 :wink:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,426 @@
...
+			this.start = start;
+			cursor = start.getPrevious();
+			vars.clear();
+			labels.clear();
+			owners.clear();
+		}
+
+		private boolean matchJavac(final JavacPattern p) {
+			// "catch (Throwable t)"
+			if (!nextIs(Opcodes.ASTORE)) {
```
marchof: @Godin Is there a specific reason that you don't check the local variable here? The functionality is there and used for other ASTORE/ALOAD checks.

Godin: @marchof no good reasons, fixed

marchof: @Godin Thx! FYI there are more occurences left without variable checks:

- 2x nextIs(Opcodes.ALOAD)
- 1x nextIs(Opcodes.ASTORE)

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,426 @@
...
+			// "throw t"
+			if (!nextIsVar(Opcodes.ALOAD, "t2")) {
+				return false;
+			}
+			if (!nextIs(Opcodes.ATHROW)) {
+				return false;
+			}
+
+			final AbstractInsnNode end = cursor;
+			AbstractInsnNode c = start.getPrevious();
```
marchof: @Godin Looks like local variable c is not just a local loop counter but also carries some semantics here (as used later in `ignore()`. Maybe a name could explain what it is meant for (`startOfXxx`).

Godin:
@marchof I did one more update to wiki page and some thoughts come out of this:
* bytecode seems easier to understand without calls of ASM API even if they contain some comments (at least for me to see similarity/difference between blocks and repeated patterns), so maybe should be added as javadoc to tests?
* while combination of filtering for javac and ECJ in single class seemed like a good idea at first in order to reuse parts of code and same style/principles, overall bytecode pattern is quite different, so maybe better to split on two classes?
* maybe more validation/unit tests should be added to demonstrate certain things?

WDYT?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,426 @@
...
+				return false;
+			case FULL:
+			case OMITTED_NULL_CHECK:
+				return nextIsVar(Opcodes.ALOAD, "primaryExc")
+						// "if (primaryExc != null)"
+						&& nextIs(Opcodes.IFNULL)
+						// "r.close()"
+						&& nextIsClose("r") && nextIs(Opcodes.GOTO)
+						// "catch (Throwable t)"
+						&& nextIs(Opcodes.ASTORE)
```
Godin: @marchof I assume that in https://github.com/jacoco/jacoco/pull/500#discussion_r108958800 you refer to this `ASTORE` and to its pair - following `ALOAD`. Second remaining `ALOAD` is in code for ecj - I'll check it separately.

Here it is probably a bit more tricky than for previous: method `nextIsJavacClose` is called to check bytecode of both an exceptional and non-exceptional paths, and I suppose those instructions might use different variables on different paths, so would have to have different names, but should be doable.

marchof:
@Godin Thanks for updating the wiki page with your analyses! Helps me to understand the implementation.

- Re JUnit Documentation: Given that the primary lifecycle doc is our code base I would also prefer to have this somewhere in the JaCoCo Git repo. The unit test is probably the best place for now. In future we might use an Java bytecode assembler to directly write the test code in assembly language.
- Re Split Javac/Ecj: Make sense to me. I think this was one of my initial review comments.
- Re Testing: Primarly validation tests (but I have no specific idea what else to test, especially I have no idea how we can test for false positives better than your rt.jar test). I wouldn't add unit tests only for the sake of code coverage -- here I still hope to find a more declarative approach which comes with less control structures within the specific filters.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
marchof: @Godin No prob. I just did a systematic re-check after your fix and wanted to let you know.

Godin:
@marchof 

> but I have no specific idea what else to test

For example in case of `return` in `catch` block

```java
class Fun implements java.io.Closeable {
  Object fun() {
    try (Fun r1 = new Fun()) {
      return read(r1);
    } catch (RuntimeException e) {
      return "catch";
    } finally {
      System.out.println("finally");
    }
  }

  public void close() {
  }

  Object read(Fun r) {
    return r;
  }
}
```

code of `finally` block will be emitted **3 times** while AFAIK in all existing validation tests just two times and one (like me just few minutes ago) could think that this is artefact of `try-with-resources`, but exactly the same happens without it

```java
class Fun2 {
  Object fun() {
    try {
      read();
      return "body";
    } catch (RuntimeException e) {
      return "catch";
    } finally {
      System.out.println("finally");
    }
  }

  void read() {
  }
}
```

so maybe both can be added to visually demonstrate that this case remains for the `finally` filter.

marchof:
@Godin Of course, the validation tests (executed with javac and ecj) should trigger at least all positive paths in the filter.

marchof:
@Godin Regarding your idea with tests on the Maven repo: For many artifacts there are source attachements. How hard would it be to identify try-with-resource statements on source level (e.g. with Sonar language tools)?

So taking this a step further one could think of a test where for every method found in the Maven repo a specific language construct is detected on source level. The corresponding bytecode filter should come to the same result. That would be a crazy integration test for the filters...

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
Godin: @marchof good catch - I indeed overlooked this :thumbsup: fixed

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,427 @@
...
+					&& nextIsClose(name);
+		}
+
+		private boolean nextIsEcjCloseAndThrow(final String name) {
+			return nextIsVar(Opcodes.ALOAD, name)
+					// "if (r != null)"
+					&& nextIsJump(Opcodes.IFNULL, name)
+					// "r.close()"
+					&& nextIsClose(name) && nextIsLabel(name)
+					&& nextIs(Opcodes.ALOAD) && nextIs(Opcodes.ATHROW);
```
Godin: this remaining unnamed `ALOAD` that was found in https://github.com/jacoco/jacoco/pull/500#discussion_r108958800 should be named "primaryExc"

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java
```diff
@@ -0,0 +1,314 @@
...
+		private boolean nextIsAddSuppressed() {
+			if (!nextIs(Opcodes.INVOKEVIRTUAL)) {
+				return false;
+			}
+			final MethodInsnNode m = (MethodInsnNode) cursor;
+			return "java/lang/Throwable".equals(m.owner)
+					&& "addSuppressed".equals(m.name);
+		}
+
+		private boolean nextIsAddSuppressed(final String name) {
```
Godin: can be inlined

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,244 @@
...
+						&& nextIsVar(Opcodes.ALOAD, ctx + "t")
+						&& nextIsAddSuppressed() && nextIs(Opcodes.GOTO)
+						// "r.close()"
+						&& nextIsClose("r");
+			default:
+				return false;
+			}
+		}
+
+		private boolean nextIsClose(final String name) {
```
Godin: `name` is always `"r"`

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
```diff
@@ -0,0 +1,864 @@
...
+import java.util.List;
+
+import org.jacoco.core.internal.instr.InstrSupport;
+import org.junit.Test;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+
+public class TryWithResourcesFilterTest implements IFilterOutput {
```
marchof: @Godin If the filter is split into two classes (javac/ecj) I wonder whether it would make sense to also split the unit test?

marchof:
@Godin Although not directly related to this PR: Would be nice to add JavaDoc to `IFilter.filter()`. This is currently the only compiler warning in Eclipse.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilterTest.java
Godin: @marchof indeed, done

org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/targets/TryWithResources.java
```diff
@@ -109,6 +110,35 @@ private static void nested() {
 		}
 	}
 
+	/**
+	 * In this case bytecode will contain 3 copies of <code>finally</code>
+	 * block, each containing 2 branches, resulting in 6 branches in total. One
+	 * could think that this is artifact of try-with-resources, but the same
+	 * happens without it.
```
Godin: @marchof just a note: now recalling semantic of `finally` it seems quite logical for me that in case of non-exceptional flow from `catch` there will be one more duplicate of `finally` block and won't be in case of exceptional (`throw`), but IMO this is not obvious at a first glance

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java
```diff
@@ -0,0 +1,310 @@
...
+			if (cursor.getOpcode() != Opcodes.GOTO) {
+				return false;
+			}
+
+			output.ignore(c.getNext(), cursor);
+			output.ignore(start.getNext(), end);
+			return true;
+		}
+
+		private boolean matchEcjNoFlowOut() {
```
Godin: return value is ignored

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
Godin: @marchof done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesFilter.java
```diff
@@ -0,0 +1,245 @@
...
+			startOnNonExceptionalPath = startOnNonExceptionalPath.getNext();
+
+			final AbstractInsnNode m = cursor;
+			next();
+			if (cursor.getOpcode() != Opcodes.GOTO) {
+				cursor = m;
+			}
+
+			output.ignore(startOnNonExceptionalPath, cursor);
+			output.ignore(start.getNext(), end);
```
Godin: `start.getNext()` here just skips label at the start of handler, so useless, but first of all unclear at a first glance

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java
```diff
@@ -0,0 +1,310 @@
...
+					return false;
+				}
+			}
+			next();
+			if (cursor.getOpcode() != Opcodes.GOTO) {
+				return false;
+			}
+
+			output.ignore(c.getNext(), cursor);
+			output.ignore(start.getNext(), end);
```
Godin: same as https://github.com/jacoco/jacoco/pull/500#discussion_r109057541

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java
```diff
@@ -0,0 +1,310 @@
...
+			// "throw primaryExc"
+			if (!nextIsVar(Opcodes.ALOAD, "primaryExc")) {
+				return false;
+			}
+			if (!nextIs(Opcodes.ATHROW)) {
+				return false;
+			}
+
+			final AbstractInsnNode end = cursor;
+			c = start.getPrevious();
```
Godin: same as https://github.com/jacoco/jacoco/pull/500#discussion_r108978298

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java
```diff
@@ -0,0 +1,310 @@
...
```
Godin: same as https://github.com/jacoco/jacoco/pull/500#discussion_r108978298

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
```diff
@@ -0,0 +1,245 @@
...
+		private boolean nextIsAddSuppressed() {
+			if (!nextIs(Opcodes.INVOKEVIRTUAL)) {
+				return false;
+			}
+			final MethodInsnNode m = (MethodInsnNode) cursor;
+			return "java/lang/Throwable".equals(m.owner)
+					&& "addSuppressed".equals(m.name);
+		}
+
+		private boolean nextIsVar(final int opcode, final String name) {
```
Godin: @marchof can extract common part of two filters (`cursor`, `nextIs`, `nextIsVar`) into abstract class to get rid of duplication and reduce their size/complexity. WDYT?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
```diff
@@ -0,0 +1,245 @@
...
+			}
+			final AbstractInsnNode end = cursor;
+
+			AbstractInsnNode startOnNonExceptionalPath = start.getPrevious();
+			cursor = startOnNonExceptionalPath;
+			while (!nextIsJavacClose(p, "n")) {
+				startOnNonExceptionalPath = startOnNonExceptionalPath
+						.getPrevious();
+				cursor = startOnNonExceptionalPath;
+				if (cursor == null) {
```
Godin: @marchof maybe can get reduce amount of branches by changing `nextIs` and `nextIsVar` methods: if cursor is `null`, then do nothing, instead of returning `false` set cursor to `null`, the loop to find closing of resource on non-exceptional path already contains null-check. However question of performance of such approach is quite interesting: will JVM be able to optimize all this by inlining in order to retain early exits in case of missmatches? if not, then what will be cost of unnecessary monomorphic method calls? null-check supposed to be cheap as JVM anyway need to do it during field access. WDYT?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
marchof: @Godin :+1:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractTryWithResourcesMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.VarInsnNode;
+
+abstract class AbstractTryWithResourcesMatcher {
```
marchof: @Godin Why not `AbstractMatcher`? And use the same also for `SynchronizedFilter`. The functionality looks very generic.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
marchof: @Godin I personally prefer readable over performance optimized code. But as discussed yesterday for know I would focus on functional correctness. Once we have a couple of working and tested filters in place we can try different implementation aproaches to simplify filter definition. BTW here is an example how the (way more simple) `SynchronizedFilter` could look like: https://github.com/jacoco/jacoco/blob/experimental-filter-patterns/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilter.java

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractTryWithResourcesMatcher.java
Godin: @marchof indeed and I've been thinking about this - in such case `SynchronizedFilter` will pay cost of unnecessary allocation of `Map vars`. This cost can be removed by addition of a contract for subclasses - it should be initialized in subclasses if used. Before doing either of those changes I want to do at least some basic measurement as a sanity check.

marchof: @Godin The map could be initialized lazily. On the other hand `SynchonizedFilter `could also check local variables.

Godin: indeed, this is also an option, give me some time to check

Godin: @marchof I don't observe degradations when `SynchonizedFiltercould` checks local variable, so done. Note that there is only two "paired" instructions on same variable and only in case of javac.

marchof: @Godin 👍 😎 (just figured out how to insert these 😉 )

Godin:
> Regarding your idea with tests on the Maven repo: For many artifacts there are source attachements. How hard would it be to identify try-with-resource statements on source level (e.g. with Sonar language tools)?

Indeed - I forgot that 3.8 Gb of 6865 JARs in my local Maven repository also includes some JARs with sources. However sources are not downloaded by default, so not that much: 309 artifacts with sources, their corresponding JARs with classes account for 125 Mb, and diversity also maybe not that big - see [log.txt](https://github.com/jacoco/jacoco/files/888698/log.txt)

> So taking this a step further one could think of a test where for every method found in the Maven repo a specific language construct is detected on source level. The corresponding bytecode filter should come to the same result. That would be a crazy integration test for the filters...

ECJ filter is not triggered by any of the above scanned JARs. In case of javac first instruction of a first ignored block is a line number corresponding to `try` keyword and matches to 174 `try` keywords of `try-with-resources` parsed from sources, 14 582 of normal `try` statements are skipped by filters.

Doing the same test using `rt.jar` and `src.zip` of JDK 7u80 and JDK 8u121: matched correctly 52 `try` keywords of `try-with-resources` and correctly skipped 13102 of normal `try` statements.

For bigger set of artifacts with sources will need to think about how to obtain them effectively: don't know how many sources is stored in cache in our corporate proxy as they are not downloaded by default; and anyway will need to think about how to get list no matter from where to download - central or proxy.

org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/TryWithResourcesTest.java
```diff
@@ -0,0 +1,175 @@
...
+	 * {@link TryWithResources#returnInBody()}
+	 */
+	@Test
+	public void returnInBody() {
+		// without filter next line covered partly:
+		assertLine("returnInBody.try", ICounter.FULLY_COVERED);
+		assertLine("returnInBody.open", ICounter.FULLY_COVERED);
+		// without filter next line has branches:
+		// TODO with javac 1.8.0_31 part of return instructions belong to close
+		// assertLine("returnInBody.close", ICounter.EMPTY);
```
Godin: In case of

```java
private static Object returnInBody() throws IOException {
    try ( // $line-returnInBody.try$
        Closeable r = new Resource() // $line-returnInBody.open$
    ) {
        return read(r); // $line-returnInBody.return$
    } // $line-returnInBody.close$
}
```

we correctly find instructions, but javac 7 as well as javac 8 up to 8u77 (tested public updates) generates:

```
LINENUMBER $line-returnInBody.return$
...
ASTORE
LINENUMBER $line-returnInBody.close$
... // close resource
ALOAD
ARETURN
LINENUMBER $line-returnInBody.try$
... // save exception
LINENUMBER $line-returnInBody.close$
... // close resource
```

while javac starting from 8u92 (notice additional `LINENUMBER` before `ALOAD`) as well as latest 9 EA:

```
LINENUMBER $line-returnInBody.return$
...
ASTORE
LINENUMBER $line-returnInBody.close$
... // close resource
LINENUMBER $line-returnInBody.return$
ALOAD
ARETURN
LINENUMBER $line-returnInBody.try$
... // save exception
LINENUMBER $line-returnInBody.close$
... // close resource
```

Looks like this relates to [JDK-8134759](https://bugs.openjdk.java.net/browse/JDK-8134759) (log of bisect - [bisect.txt](https://github.com/jacoco/jacoco/files/892452/bisect.txt))

There is no such problem when `finally` is present, but this case is covered by another existing test.

So I'm wondering what can we do with this test?

Travis provides JDK 8u31 by default and we also test JDK 8 EA. Can imagine that for the bytecode version 7 generated by javac we will be testing old behavior, while will be testing and using updated stable JDK 8 in Travis as well as locally.

But maybe @marchof you have some other thoughts/ideas?

bjkail: @Godin Parse `java.version`, and then skip or adjust the asserts?  We have precedence in `BadCycleInterfaceTest`.

Godin: @bjkail `java.version` is about version of JRE and `BadCycleInterfaceTest` is exactly about runtime behavior. This will work here if we assume that execution and compilation of tests use the same JDK version.

bjkail: @Godin You're right.  Is that too fragile?  If so, the only other option that comes to mind is to use ASM to detect the missing line number.  That would weaken the test a bit, but perhaps it's better than nothing?

marchof: @Godin @bjkail Maybe our integration tests for filtering are too specific. What if we just assert that our filtering examples do not produce any uncovered lines and branches? We already have plenty of validation test for expected covered/missed lines.

bjkail: @marchof Sounds fine.  I was just providing some ideas and am indifferent on the actual fix.

Godin: @marchof Excuse me for the delay with reply - was at DevoxFR, where BTW session "Java Code Coverage Mechics" was a quite success.

On a subject:

@bjkail `java.version` maybe not so fragile.

> ASM to detect the missing line number

is fragile in terms that it allows compiler to regress on this.

@marchof 

> Maybe our integration tests for filtering are too specific.

I personally like such specificity for the following reasons
* not all examples here reach full coverage (no uncovered lines and branches) and also IMO will not reach in case of future filters
* consistency across tests
* allows to see exactly what user will see
* allows to catch regressions in compilers

> We already have plenty of validation test for expected covered/missed lines

Apparently not for such tricky control flows as shown here :wink: 

Do you see any troubles to simply bump version of javac 8 up to latest update as in my initial proposal - "past is past"?

marchof: @Godin We have a large user base out there where older JDK versions still might be in use. The primary purpose of the integration tests for filters should be to ensure that the filters fulfill their purpose. I would imagine filter integration tests consisting of source code only (maybe with annotated lines for the intentionally un-covered ones).

My primary concern is maintainability. If you think we can handle the specific checks in the long run, I can live with the current approach.

Godin: @marchof 

> We have a large user base out there where older JDK versions still might be in use. 

I haven't said that we should stop testing all old JDK versions - only range of **JDK 8 till 8u92**.
JDK 7 will be testing the same bytecode that generates JDK 8 from this range. And this range was shown to be working with other existing test cases.

I wish that we run tests against range of all public updates of all supported JDK versions, but this is quite overkill and we are not doing this anyway :wink: 

Some other options:
* version of compiler can be passed by Maven into tests, but then such conditional assertions should be disabled in IDEs
* tests can perform execution of compiler so that they will precisely know version independently from build system and IDEs

But all this quite out of scope of this PR. So for now I excluded assertion on this line only for bytecode version 8 - this is at least better that current TODO-comment and we can come back to this discussion later.

Godin: > excluded assertion on this line only for bytecode version 8

arr, in fact even this is fragile as one can ask generation of bytecode version 7 or 8 using JDK 9

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
```diff
@@ -0,0 +1,253 @@
...
+	 * Case of multiple resources looks like multiple nested try-with-resources
+	 * statements. javac from JDK 9 EA b160 does the same, but with some
+	 * optimizations (see <a href=
+	 * "https://bugs.openjdk.java.net/browse/JDK-7020499">JDK-7020499</a>):
+	 * <ul>
+	 * <li><code>null</code> check for resource is omitted when it is
+	 * initialized using <code>new</code></li>
+	 * <li>synthetic method <code>$closeResource</code> containing
+	 * <code>null</code> check of primaryExc and calls to methods
+	 * <code>addSuppressed</code> and <code>close</code> is used when numbers of
```
Godin: typo - `numbers` instead of `number`

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java
Godin: @marchof I made this refactoring and don't observe impact on performance. As expected - far less of uncovered branches. And IMO far more declarative and more readable. Also added explanatory javadocs.

org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/TryWithResourcesTest.java
Godin: Experimented with this over past weeks and few times experienced fragility on
`mvn clean install -Dbytecode.version=1.7` using JDK 8, so finally in favor of @bjkail  suggestion to assume that execution and compilation of tests is done with same JDK.

Godin:
@marchof I scanned bigger subset of Maven Central - latest versions of some artifacts within group `org`. This was not easy, because not counting time that were spent to download - some jars with sources should be excluded as they are incomplete, contain more than jars with classes,  contain templates for generation of sources, etc...

Nevertheless **9753** try-with-resources statements were filtered correctly from **≈20 Gb** of JAR files with classes. And this time ECJ filter was triggered quite a good amount of times.

Case that was already presented in our tests and not filtered - when body of try-with-resources is empty
```
try (Resource r = new Resource()) {
}
```
was encountered in
- org/apache/cassandra/cassandra-all/3.10
- org/apache/cayenne/modeler/cayenne-modeler/4.0.M5
- org/apache/reef/reef-runtime-hdinsight/0.15.0
- org/apache/tomcat/embed/tomcat-embed-jasper/9.0.0.M19
- org/apache/tomcat/tomcat-jasper/9.0.0.M19
- org/apache/solr/solr-core/6.5.0
- org/eclipse/scout/sdk/deps/org.eclipse.jdt.launching/3.8.100.v20160505-0636
- org/integratedmodelling/klab-common/0.9.10
- org/jboss/arquillian/container/arquillian-gae-common/1.0.0.Beta10
- org/jboss/cdi/tck/cdi-tck-impl/2.0.0.CR1
- org/neo4j/app/neo4j-server/3.2.0-alpha07
- org/neo4j/neo4j-ha/3.2.0-alpha07
- org/neo4j/neo4j-ogm/1.1.6
- org/sonatype/nexus/bundles/org.sonatype.nexus.bundles.elasticsearch/3.2.1-01
- org/sonarsource/sonarqube/sonar-process/6.3
- org/simplericity/jettyconsole/jetty-console-core/1.60
- org/vesalainen/comm/1.0.3
- org/qi4j/core/org.qi4j.core.testsupport/2.1

Another funny case that was already presented in our tests and not filtered - when try-with-resources contains only `throw` statement
```
try (Resource r = new Resource()) {
  throw new Exception();
}
```
was encountered in
- org/apache/nifi/nifi-framework-core/1.1.2
- org/apache/reef/reef-common/0.15.0
- org/codelibs/fess/fess-crawler/1.1.0
- org/tentackle/tentackle-persistence/2.0.7
- org/sonatype/nexus/bundles/org.sonatype.nexus.bundles.elasticsearch/3.2.1-01

Not explicitly presented in out tests, but a kind of variation of previous and not filtered - when there is no non-exceptional path from a body as for example
```
try (Resource r = new Resource()) {
  while (true) {
    read(r);
  }
} catch (EOFException e) {
}
```
was encountered in
- org/apache/cassandra/cassandra-all/3.10
- org/apache/fluo/fluo-accumulo/1.0.0-incubating
- org/github/evenjn/knit/0.6.0
- org/infinispan/infinispan-embedded/9.0.0.Final
- org/jgroups/jgroups/4.0.1.Final
- org/kohsuke/elb-dns/1.1

An interesting case where source code contains try-with-resources, but [retrolambda](https://github.com/orfjackal/retrolambda#backported-language-features) removes call to `Throwable.addSuppressed` and so not filtered, was encountered in
- org/ansj/ansj_seg/5.1.1

An interesting case where AspectJ adds try-with-sources into bytecode was encountered in
- org/no-hope/aspectj/metrics-aspectj-el/1.7

All in all: I satisfied by results and would like to merge this PR. We can deal with cases of an empty body and absence of non-exceptional path in a separate PR. If you agree I'll squash and merge into master.

marchof:
@Godin Awesome work, Evgeny! I'm absolutly curious about the exact test setup. And yes, we can add corner cases later.

I was already about to ask about the performance of the new, way more readable matching code. As there is no negative impact -> yes, please merge this!

romani:
@Godin , when do you plan to release this ?
Checkstyle project plan to give jacoco another try this summer, this feature could be a good stimulus to switch to jacoco for coverage.

Godin:
@romani please see already existing answers on such question:
https://groups.google.com/forum/#!msg/jacoco/gd8xD30TDNo/SHj-hEq1AAAJ
https://groups.google.com/d/msg/jacoco/-MmsVzC6szM/Q9sPTbAeBAAJ

PR #501, opened 2017-03-19
Exclude from a report a part of bytecode that compiler generates for a synchronized statement and that is hard to cover by tests

Godin:


Godin:
Analysis of 17391 classes from `rt.jar` (≈ 65 Mb) from JDK 1.8.0_121 using following `build.xml`
```xml
<project xmlns:jacoco="antlib:org.jacoco.ant" default="report">
    <classpath path="jacocoant.jar"/>
  </taskdef>
  <target name="report">
    <jacoco:report>
      <executiondata>
        <file file="jacoco.exec"/>
      </executiondata>
      <structure name="rt.jar">
        <classfiles>
          <fileset file="rt.jar"/>
        </classfiles>
      </structure>
    </jacoco:report>
  </target>
</project>
```
takes 5 seconds, both without this change as well as with it. So I believe we are safe to merge this.

@marchof do you have objections or some review comments?

org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesVisitor.java
```diff
@@ -147,4 +148,8 @@ public void visitLookupSwitchInsnWithProbes(final Label dflt,
 			final int[] keys, final Label[] labels, final IFrame frame) {
 	}
 
+	public void accept(final MethodNode methodNode,
```
marchof: Whooo, here we start mixing the ASM event and tree API. I consider this as a temporary hack until we fully move to the tree API ;-)

Nevertheless we should clearly document the method, especially the point in time when it is called.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
```diff
@@ -108,6 +108,7 @@ public void visitEnd() {
 	public void analyzeClass(final ClassReader reader) {
 		final ClassVisitor visitor = createAnalyzingVisitor(
 				CRC64.checksum(reader.b), reader.getClassName());
+		// TODO maybe we can SKIP_FRAMES here?
```
marchof: I agree in the comment but separate issue? On the other hand I'm not sure what the benefit would be as frames are ignored at analysis time.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
```diff
@@ -86,6 +99,38 @@ public IMethodCoverage getCoverage() {
...
+			final MethodVisitor methodVisitor) {
+		this.ignored.clear();
+		for (final IFilter filter : FILTERS) {
+			filter.filter(methodNode, this);
+		}
+
+		for (final TryCatchBlockNode n : methodNode.tryCatchBlocks) {
+			n.accept(methodVisitor);
+		}
+		for (int i = 0; i < methodNode.instructions.size(); i++) {
```
marchof: If a InsnList is traversed by index an internal array is created lazily. This can probably be avoided by using the linked list API.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
```diff
@@ -25,6 +27,8 @@
 
 	private Instruction predecessor;
 
+	public AbstractInsnNode node;
```
marchof: As it is always set after the instance is created this should be final and set within the constructor.

For consistency I would prefer to make it private with a public getter.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -0,0 +1,20 @@
...
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.tree.AbstractInsnNode;
+
+public interface IFilterOutput {
+
+	void ignore(AbstractInsnNode from, AbstractInsnNode to);
```
Godin: @marchof Actually both are inclusive. And we already use term "successor" in analysis of CFG, where this usage implies that not each instruction is a "successor" of preceding instruction from a list :wink: so not sure that "successor" will be a good unambiguous word here.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: @marchof I believe that in this case `FrameNode` won't be appear in `methodNode.instructions` and since they anyway not used for analysis, we won't be wasting CPU cycles to skip them in filters - see method `SynchronizedFilter.Matcher#next()`. While this doesn't play any role for `SynchronizedFilter`, the same method is used in filter for try-with-resources (#500). And now that you reminded me about that - probably I can demonstrate observable impact on performance using its implementation prior to the last optimizations. I'm fine with postponing this, however this shouldn't be forgotten and in fact I almost forgot about this, but good that left this comment.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Synchronized.java
```diff
@@ -0,0 +1,59 @@
...
+public class Synchronized {
+
+	private static final Object lock = new Object();
+
+	private static void normal() {
+		nop(); // $line-before$
+		synchronized (lock) { // $line-monitorEnter$
+			nop(); // $line-body$
+		} // $line-monitorExit$
+	} // $line-after$
```
Godin: Such usage seems legitime to me given that implicit `return` is always presented and always assigned to such line as verified by another test. Moreover we already use such pattern in `Target03` and `EnumImplicitMethods`, however in `InterfaceDefaultMethodsTarget` and `InterfaceOnlyDefaultMethodsTarget` we use explicit `return` statements :laughing: And BTW I've just realized that we test "before" and "after" only in case of exceptions, but not in case of control structures :wink:

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
Godin: @marchof as link in Javadoc shows - this was inspired by implementation of `org.objectweb.asm.tree.MethodNode#accept(org.objectweb.asm.MethodVisitor)`, so strictly speaking there is no change in behavior in regards to creation of array. But you right that we can avoid this. And if so, then I'm wondering - how to prevent regression of this? Probably it can/should be avoided for instrumentation too? Are there other places that trigger or might trigger its creation?

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilterTest.java
```diff
@@ -0,0 +1,74 @@
...
+import static org.junit.Assert.assertNull;
+
+import org.jacoco.core.internal.instr.InstrSupport;
+import org.junit.Test;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+
+public class SynchronizedFilterTest implements IFilterOutput {
```
Godin: @marchof I agree that code generated by javac should be captured in unit test and will do this.

However regarding

> full coverage on its own (with all paths)

I doubt that generation of synthetic non-realistic bytecode in unit test or change of the code just to achieve full branch/path coverage makes sense. This is probably not that hard here for `SynchronizedFilter`, but I expect this to be really tedious in case of filter for try-with-resources.

Godin: @marchof BTW we probably can exclude/separate verification tests from unit test coverage to enforce practice of completeness of unit tests.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
Godin: @marchof indeed - this is a leftover of an early prototyping.

org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesVisitor.java
Godin: "Nothing is more permanent than the temporary" :wink: but future will show. We kind of already had such mix - see `org.jacoco.core.internal.flow.LabelFlowAnalyzer#markLabels`, and exactly this gave me an idea of how to use ASM Tree API for filters without massive changes of everything, so this mix is worth it.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilter.java
```diff
@@ -0,0 +1,90 @@
...
+
+	public void filter(MethodNode methodNode, IFilterOutput output) {
+		for (TryCatchBlockNode tryCatch : methodNode.tryCatchBlocks) {
+			if (tryCatch.type != null) {
+				continue;
+			}
+			if (tryCatch.start == tryCatch.handler) {
+				continue;
+			}
+			final AbstractInsnNode to = new Matcher(tryCatch.handler).match();
```
Godin: Number of allocations of `Matcher` can be reduced here - can allocate and reuse just one for the whole lifetime of a method instead of allocation for `TryCatchBlockNode`.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilterTest.java
```diff
@@ -0,0 +1,74 @@
...
+		m.visitLabel(handler);
+		m.visitVarInsn(Opcodes.ALOAD, 1);
+		m.visitInsn(Opcodes.MONITOREXIT);
+		m.visitLabel(handlerEnd);
+		m.visitInsn(Opcodes.ATHROW);
+		m.visitLabel(exit);
+		m.visitInsn(Opcodes.RETURN);
+
+		new SynchronizedFilter().filter(m, this);
+		assertEquals(m.instructions.get(12), from);
```
Godin: Sorry, but I'm not sure that addition of mark for each line is a good idea -  in case of changes this will force to change the whole list of subsequent instructions and will probably affect diff in a bad way complicating review of changes. Also IMA harder to spot target of a label that is represented as number instead of variable.

But I agree that `12` and `16` here are not understandable.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
Godin: In fact I was wrong - default implementation in `org.objectweb.asm.tree.MethodNode#accept(org.objectweb.asm.MethodVisitor)` does not construct array, don't know from where opposite came to my mind. Still wondering how to prevent such regressions, especially given that `MethodNode#instructions` is accessible by filters now.

org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesVisitor.java
Godin: > Nevertheless we should clearly document the method, especially the point in time when it is called.

Not sure I understand this point - do you want to document method or its call site or both? And in any case - do you have idea about wording?

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
bjkail: Only suggestions that come to mind are:
1. Add build-time class file scanning for use of `InsnList.get(int)`.
2. Create `InsnList` subclass that throws and use it in unittests.  I guess this would require subclassing `ClassAnalyzer` to override `visitMethod`, which probably requires reimplementing `Analyzer.analyzeClass`/`createAnalyzingVisitor`.

marchof: More suggestions:

1. A reflection based assertion that `InsnList.cache` is stays `null`
2. A JaCoCo coverage check that method `InsnList.toArray()` is not executed during our tests.

But I woudn't worry too much right now - unless performance testing shows a bottle neck here.

org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesVisitor.java
marchof: @Godin The method itself:

	/**
	 * This method can be overwritten to hook into the process of emitting the
	 * instructions of this method as <code>visitX()</code> events.
	 * 
	 * @param methodNode
	 *            the content to emit
	 * @param methodVisitor
	 *            A visitor to emit the content to. Note that this is not
	 *            necessarily this visitor instance but some wrapper which
	 *            calculates the probes.
	 */

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilterTest.java
```diff
@@ -0,0 +1,74 @@
...
+	private AbstractInsnNode from;
+	private AbstractInsnNode to;
+
+	@Test
+	public void ecj() {
+		final Label start = new Label();
+		final Label end = new Label();
+		final Label handler = new Label();
+		final Label handlerEnd = new Label();
+		m.visitTryCatchBlock(start, end, handler, null);
```
marchof: We can use a modified version of my builder API which returns a label or the instruction node created for each visitX() method. Then we can store the corresponding nodes directly in a local variable:
    
    builder.visitA();
    builder.visitB();
    AbstractInsnNode expectedFrom = builder.visitC();
    builder.visitD();
    builder.visitE();
    builder.visitF();
    AbstractInsnNode expectedTo = builder.visitG();

But e can use the "magic numbers" for now and I prepare a separate PR for testing.

Godin: @marchof I added commit to use labels instead of magic numbers without any new builders.

marchof: @Godin Ok, great! Just saw the new commit.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
Godin: And one more suggestion 😆  - `InsnList` as well as whole `MethodNode` comes to us via `MethodSanitizer#visitEnd()` in `ClassProbesAdapter`, where we can replace value of field `instructions` (that is public in `MethodNode`) by a new instance with overriden method `toArray()`.

> But I woudn't worry too much right now - unless performance testing shows a bottle neck here.

I've checked that our code doesn't trigger creation of array anymore.
So indeed - let's don't worry about this right now.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: @marchof I removed comment from this PR as anyway it will be presented in #500 , so that we can come back to this later.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
Godin: @marchof fixed

org.jacoco.core/src/org/jacoco/core/internal/flow/MethodProbesVisitor.java
Godin: @marchof done

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilterTest.java
Godin: @marchof fixed :white_check_mark:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
Godin: @marchof Javadoc added :white_check_mark:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.tree.AbstractInsnNode;
+
+public interface IFilterOutput {
```
Godin: @marchof done :white_check_mark:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilter.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.tree.MethodNode;
+
+public interface IFilter {
```
Godin: @marchof done :white_check_mark:

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Synchronized.java
Godin: @marchof added `nop()` here so that to have a symmetry with `before` marker at least in this test :white_check_mark:

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/SynchronizedFilterTest.java
Godin: @marchof added bytecode that `javac` produces and negative test. However this does not lead to a full branch coverage (few are still missed), but at least main paths are covered :white_check_mark:

marchof: @Godin No problem. I assume we will need a (well tested) general purpose matcher facility soon. Then the number of branches within the specific filters will greatly reduce.

marchof:
@Godin Thanks for following up all my picky comments ;-) Let's merge this and continue with another filtering PR. I assume the overall implementation will improve step by step as we see more requirements and use cases.

Godin:
@marchof Hooray! 😋  Will do squash and merge quickly and then will rebase #500 so that to continue with filter that doesn't require changes outside of filter 😉

PR #503, opened 2017-03-20
Use new compiler detection facility

marchof:
Use new `ValidationTestBase.isJDKCompiler` constant for missing assertions.


marchof:
@Godin Can you please review? Thx

Godin:
@marchof this is quite interesting example - partial line coverage is caused by the fact that `while` statement is preceded by `if` statement :laughing:

PR #504, opened 2017-03-23
Separate structure constructs in validation tests 

marchof:
As noted by @Godin we actually saw a side effect in #503. To avoid such side effects the tested structures should better be tested in separate methods.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ControlStructuresTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ControlStructuresTest.java
@@ -41,8 +41,7 @@ public class ControlStructuresTest extends ValidationTestBase {
 		assertLine("missedelse", ICounter.NOT_COVERED);
 
 		// 4. Missed while block
-		assertLine("whilefalse", isJDKCompiler ? ICounter.FULLY_COVERED
-				: ICounter.PARTLY_COVERED, 1, 1);
+		assertLine("whilefalse", ICounter.FULLY_COVERED, 1, 1);
 		assertLine("missedwhile", ICounter.NOT_COVERED);
 
 		// 5. Always true while block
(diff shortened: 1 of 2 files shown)


org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target01.java
```diff
@@ -173,40 +246,41 @@ public static void main(String[] args) {
...
 		}
 		nop(); // $line-afterreturn$
 
 	}
 
-	private static void runImplicitReturn() {
+	private static void implicitReturn() {
 
-		// 21. Implicit return
 	} // $line-implicitreturn$
```
Godin: IMO dedicated method is not needed for `$line-implicitreturn$` - it can be placed into `returnStatement`.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ControlStructuresTest.java
```diff
@@ -41,8 +41,7 @@ public void testCoverageResult() {
 		assertLine("missedelse", ICounter.NOT_COVERED);
 
 		// 4. Missed while block
-		assertLine("whilefalse", isJDKCompiler ? ICounter.FULLY_COVERED
-				: ICounter.PARTLY_COVERED, 1, 1);
```
Godin: Anyway this case was quite interesting IMO, so maybe we should preserve it somewhere?

marchof: Opened issue #505.

org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target01.java
marchof: @Godin Correct, every target method without an return statement could serve as a test target. Would prefer to keep aspects in different methods though.

What is actually missing is a test case for a explicit unconditional return. In this case there should be no code assigned to the closing brace. Will add this.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ControlStructuresTest.java
Godin: Okay.

org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target01.java
Godin: You're right.

PR #511, opened 2017-04-02
Use new filtering API for existing filter of synthetic methods

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java
```diff
@@ -0,0 +1,30 @@
...
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.MethodNode;
+
+/**
+ * Filters synthetic methods.
```
marchof: Maybe we should be more specific here: "Filters synthetic methods unless they represent desugared lambda bodies."

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -102,8 +102,7 @@ private boolean isMethodFiltered(final String className,
 				return true;
 			}
 		}
-		return (access & Opcodes.ACC_SYNTHETIC) != 0
```
marchof: Deleting these lines needs more cleanup to get rid of compiler warnings:

- `import org.objectweb.asm.Opcodes;` is unused and should be deleted
- Method parameter `access` is unused and should be removed

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java
Godin: @marchof fixed

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof fixed

PR #512, opened 2017-04-03
Use new filtering API for existing filter of methods in enums

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilter.java
```diff
@@ -23,12 +23,17 @@
...
+	 *            class name
+	 * @param superClassName
+	 *            superclass name
 	 * @param methodNode
 	 *            method to inspect
 	 * @param output
 	 *            callback to report filtering results to
 	 */
-	void filter(MethodNode methodNode, IFilterOutput output);
+	void filter(String className, String superClassName, MethodNode methodNode,
```
marchof: @Godin We might need to add more params in the future here. What do you think about a new interface `IFilterContext` where we pass a instance here?

Godin: @marchof Well, I'd leave it simple as it is for the time being.

marchof: @Godin Good for me!

PR #521, opened 2017-04-22
Merge duplicate finally blocks in a report

Godin:
https://github.com/jacoco/jacoco/wiki/filtering-JAVAC.FINALLY


Godin:
@marchof To merge conditional jumps where different branches are covered in different instructions we need to determine which branch covered - this is what first change does. Second introduces actual merging. While there is some room for improvements, changes are pretty straightforward, and so main care should be paid for third change that introduces filter for finally blocks. Thus wondering what you think about it?

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
```diff
@@ -102,11 +121,11 @@ public int getBranches() {
...
 	/**
-	 * Returns the number of covered branches starting from this instruction.
-	 * 
-	 * @return number of covered branches
+	 * Returns coverage of branches of this instruction.
+	 *
+	 * @return coverage of branches of this instruction
 	 */
-	public int getCoveredBranches() {
+	public BitSet getCoveredBranches() {
```
marchof: @Godin I would prefer to keep the BitSet internally. So

- Keep `int getCoveredBranches()`
- Add `void merge(Instruction)`

This encapsulation looks cleaner to me allows internal optimizations later.

marchof:
@Godin I definitely like the straightforward approach! I will go though the implementation and add comments if I have specific questions or concerns.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
```diff
@@ -65,20 +69,35 @@ public void addBranch() {
 	 * @see #addBranch()
 	 * @param predecessor
 	 *            predecessor instruction
+	 * @param branch
+	 *            branch number in predecessor that should be marked as covered
+	 *            when this instruction marked as covered
 	 */
-	public void setPredecessor(final Instruction predecessor) {
+	public void setPredecessor(final Instruction predecessor,
```
marchof: Unrelated to this change but while trying to understand my own code after some time I wondered, why a instruction can't have multiple predecessors: While this is of course possible in byte code JaCoCo adds probes in a way that each instruction has at most one predecessor. Otherwise probes are inserted. Maybe we can improve documentation here:

> Sets the given instruction as a direct predecessor to this instruction. Every instruction can have at most one direct predecessor otherwise probes are inserted in-between. This will add an branch to the predecessor.

What about adding a assertion to enforce this invariant?

		if (this.predecessor != null) {
			throw new AssertionError();
		}

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,129 @@
...
+		if (Opcodes.ATHROW != e.getOpcode()) {
+			return 0;
+		}
+		return size;
+	}
+
+	private static boolean isSame(final int size, AbstractInsnNode e,
+			AbstractInsnNode n) {
+		for (int i = 0; i < size; i++) {
+			if (e.getOpcode() != n.getOpcode()) {
```
marchof: What about also checking operands here? Should be the same except for labels.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,73 @@
...
+package org.jacoco.core.test.filter.targets;
+
+import static org.jacoco.core.test.validation.targets.Stubs.nop;
+
+/**
+ * This test target is a finally block.
+ */
+public class Finally {
+
+	private static void ex(boolean t) {
```
marchof: Move this to `org.jacoco.core.test.validation.targets.Stubs`?

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.filter.targets;
+
+import static org.jacoco.core.test.validation.targets.Stubs.nop;
+
+/**
+ * This test target is a finally block.
+ */
+public class Finally {
```
marchof: I wonder whether filter works for nested finally blocks too. These create really ugly flows.

Godin: @marchof Do you think about some particular example of an "ugly flow"? or just nested? in any case we can talk/test tomorrow at GenevaJUG 😉

marchof: Just nested: A try/finally block within a finally block. [This](https://github.com/marchof/will-this-compile/blob/master/will-this-compile/src/NestedFinally.java) is probably too much 😉 .

Also a test for finally blocks with some control structures (e.g. `if`) would be a nice test for the `isSame()` method.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
Godin: > allows internal optimizations later

I suppose internal in respect to a class and not project since class is already in internal package 😉 

Jokes aside - while indeed such encapsulation looks cleaner, this will complicate a test a bit which currently able to directly assert on a content of this `BitSet`, but ok can do this.

marchof: Of course internal to the internal class 😉 . For whitebox testing of `Instruction` we can still expose the `BitSet` but for interaction with other classes classes I would prefer a more semantic interface (like `merge()`).

Godin: > For whitebox testing of `Instruction` we can still expose the `BitSet`

I think that so far we don't have fields/methods that are "only for testing".

marchof: I think the only ones are in `LabelFlowAnalyzer`.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: In fact we already have finally of try-with-resources nested into finally of another try-with-resources and this new filter changed assertion there 😉 but indeed - can add explicit one here.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
Godin: @marchof turns out that not only for labels: for example for
```
class Fun {
  static void nop(String s) {
  }

  void fun() {
    try {
      nop("1");
    } finally {
      try {
        nop("2");
      } catch (Exception e) {
        nop("catch " + e);
      } finally {
        nop("finally");
      }
    }
  }
}
```

javac will generate code that uses `astore 1` and `aload 1` on non exceptional path, while `astore 4` and `aload 4` on exceptional path. This can be easily seen in `LocalVariableTable`, which we can't use during comparison:

```
      LocalVariableTable:
           19      22     1     e   Ljava/lang/Exception;
           76      23     4     e   Ljava/lang/Exception;
```

and now I recall that I already saw this during implementation of filter for `try-with-resources`. So let's have a talk about this tomorrow at GenevaJUG.

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
Godin: @marchof encapsulated, added `toString` instead of package-local field to avoid its accidental usage in other classes

marchof: @Godin Thanks, LGTM now! :+1:

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
Godin: @marchof I also was puzzled by this multiple times.

Maybe terms "direct predecessor" and "in-between" are a bit vague: in case of
```
L0: IFEQ L2
L1: NOP
L2: NOP
```
`L1` will not be marked as predecessor of `L2` and there is probe "in-between" them
`L2` will be marked as predecessor of `L0`, however not so "direct" (i.e. not a previous instruction) - via jump and there is an "in-between" probe that is executed on this jump from `L0` to `L2` 😉 

But your proposal anyway better than nothing.

I updated javadoc and added check of invariant using a slightly modified version of your proposal.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/InstructionTest.java
```diff
@@ -58,6 +59,13 @@ public void testSetPredecessor() {
 				new InsnNode(Opcodes.NOP), 122);
 		instruction.setPredecessor(predecessor, 0);
 		assertEquals(1, predecessor.getBranches());
+
+		try {
+			instruction.setPredecessor(predecessor, 0);
+			fail("exception expected");
+		} catch (IllegalStateException e) {
```
marchof: @Godin Why not `@Test(expected=IllegalStateException.class)`?

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
marchof: @Godin Thanks, hopefully helps us and others with this tricky piece of code in future.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/InstructionTest.java
Godin: @marchof to isolate operation that throws since there was already code in this test

marchof: @Godin Ah I see. But in this case I would prefer to have a separate test case specifically for this scenario.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: @marchof About control structures such as `if`: there is already test of conditional jump that is generated for `nop(i == 0)`, so I'm hesitating with addition of test for `if` statement - in this case why not add tests for other statements? and what about their relative order and so on? i.e. how to not go mad trying to generate all possible variants that are infinite in general.

Nested `finally` in his turn shows difference of slots for different duplicates of same `finally` block, which is not shown in existing tests here, so added it.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/InstructionTest.java
Godin: @marchof done

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
marchof: @Godin Ok, this is enough of control structures for now.

org.jacoco.core.test/src/org/jacoco/core/internal/flow/InstructionTest.java
marchof: @Godin 👍  Thanks!

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,132 @@
...
+
+	private static int size(AbstractInsnNode e) {
+		if (e.getOpcode() != Opcodes.ASTORE) {
+			return 0;
+		}
+		final int var = ((VarInsnNode) e).var;
+		int size = 0;
+		while (true) {
+			e = next(e);
+			if (e.getOpcode() == Opcodes.ALOAD
```
Godin: Found NPE here during analysis of `javax/swing/JEditorPane$PageLoader#doInBackground()` from `rt.jar` of JDK 7u80 in case when `finally` on exceptional path doesn't throw stored exception - for example:
```
try {
  ...
} finally {
  return ...;
}
```

org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/TryWithResourcesTest.java
```diff
@@ -62,8 +62,7 @@ public void test2() {
 		// without filter next line has branches:
 		assertLine("test2.close", ICounter.EMPTY);
 		assertLine("test2.catch", ICounter.NOT_COVERED);
-		assertLine("test2.finally", ICounter.PARTLY_COVERED);
-		assertLine("test2.after", ICounter.FULLY_COVERED);
```
Godin: Part of filter for ECJ should not merge this line with finally block.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: @marchof done

org.jacoco.core/src/org/jacoco/core/internal/flow/Instruction.java
```diff
@@ -25,10 +27,13 @@
 
 	private int branches;
 
-	private int coveredBranches;
+	// TODO in majority of cases BitSet not needed, needed only for big switches
```
Godin: @marchof I'm tempted to simply remove this TODO-comment. WDTY?

marchof: @Godin Agreed. We can do some performance testing later and see where the actual bottlenecks are.

marchof:
@Godin I did a quick look with JProfiler: The BitSet does not show up. Actually our hotspot seems to be the   `nodeToInstruction` Map. Putting entries consumes 44% in total. No idea yet why.

image

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,143 @@
...
+			AbstractInsnNode e, AbstractInsnNode n) {
+		output.ignore(e, e);
+		e = next(e);
+		for (int i = 0; i < size; i++) {
+			output.merge(e, n);
+			n = next(n);
+			e = next(e);
+		}
+		output.ignore(e, next(e));
+		if (n.getOpcode() == Opcodes.GOTO) {
```
Godin: Found NPE here during analysis of `org/apache/hadoop/hadoop-mapred-instrumented/0.22.0` , source code that seems to be compiled with ECJ looks like

```
while (true) {
  try {
    nop();
  } finally {
    synchronized (this) {
      nop();
    }
  }
}
```

so that bytecode contains duplicate of a finally handler as a last sequence of instructions.

marchof:
@Godin FYI: I ran the analyzer on a 1mio LOC project fully compiled with ECJ without exceptions. :white_check_mark: 

I also might have the chance to run it on a 3 TB internally used Nexus repository.

Godin:
@marchof :yum: I suppose that you verified only the fact that there is no exceptions? Verification that  we don't merge unrelated lines is easy and doesn't require parsing. I've just completed this check for same ≈20 Gb of JAR files with classes from Maven central as for filter for `try-with-resources`. Will try to do smarter check involving sources.

marchof:
@Godin Correct, exceptions only! Will do the same for the **big** repo for now.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,143 @@
...
+			AbstractInsnNode e, AbstractInsnNode n) {
+		output.ignore(e, e);
+		e = next(e);
+		for (int i = 0; i < size; i++) {
+			output.merge(e, n);
+			n = next(n);
+			e = next(e);
+		}
+		output.ignore(e, next(e));
+		if (n != null && n.getOpcode() == Opcodes.GOTO) {
```
Godin: Most likely it is not safe to ignore `GOTO` here, because `break` and `continue` statements are compiled into it.

marchof:
@Godin Scanning a larger repo (Thanks @p45q!) found the following exception:

	java.lang.NullPointerException
	at org.jacoco.core.internal.analysis.filter.FinallyFilter.isSame(FinallyFilter.java:93)
	at org.jacoco.core.internal.analysis.filter.FinallyFilter.filter(FinallyFilter.java:45)
	at org.jacoco.core.internal.analysis.filter.FinallyFilter.filter(FinallyFilter.java:30)
	at org.jacoco.core.internal.analysis.MethodAnalyzer.accept(MethodAnalyzer.java:130)
	at org.jacoco.core.internal.flow.ClassProbesAdapter$2.visitEnd(ClassProbesAdapter.java:89)
	at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1036)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:708)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:521)
	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:111)
	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:127)

It can be reproduced with the following artifact:

    <dependency>
        <groupId>colt</groupId>
        <artifactId>colt</artifactId>
        <version>1.0.3</version>
    </dependency>

Godin:
@marchof @p45q Thanks to both of you! Actually I was thinking about possibility of NPE in `isSame` method, but wasn't able to imagine pair of real source code and bytecode that would cause it, so really interesting to see what's in `colt:colt:1.0.3`.

marchof:
@Godin BTW, the command line tool was helpful to run this test in a docker container directly on the repo server :wink: (I have a local branch rebased on your latest work).

marchof:
@Godin Re colt: The corresponding source îs probably [here](https://dst.lbl.gov/ACSSoftware/colt/colt-download/releases/colt-1.0.3.zip).

Godin:
@marchof Thanks for the link. What I see so far based only on class file: it has version 45 and `ret` instructions, so that `JSRInlinerAdapter` kicks in before filter. And yeah - I felt the lack of CLI for testing purposes.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
Godin: :laughing: in presence of `finally` javac does not assign line number to `GOTO` instruction, when `break` statement refers to a loop outside of `try`, while ECJ assigns as was expected:

```
void fun() {
  while (t())
    try {
      if (t())
        break;
    } finally {
      t();
    }
}
```

Godin:
@marchof While fix for NPE in this case is trivial, this raises question about result in case of older compiler versions - in this particular case we'll have false-negative as nothing will be merged by filter.

While JDK 5 can produce bytecode of older versions, this is not the same as running older compiler version - in particular as far as I can see JDK 5 doesn't produce `jsr` and `ret` instructions. And nowadays not trivial to run older versions, what heavily complicates investigations of their bytecode patterns.

All in all -  I think that we shouldn't work on improvements of result of filter for older compiler versions, but only fix NPE. WDYT?

marchof:
@Godin I wouldn't consider or test functional correctness for compiler versions < 1.5. But our code should not throw exceptions on any valid class file supplied.

As a side note I think all matching should happen with well-unit-tested matching primitives (like provided by `AbstractMatcher`) which by their API design cannot throw NPEs in the specific filter implementations.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
Godin: :disappointed: however in case of ECJ this example demonstrates that we miss merge of instructions that are at the exit of a loop without `break`. And for slightly modified example same happens with javac, but not with ECJ:

```
void fun() {
  while (t())
    try {
      if (t())
	break;
      nop();
    } finally {
      nop();
    }
}
```

marchof:
@Godin Another finding on the "big repo" by @p45q:

	java.lang.IllegalStateException
	at org.jacoco.core.internal.flow.Instruction.setPredecessor(Instruction.java:81)
	at org.jacoco.core.internal.analysis.MethodAnalyzer.visitEnd(MethodAnalyzer.java:371)
	at org.jacoco.core.internal.analysis.ClassAnalyzer$1.visitEnd(ClassAnalyzer.java:71)
	at org.objectweb.asm.MethodVisitor.visitEnd(MethodVisitor.java:878)
	at org.jacoco.core.internal.analysis.MethodAnalyzer.accept(MethodAnalyzer.java:142)
	at org.jacoco.core.internal.flow.ClassProbesAdapter$2.visitEnd(ClassProbesAdapter.java:89)
	at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1036)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:708)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:521)
	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:111)
	at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:127)
	... 5 more

Happens on a Eclipse EMF class of unknown version (packaged in some IBM product): [ResourceImpl.zip](https://github.com/jacoco/jacoco/files/992940/ResourceImpl.zip)

Godin:
@marchof wow, thanks! However this one violates contract that was already here for method `Instruction.setPredecessor`,  in this PR we just added check of this contract. Anyway I'll have a look.

marchof:
@Godin That's why I was also considering opening a separate issue. If this is a bigger think we should move the assertion and fix into a separate PR.

Godin:
@marchof here is what I've discovered for now:
This class also contains `JSR` and `RET` instructions, so that `JSRInlinerAdapter` kicks in before analysis. It adds labels, so that same instruction has multiple labels - comment about that was added in commit bead2e4d671b2ec3bc5aca73c79830f5313f6c95 to `MethodAnalyzer`:
> Due to ASM issue #315745 there can be more than one label per instruction

As a very simplified example of code after `JSRInlinerAdapter` for this class, which triggers `IllegalStateException`:
```
  ALOAD 1
  IFNE L1
  GOTO L2
 L1
 L2
  RETURN
```

And seems that in such case there is only 1 probe, while expected 3 as in case with single label.
So I have feeling that problem comes from `LabelFlowAnalyzer` that wasn't updated in commit bead2e4d671b2ec3bc5aca73c79830f5313f6c95.

Godin:
@marchof as was decided by phone: removed verification of contract of `Instruction.setPredecessor` and created separate ticket - #533

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,286 @@
...
+				nop(); // $line-breakStatement.finally$
+			}
+		}
+	}
+
+	/**
+	 * Note that difference of this case from {@link #breakStatement()} is that
+	 * there is nothing between <code>break</code> statement and end of
+	 * <code>try</code> block. And current implementation of filter does not
+	 * handle such case.
```
Godin: Filter that was implemented in #604 handles this case.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,286 @@
...
+		} catch (Exception e) { // $line-emptyCatch.catch$
+		} finally { // $line-emptyCatch.finally$
+			nop(); // $line-emptyCatch.finallyBlock$
+		} // $line-emptyCatch.finallyBlockEnd$
+	}
+
+	/**
+	 * Note that javac generates <code>goto</code> instruction at the end of
+	 * <code>while</code> statement that refers to a line of previous
+	 * instruction. And so causes partial coverage of last line of finally
```
Godin: Seems we forgot to preserve this case in #604.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,286 @@
...
+		try {
+			return null;
+		} finally {
+			nop(); // $line-returnInBody$
+		}
+	}
+
+	/**
+	 * Note that in this case javac generates bytecode that uses different slots
+	 * in different duplicates of same finally block for variable that stores
```
Godin: Seems that we forgot to preserve this or at least similar comment in #604.

Issue #506, opened 2017-03-28
Create separate modules for core integration tests


marchof:
The "validation" tests from the `org.jacoco.core.test` project should be moved to a new project `org.jacoco.core.it`.

- These tests are not unit tests
- They require compiler version specific setups/filtering
- They should not account for code coverage

Bonus:

- We could run these tests with the [ECJ compiler](http://stackoverflow.com/questions/33164976/using-eclipse-java-compiler-ecj-in-maven-builds).

marchof:
Implemented in #711.

PR #522, opened 2017-04-23
Add profile for compilation with ECJ

Godin:


Godin:
@marchof maybe its a bit of an overkill to run full build with ECJ, but leaving this aside as well as extraction of core integration tests (#506) - such automatic validation is anyway better than no validation, especially that strictly speaking not guaranteed that Eclipse IDE is used for contributions. So could you please review?

PR #525, opened 2017-04-28
Simple Command Line Interface for JaCoCo

marchof:
A feature often asked for is a command line interface for JaCoCo. We have a pull request (#86) since quite some time, unfortunately without any tests and documentation. This is a new approach based on some ideas of #86.

**To clearly manage expectations: The scope is about simple use cases only! For complex scenarios we have plugins to all common build systems.**

## Open Issues

- [x] Clarify and test usage of multi value options (e.g. `-classfiles`)
- [x] Improve documentation layout (long usage lines)
- [x] Test class `XmlDocumenation` or remove from test scope
- [x] Clarify class file version issue in build with exec-maven-plugin:1.6.0
- [x] Clarify why build result gets too big when compiled with Java 9

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -94,12 +94,12 @@
     <module>../org.jacoco.agent.rt</module>
     <module>../org.jacoco.agent</module>
     <module>../org.jacoco.ant</module>
-
+    <module>../org.jacoco.cli</module>
+    <module>../org.jacoco.examples</module>
     <module>../jacoco-maven-plugin</module>
 
     <module>../org.jacoco.tests</module>
 
-    <module>../org.jacoco.examples</module>
     <module>../org.jacoco.doc</module>
     <module>../jacoco</module>
   </modules>
@@ -143,6 +143,7 @@
     <!-- Dependencies versions -->
     <asm.version>5.2</asm.version>
     <ant.version>1.7.1</ant.version>
+    <args4j.version>2.0.28</args4j.version>
     <junit.version>4.8.2</junit.version>
 
     <!-- ================== -->
@@ -198,6 +199,11 @@
       </dependency>
       <dependency>
         <groupId>${project.groupId}</groupId>
+        <artifactId>org.jacoco.cli</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
         <artifactId>org.jacoco.examples</artifactId>
         <version>${project.version}</version>
       </dependency>
@@ -221,6 +227,11 @@
         <groupId>org.apache.ant</groupId>
         <artifactId>ant-antunit</artifactId>
         <version>1.2</version>
+      </dependency>
+      <dependency>
+        <groupId>args4j</groupId>
+        <artifactId>args4j</artifactId>
+        <version>${args4j.version}</version>
       </dependency>
       <dependency>
         <groupId>junit</groupId>
@@ -393,6 +404,11 @@
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>xml-maven-plugin</artifactId>
           <version>1.0</version>
+        </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>exec-maven-plugin</artifactId>
+          <version>1.6.0</version>
         </plugin>
         <!-- Third-party plugins -->
         <plugin>
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -408,7 +408,8 @@
         <plugin>
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>exec-maven-plugin</artifactId>
-          <version>1.6.0</version>
+          <!--  latest version which runs with Java 5 -->
+          <version>1.5.0</version>
         </plugin>
         <!-- Third-party plugins -->
         <plugin>
--- a/org.jacoco.cli.test/pom.xml
+++ b/org.jacoco.cli.test/pom.xml
@@ -31,16 +31,6 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>org.jacoco.cli</artifactId>
-      <!-- 
-      <version>${project.version}</version> 
-      <classifier>nodeps</classifier>
-      <exclusions>
-        <exclusion>
-          <groupId>*</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-      </exclusions>
-       -->
     </dependency>
     <dependency>
       <groupId>junit</groupId>
--- a/org.jacoco.cli.test/pom.xml
+++ b/org.jacoco.cli.test/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+   All rights reserved. This program and the accompanying materials
+   are made available under the terms of the Eclipse Public License v1.0
+   which accompanies this distribution, and is available at
+   http://www.eclipse.org/legal/epl-v10.html
+
+   Contributors:
+      Evgeny Mandrikov - initial API and implementation
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jacoco</groupId>
+    <artifactId>org.jacoco.tests</artifactId>
+    <version>0.7.10-SNAPSHOT</version>
+    <relativePath>../org.jacoco.tests</relativePath>
+  </parent>
+
+  <artifactId>org.jacoco.cli.test</artifactId>
+
+  <name>JaCoCo :: Test :: Command Line Interface</name>
+
+  <properties>
+    <jacoco.includes>org.jacoco.cli.*</jacoco.includes>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>org.jacoco.cli</artifactId>
+      <!-- 
+      <version>${project.version}</version> 
+      <classifier>nodeps</classifier>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+       -->
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
@@ -25,7 +25,8 @@ public class MainTest extends CommandTestBase {
 		assertFailure();
 		assertNoOutput(out);
 		assertContains("Argument \"<command>\" is required", err);
-		assertContains("Usage: java -jar jacococli.jar -help | <command>", err);
+		assertContains("Usage: java -jar jacococli.jar --help | <command>",
+				err);
 		assertContains("Command line interface for JaCoCo.", err);
 	}
 
@@ -37,24 +38,26 @@ public class MainTest extends CommandTestBase {
 		assertFailure();
 		assertNoOutput(out);
 		assertContains("\"foo\" is not a valid value for \"<command>\"", err);
-		assertContains("Usage: java -jar jacococli.jar -help | <command>", err);
+		assertContains("Usage: java -jar jacococli.jar --help | <command>",
+				err);
 	}
 
 	@Test
 	public void should_print_general_usage_when_help_option_is_given()
 			throws Exception {
-		execute("-help");
+		execute("--help");
 
 		assertOk();
 		assertNoOutput(err);
-		assertContains("Usage: java -jar jacococli.jar -help | <command>", out);
+		assertContains("Usage: java -jar jacococli.jar --help | <command>",
+				out);
 		assertContains("<command> : dump|instrument|merge|report", out);
 	}
 
 	@Test
 	public void should_print_command_usage_when_command_and_help_option_is_given()
 			throws Exception {
-		execute("dump", "-help");
+		execute("dump", "--help");
 
 		assertOk();
 		assertNoOutput(err);
@@ -67,7 +70,7 @@ public class MainTest extends CommandTestBase {
 	@Test
 	public void should_not_print_any_output_when_quiet_option_is_given()
 			throws Exception {
-		execute("version", "-quiet");
+		execute("version", "--quiet");
 
 		assertOk();
 		assertNoOutput(out);
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
@@ -64,4 +64,14 @@ public class MainTest extends CommandTestBase {
 				out);
 	}
 
+	@Test
+	public void should_not_print_any_output_when_quiet_option_is_given()
+			throws Exception {
+		execute("version", "-quiet");
+
+		assertOk();
+		assertNoOutput(out);
+		assertNoOutput(err);
+	}
+
 }
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
@@ -19,7 +19,7 @@ import org.junit.Test;
 public class MainTest extends CommandTestBase {
 
 	@Test
-	public void shouldPrintUsage_whenNoArgumentsGiven() throws Exception {
+	public void should_print_usage_when_no_arguments_given() throws Exception {
 		execute();
 
 		assertFailure();
@@ -30,7 +30,7 @@ public class MainTest extends CommandTestBase {
 	}
 
 	@Test
-	public void shouldPrintErrorMessage_whenInvalidCommandIsGiven()
+	public void should_print_error_message_when_invalid_command_is_given()
 			throws Exception {
 		execute("foo");
 
@@ -41,7 +41,7 @@ public class MainTest extends CommandTestBase {
 	}
 
 	@Test
-	public void shouldPrintGeneralHelp_whenHelpOptionIsGiven()
+	public void should_print_general_usage_when_help_option_is_given()
 			throws Exception {
 		execute("-help");
 
@@ -52,7 +52,7 @@ public class MainTest extends CommandTestBase {
 	}
 
 	@Test
-	public void shouldPrintCommandHelp_whenHelpOptionIsGiven()
+	public void should_print_command_usage_when_command_and_help_option_is_given()
 			throws Exception {
 		execute("dump", "-help");
 
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Marc R. Hoffmann - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.cli.internal;
+
+import org.junit.Test;
+
+/**
+ * Unit tests for {@link Main}.
+ */
+public class MainTest extends CommandTestBase {
+
+	@Test
+	public void shouldPrintUsage_whenNoArgumentsGiven() throws Exception {
+		execute();
+
+		assertFailure();
+		assertNoOutput(out);
+		assertContains("Argument \"<command>\" is required", err);
+		assertContains("Usage: java -jar jacococli.jar -help | <command>", err);
+		assertContains("Command line interface for JaCoCo.", err);
+	}
+
+	@Test
+	public void shouldPrintErrorMessage_whenInvalidCommandIsGiven()
+			throws Exception {
+		execute("foo");
+
+		assertFailure();
+		assertNoOutput(out);
+		assertContains("\"foo\" is not a valid value for \"<command>\"", err);
+		assertContains("Usage: java -jar jacococli.jar -help | <command>", err);
+	}
+
+	@Test
+	public void shouldPrintGeneralHelp_whenHelpOptionIsGiven()
+			throws Exception {
+		execute("-help");
+
+		assertOk();
+		assertNoOutput(err);
+		assertContains("Usage: java -jar jacococli.jar -help | <command>", out);
+		assertContains("<command> : dump|instrument|merge|report", out);
+	}
+
+	@Test
+	public void shouldPrintCommandHelp_whenHelpOptionIsGiven()
+			throws Exception {
+		execute("dump", "-help");
+
+		assertOk();
+		assertNoOutput(err);
+		assertContains("Usage: java -jar jacococli.jar dump", out);
+		assertContains(
+				"Request execution data from a JaCoCo agent running in 'tcpserver' output mode.",
+				out);
+	}
+
+}
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
@@ -117,4 +117,26 @@ public class ReportTest extends CommandTestBase {
 						.isFile());
 	}
 
+	@Test
+	public void should_use_all_values_when_multiple_classfiles_options_are_provided()
+			throws Exception {
+		File html = new File(tmp.getRoot(), "coverage");
+
+		final String c1 = getClassPath()
+				+ "/org/jacoco/cli/internal/commands/ReportTest.class";
+		final String c2 = getClassPath()
+				+ "/org/jacoco/cli/internal/commands/DumpTest.class";
+
+		execute("report", "-classfiles", c1, "-classfiles", c2, "-html",
+				html.getAbsolutePath());
+
+		assertOk();
+		assertTrue(html.isDirectory());
+		assertTrue(new File(html,
+				"org.jacoco.cli.internal.commands/ReportTest.html").isFile());
+		assertTrue(
+				new File(html, "org.jacoco.cli.internal.commands/DumpTest.html")
+						.isFile());
+	}
+
 }
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
@@ -1,0 +1,118 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Marc R. Hoffmann - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.cli.internal.commands;
+
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.io.FileOutputStream;
+
+import org.jacoco.cli.internal.CommandTestBase;
+import org.jacoco.core.data.ExecutionData;
+import org.jacoco.core.data.ExecutionDataWriter;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/**
+ * Unit tests for {@link Report}.
+ */
+public class ReportTest extends CommandTestBase {
+
+	@Rule
+	public TemporaryFolder tmp = new TemporaryFolder();
+
+	@Test
+	public void shouldPrintUsage_whenNoArgumentsGiven() throws Exception {
+		execute("report");
+
+		assertFailure();
+		assertContains("Option \"-classfiles\" is required", err);
+		assertContains(
+				"Usage: java -jar jacococli.jar report [<execfiles> ...]", err);
+	}
+
+	@Test
+	public void shouldPrintNumberOfAnalyzedClasses() throws Exception {
+		execute("report", "-classfiles", getClassPath());
+
+		assertOk();
+		assertContains("[INFO] Writing report with 11 classes.", out);
+	}
+
+	@Test
+	public void shouldPrintWarning_whenExecDataDoesNotMatch() throws Exception {
+		File exec = new File(tmp.getRoot(), "jacoco.exec");
+		final FileOutputStream execout = new FileOutputStream(exec);
+		ExecutionDataWriter writer = new ExecutionDataWriter(execout);
+		// Add probably invalid id for this test class:
+		writer.visitClassExecution(
+				new ExecutionData(0x123, getClass().getName().replace('.', '/'),
+						new boolean[] { true }));
+		execout.close();
+
+		execute("report", exec.getAbsolutePath(), "-classfiles",
+				getClassPath());
+
+		assertOk();
+		assertContains("[WARN] Some classes do not match with execution data.",
+				out);
+		assertContains(
+				"[WARN] For report generation the same class files must be used as at runtime.",
+				out);
+		assertContains(
+				"[WARN] Execution data for class org/jacoco/cli/internal/commands/ReportTest does not match.",
+				out);
+	}
+
+	@Test
+	public void shouldCreateXmlReport_whenXmlOptionIsProvided()
+			throws Exception {
+		File xml = new File(tmp.getRoot(), "coverage.xml");
+
+		execute("report", "-classfiles", getClassPath(), "-xml",
+				xml.getAbsolutePath());
+
+		assertOk();
+		assertTrue(xml.isFile());
+	}
+
+	@Test
+	public void shouldCreateCsvReport_whenXmlOptionIsProvided()
+			throws Exception {
+		File csv = new File(tmp.getRoot(), "coverage.csv");
+
+		execute("report", "-classfiles", getClassPath(), "-csv",
+				csv.getAbsolutePath());
+
+		assertOk();
+		assertTrue(csv.isFile());
+	}
+
+	@Test
+	public void shouldCreateHtmlReport_whenHtmlOptionIsProvided()
+			throws Exception {
+		File html = new File(tmp.getRoot(), "coverage");
+
+		execute("report", "-classfiles", getClassPath(), "-sourcefiles",
+				"./src", "-html", html.getAbsolutePath());
+
+		assertOk();
+		assertTrue(html.isDirectory());
+		assertTrue(new File(html,
+				"org.jacoco.cli.internal.commands/ReportTest.html").isFile());
+		assertTrue(new File(html,
+				"org.jacoco.cli.internal.commands/ReportTest.java.html")
+						.isFile());
+	}
+
+}
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
@@ -87,7 +87,7 @@ public class ReportTest extends CommandTestBase {
 	}
 
 	@Test
-	public void shouldCreateCsvReport_whenXmlOptionIsProvided()
+	public void shouldCreateCsvReport_whenCsvOptionIsProvided()
 			throws Exception {
 		File csv = new File(tmp.getRoot(), "coverage.csv");
 
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
@@ -32,7 +32,8 @@ public class ReportTest extends CommandTestBase {
 	public TemporaryFolder tmp = new TemporaryFolder();
 
 	@Test
-	public void shouldPrintUsage_whenNoArgumentsGiven() throws Exception {
+	public void should_print_usage_when_no_options_are_given()
+			throws Exception {
 		execute("report");
 
 		assertFailure();
@@ -42,7 +43,7 @@ public class ReportTest extends CommandTestBase {
 	}
 
 	@Test
-	public void shouldPrintNumberOfAnalyzedClasses() throws Exception {
+	public void should_print_number_of_analyzed_classes() throws Exception {
 		execute("report", "-classfiles", getClassPath());
 
 		assertOk();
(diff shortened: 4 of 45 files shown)


marchof:
@mathjeff Can you please take a look and check whether your scenario is covered by this new PR? What is missing?

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+		execute("report", "-classfiles", getClassPath(), "-xml",
+				xml.getAbsolutePath());
+
+		assertOk();
+		assertTrue(xml.isFile());
+	}
+
+	@Test
+	public void shouldCreateCsvReport_whenXmlOptionIsProvided()
```
mathjeff: "Xml" -> "Csv"

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
```diff
@@ -0,0 +1,96 @@
...
+import org.jacoco.core.runtime.OfflineInstrumentationAccessGenerator;
+import org.kohsuke.args4j.Argument;
+import org.kohsuke.args4j.Option;
+
+/**
+ * The <code>instrument</code> command.
+ */
+public class Instrument extends Command {
+
+	@Option(name = "-destdir", usage = "directory to write instrumented Java classes to", metaVar = "<dir>", required = true)
```
mathjeff: Could we remove the 'dir' from '-destdir', possibly calling it '-destpath', to clarify the case where the input files is a .jar file and the value of '-destdir' is a .jar file?

org.jacoco.cli/src/org/jacoco/cli/internal/Main.java
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		if (help) {
+			printHelp(out);
+			return 0;
+		} else if (command.help) {
+			command.printHelp(out);
+			return 0;
+		} else {
+			return command.execute(out, err);
```
mathjeff: We'll want to flush out and err somewhere, probably here; I was surprised that "[INFO] %s classes instrumented to %s.%n" wasn't getting displayed when I tested this. Of course, it's more convenient for Android if the output is empty, although I'm thinking that either we'll add a '-quiet' option to Jacoco or we'll redirect the output within Android.

org.jacoco.cli.test/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+      <!-- 
+      <version>${project.version}</version> 
+      <classifier>nodeps</classifier>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+       -->
```
mathjeff: Is this intentional?

marchof:
@mathjeff Hi, thanks for the detailled review! I will adress the issues you observed.

Regarding filtering: I would prefer to handle this in a separate PR as it should then be consistently implemented for the other integrations. Also the core API should be improved for this.

org.jacoco.cli.test/pom.xml
marchof: @mathjeff I tried to directly test the "shaded" version. Unfortunatelly this does not work within the m2e IDE. Should be be removed of course.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
marchof: @mathjeff Correct, `-dest` is the way to go.

org.jacoco.cli/src/org/jacoco/cli/internal/Main.java
marchof: @mathjeff Not sure what the problem here is: in `Main.main()` the `PrintWriter`s are created with auto-flushing after every line. How do you reproduce this?

org.jacoco.cli.test/pom.xml
marchof: @mathjeff Fixed.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
marchof: @mathjeff Fixed.

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
marchof: @mathjeff Fixed.

org.jacoco.build/pom.xml
```diff
@@ -394,6 +405,11 @@
           <artifactId>xml-maven-plugin</artifactId>
           <version>1.0</version>
         </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>exec-maven-plugin</artifactId>
+          <version>1.6.0</version>
```
Godin: @marchof so far our lower bound of Maven version for build of project - is 3.0, which in his turn has Java 5 as lower bound, while exec-maven-plugin version 1.6.0 requires Java 7. This explains failure in Travis where we use JDK 6 to run Maven.

org.jacoco.cli/src/org/jacoco/cli/internal/Main.java
mathjeff: Yeah I was surprised too. The simplest example I have is to ask jacoco to instrument itself:

/workspace/jacoco/org.jacoco.cli (work)$ java -jar target/org.jacoco.cli-0.7.10-SNAPSHOT-nodeps.jar instrument target/org.jacoco.cli-0.7.10-SNAPSHOT.jar -destdir /tmp/dummy.jar

, which doesn't give any output when I run it.

Godin: @marchof actually as far as I can see currently `main` does `new PrintWriter(System.out)` and [JavaDoc says](https://docs.oracle.com/javase/7/docs/api/java/io/PrintWriter.html#PrintWriter(java.io.OutputStream)) :
> Creates a new PrintWriter, **without automatic line flushing**

org.jacoco.build/pom.xml
marchof: @Godin Ok, thanks for the Info! We can execute the documentation as part of the test. Or is that too much of a hack?

Godin: @marchof why not simply use 1.5.0 version of exec-maven-plugin that requires Java 5?

marchof: @Godin :white_check_mark: Done.

org.jacoco.cli/src/org/jacoco/cli/internal/Main.java
marchof: @mathjeff @Godin May bad -- missread JavaDoc. Fixed :white_check_mark:

marchof: @mathjeff `-quiet` option is now in place.

org.jacoco.doc/docroot/index.html
```diff
@@ -68,6 +68,12 @@
       <td>Ant <i>(all other dependencies included)</i></td>
     </tr>
     <tr>
+      <td><span class="el_jar">jacococli.jar</span></td>
+      <td>no</td>
+      <td>JaCoCo Ant tasks</td>
```
Godin: @marchof copy-paste detected :wink:

org.jacoco.cli/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
+/bin
```
Godin: @marchof I don't think this is needed - we have consolidated `.gitignore`

org.jacoco.doc/docroot/index.html
marchof: @Godin Thanks, fixed!

org.jacoco.cli/.gitignore
marchof: @Godin Thanks, fixed! (started this branch before we consolidates `.gitignore`)

marchof:
Resulting jacoco-xxx.zip files for different build:

| ZIP size  | master     | cli        |
|-----------|------------|------------|
| Java 5    | 3,035,510  | 3,523,558  |
| Java 9    | 3,462,577  | 3,956,019  |

Looks like we have constant overheads which sums up.

**master -> cli:  475 KB**

This is primarly due to a new all-in JAR and bigger coverage and test report.

**Java 5 -> Java 9: 417 KB**

The reason here is that the JavaDoc grows from 2,8MB to 4,1MB, probably as JavaDoc created with Java 9 comes with new features and bundles JQuery.  

For Java 9 and cli both effects add-up beyond our current check limit.

As this is a pure sanity check I think we can safely lift the upper limit a bit.

marchof:
@Godin @mathjeff From my point of view this feature is ready to merge. If you find some time please double check and approve the review if you agree. Thx!

mathjeff:
Hey thanks marchof@ for spending the time on this. Based on reading the code and trying out the resultant jar, this looks good to me. I'll leave the final approval/disapproval up to Godin@.

Android won't be able to directly use this yet until filtering by classname is implemented in a subsequent pull request, but this pull request looks consistent and reasonable to me.

Thanks!

marchof:
@mathjeff Thanks for your feedback!

mathjeff:
Godin@ , do you have any updated opinions about this?

It occurred to me that it'd be convenient for me if this gets merged somewhat soon because of the pull request that I'm building on top if it that adds filtering ( https://github.com/jacoco/jacoco/pull/537 ).

I'm not sure whether it's reasonable for me to approve this pull request given that I'm not part of the Jacoco project, but I've looked at the code and tried it out and it seems good to me. If you think it's reasonable for me to approve it then I'll do that.

Thanks

marchof:
@mathjeff The review is on @Godin's TODO list. But please understand that preparing EclEmma for the Eclipse Oxygen release is currently out top priority. Thx!

Godin:
@mathjeff @marchof excuse me - I was quite busy this week, will do my best to have a look this weekend. However indeed as @marchof said - preparation of EclEmma for the Eclipse Oxygen is in top priority.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Merge.java
```diff
@@ -0,0 +1,62 @@
...
+import org.kohsuke.args4j.Argument;
+import org.kohsuke.args4j.Option;
+
+/**
+ * The <code>merge</code> command.
+ */
+public class Merge extends Command {
+
+	@Argument(usage = "list of JaCoCo *.exec files to read", metaVar = "<execfiles>")
+	List<File> execfiles = new ArrayList<File>();
```
Godin: @marchof when list is empty

```
java -jar jacococli.jar merge -destfile /tmp/jacoco.exec
```

creates an empty exec file, wondering if this is intentional?

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
```diff
@@ -0,0 +1,142 @@
...
+		assertContains(
+				"Usage: java -jar jacococli.jar report [<execfiles> ...]", err);
+	}
+
+	@Test
+	public void should_print_number_of_analyzed_classes() throws Exception {
+		execute("report", "-classfiles", getClassPath());
+
+		assertOk();
+		assertContains("[INFO] Writing report with 14 classes.", out);
```
Godin: @marchof message is a bit misleading as nothing has been actually written on disk

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
```diff
@@ -0,0 +1,170 @@
...
+	@Option(name = "-sourcefiles", usage = "location of the source files", metaVar = "<path>")
+	List<File> sourcefiles = new ArrayList<File>();
+
+	@Option(name = "-tabwith", usage = "tab stop width for the source pages (default 4)", metaVar = "<n>")
+	int tabwidth = 4;
+
+	@Option(name = "-name", usage = "name used for this report", metaVar = "<name>")
+	String name = "JaCoCo Coverage Report";
+
+	@Option(name = "-encoding", usage = "source file encoding (default platform encoding)", metaVar = "<charset>")
```
Godin: @marchof Maybe `by default platform encoding is used` or `when not specified - platform encoding is used` ?

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
```diff
@@ -0,0 +1,170 @@
...
+import org.kohsuke.args4j.Argument;
+import org.kohsuke.args4j.Option;
+
+/**
+ * The <code>report</code> command.
+ */
+public class Report extends Command {
+
+	@Argument(usage = "list of JaCoCo *.exec files to read", metaVar = "<execfiles>")
+	List<File> execfiles = new ArrayList<File>();
```
Godin: @marchof I have a fear that an empty list will be a common source of mistakes that lead to a report that shows zeros. However at least in case of HTML report we can improve later the report itself for such situations, but unfortunately not other formats. So wondering if we should print at least a warning? However maybe will be better to ban this because of existence of `quiet` option? Especially that there is command `classinfo` that just runs analysis of class files without exec files.

BTW seems that in this case not possible to navigate lower than method level or switch view to sources even when they are provided, but this is probably something unrelated to this PR:
```
java -jar /tmp/jacococli.jar report -classfiles /tmp/jacococli.jar -sourcefiles src -html /tmp/report
```

Also maybe a separate question, but I'm wondering about correctness of a behavior when destination already exists?

org.jacoco.doc/xsl/cli.xsl
```diff
@@ -0,0 +1,113 @@
...
+				<div id="content">
+				
+					<h1>Command Line Interface</h1>
+					
+					<p>
+					  JaCoCo comes with a command line interface to perform
+					  basic operations from the command line. The command line
+					  tools with all dependencies are packaged in
+					  <code>jacococli.jar</code> and are available with the
+					  JaCoCo download. A Java VM with version 1.5 or greater is
```
Godin: @marchof `A Java VM version 1.5 or greater is required for execution.` i.e. without word `with`?

org.jacoco.core/src/org/jacoco/core/tools/ExecDumpClient.java
```diff
@@ -123,7 +123,10 @@ public ExecFileLoader dump(final InetAddress address, final int port)
 					.setExecutionDataVisitor(loader.getExecutionDataStore());
 
 			remoteWriter.visitDumpCommand(dump, reset);
-			remoteReader.read();
+
+			if (!remoteReader.read()) {
+				throw new IOException("Socket closed unexpectedly.");
```
Godin: @marchof seems that this change is a bit unrelated to PR and test is missing for this case

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Dump.java
```diff
@@ -0,0 +1,77 @@
...
+	@Option(name = "-port", usage = "the port to connect to (default 6300)", metaVar = "<port>")
+	int port = AgentOptions.DEFAULT_PORT;
+
+	@Option(name = "-destfile", usage = "file to write execution data to", metaVar = "<path>", required = true)
+	File destfile;
+
+	@Option(name = "-reset", usage = "reset execution data on test target after dump")
+	boolean reset = false;
+
+	@Option(name = "-retry", usage = "number of retries (default 0)", metaVar = "<count>")
```
Godin: @marchof default in Ant task and Maven mojo is 10.

org.jacoco.doc/docroot/index.html
```diff
@@ -68,6 +68,12 @@
       <td>Ant <i>(all other dependencies included)</i></td>
     </tr>
     <tr>
+      <td><span class="el_jar">jacococli.jar</span></td>
+      <td>no</td>
+      <td>JaCoCo Command Line Interface</td>
```
Godin: @marchof at the bottom of this page we refer to `repo.html` that lists all deployed artifacts and their coordinates, this artifact will be deployed, but `repo.html` wasn't updated to mention its coordinates.

org.jacoco.cli/about.html
```diff
@@ -0,0 +1,40 @@
...
+<h2>About This Content</h2>
+
+<p>
+  @build.date@
+</p>
+
+<h3>License</h3>
+
+<p>
+  All Content in this plug-in is made available by Mountainminds GmbH &amp; Co.
```
Godin: @marchof this is nor OSGi bundle, nor Eclipse plug-in, nor any other plugin 😉  however we also have such file for `org.jacoco.agent.rt` but not for `org.jacoco.maven-plugin`

org.jacoco.cli/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>nodeps</shadedClassifierName>
+              <minimizeJar>true</minimizeJar>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                  <manifestEntries>
+                    <Main-Class>org.jacoco.cli.internal.Main</Main-Class>
```
Godin: @marchof we miss `Implementation-Title`, `Implementation-Vendor` and `Implementation-Version` for which you was fighting for `jacocoant.jar` in https://github.com/jacoco/jacoco/pull/211#issuecomment-271177924 😉

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
```diff
@@ -0,0 +1,170 @@
...
+
+	@Option(name = "-classfiles", usage = "location of Java class files", metaVar = "<path>", required = true)
+	List<File> classfiles = new ArrayList<File>();
```
marchof: @Godin Actually the decision for consistency with the existing tools. I wasn't aware of the JEP. In this case I would actually stick with the standard. This would allow to add single character aliases in future. I'll change it to double dash, ok?

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
marchof: @Godin Fixed in 5a4887dc57ced9c9e69a7de57d19375f564ace67 ✅

org.jacoco.build/pom.xml
```diff
@@ -94,12 +94,12 @@
     <module>../org.jacoco.agent.rt</module>
     <module>../org.jacoco.agent</module>
     <module>../org.jacoco.ant</module>
-
+    <module>../org.jacoco.cli</module>
+    <module>../org.jacoco.examples</module>
```
marchof: @Godin Yes, I think it was somehow confusing before: As `org.jacoco.tests` depends indirecty on `org.jacoco.examples` it is built before `org.jacoco.tests` anyways. Wanted to reflect the actual execution order here.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -41,6 +41,8 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/529">#529</a>).</li>
   <li>Maven aggregated reports will now also include modules of runtime dependencies
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/498">#498</a>).</li>
+  <li>JaCoCo now comes with a simple command line interface
```
marchof: @Godin I want to reword/re-arrange the whole block anyways before we do a 0.8.0 release. Can we do this in a separate task for 0.8.0?

org.jacoco.doc/docroot/doc/integrations.html
```diff
@@ -40,7 +40,7 @@
     </tr>
     <tr>
       <td>Command Line</td>
-      <td><a href="agent.html">JaCoCo Manual</a></td>
+      <td><a href="cli.html">JaCoCo Manual</a></td>
```
marchof: @Godin Good idea, added the link back (and also since version as for EclEmma) in ebcd1771581c6841003d872dcba933f0fb6fa238 ✅

org.jacoco.tests/pom.xml
```diff
@@ -31,8 +31,9 @@
     <module>../org.jacoco.agent.rt.test</module>
     <module>../org.jacoco.agent.test</module>
     <module>../org.jacoco.ant.test</module>
-    <module>../jacoco-maven-plugin.test</module>
+    <module>../org.jacoco.cli.test</module>
     <module>../org.jacoco.examples.test</module>
+    <module>../jacoco-maven-plugin.test</module>
```
marchof: @Godin Just to have the same order as in `org.jacoco.build`. But I can revert that if you prefer the other sequence.

org.jacoco.doc/docroot/doc/license.html
```diff
@@ -275,6 +275,14 @@
       of your accepting any such warranty or additional liability.
 </pre>
 
+<h3>args4j, Version 2.0.28</h3>
+
+<p>
+  <a href="http://args4j.kohsuke.org/">args4j</a> is subject to the terms and
+  conditions of
+  <a href="http://www.opensource.org/licenses/mit-license.php">the MIT license</a>.
```
marchof: @Godin For the others I took the original license that comes with the products. Args4j does not come with a license, only with a link to he MIT license. That's why. Let's clarify this in a separate issue?

org.jacoco.doc/docroot/doc/license.html
```diff
@@ -275,6 +275,14 @@
       of your accepting any such warranty or additional liability.
 </pre>
 
+<h3>args4j, Version 2.0.28</h3>
```
marchof: @Godin Fixed in b7423aa05517adfd4c0c5d38cbdef55c8b8be9fb ✅

org.jacoco.cli/src/org/jacoco/cli/internal/commands/ClassInfo.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jacoco.core.data.ExecutionDataStore;
+import org.kohsuke.args4j.Argument;
+
+/**
+ * The <code>classinfo</code> command.
+ */
+public class ClassInfo extends Command {
+
+	@Argument(usage = "location of Java class files", metaVar = "<classlocations>")
+	List<File> classfiles = new ArrayList<File>();
```
marchof: @Godin Intention was to keep the commands generic. So one can use it with wildcard expansion on the command line and it will also work for no matches. We can make the argument mandatory of course.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
marchof: @Godin Re empty list: As for exec files I wanted to keep the command generic and allow execution without exec files. Also there is a possible use case to get various metrics (lines, complexity etc.) in different formats.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/ExecInfo.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jacoco.core.data.SessionInfo;
+import org.kohsuke.args4j.Argument;
+
+/**
+ * The <code>execinfo</code> command.
+ */
+public class ExecInfo extends Command {
+
+	@Argument(usage = "list of JaCoCo *.exec files to read", metaVar = "<execfiles>")
+	List<File> execfiles = new ArrayList<File>();
```
marchof: @Godin See my comments about `-classfiles` and `-execfiles` before. Generic usage vs. mandatory argument.

org.jacoco.doc/xsl/cli.xsl
marchof: @Godin Hm, I wanted to express two mandatory prerequisties:

1. A JVM is required
2. The version of the JVM has to be at least 1.5 

"A Java VM with a version of 1.5 or greater"?

org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
```diff
@@ -0,0 +1,77 @@
...
+		assertContains("Usage: java -jar jacococli.jar dump", out);
+		assertContains(
+				"Request execution data from a JaCoCo agent running in 'tcpserver' output mode.",
+				out);
+	}
+
+	@Test
+	public void should_not_print_any_output_when_quiet_option_is_given()
+			throws Exception {
+		execute("version", "-quiet");
```
marchof: @Godin To keep things simple I used a args4j sub command for everything except when no command is specified. We can probably support both but that would require some additional code or tweaks to args4j.

org.jacoco.cli/about.html
marchof: @Godin Overall cleanup in a separate PR?

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
marchof: @Godin Right, we don't know which output formats have been configured. If we simple re-word this, e.g. "Analyzing 14 classes"?

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Merge.java
marchof: @Godin Same story about generic usage. We can make it mandatory, but then

    java -jar jacococli.jar merge -destfile /tmp/jacoco.exec testclients/*.exec

will fail if there are no matches.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
```diff
@@ -0,0 +1,96 @@
...
+		try {
+			final OutputStream output = new FileOutputStream(dest);
+			try {
+				return instrumenter.instrumentAll(input, output,
+						src.getAbsolutePath());
+			} finally {
+				output.close();
+			}
+		} catch (final IOException e) {
+			dest.delete();
```
marchof: @Godin Took this idea from @mathjeff 's PR. It prevents that we leave corrupted files, e.g. partially written JARs.

org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	@Test
+	public void should_print_error_message_when_invalid_command_is_given()
+			throws Exception {
+		execute("foo");
+
+		assertFailure();
+		assertNoOutput(out);
+		assertContains("\"foo\" is not a valid value for \"<command>\"", err);
```
marchof: @Godin Excellent catch! The `main` method (which unfortunately is not tested by the unit tests) did now consider the calculated return code 😞 . Fixed in 1a8ad0f30508e7ca16e1e7de37612137380322fb.

What do you think, should we add integration tests for the command line JAR?

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Dump.java
marchof: @Godin Fixed in c3e2bc9424698bac4435fb2aba51714a5c396cd3 ✅

org.jacoco.doc/docroot/index.html
marchof: @Godin Thanks for bringing this point up! What classifers do/should we actually deploy here? The plain jar or the nodeps variant?

Also what is missing here is the `nodeps` classifier on `org.jacoco.ant` and the `runtime` classifier on `org.jacoco.agent`. It looks like the table requires a new "classifer" column. Cleanup in a separate PR?

org.jacoco.core/src/org/jacoco/core/tools/ExecDumpClient.java
marchof: @Godin Yep, it's unrelated. I came across the following issue when writing test cases for cli: If a agent closes the connection without answering the the dump request we don't get an exception. The corresponding test case should be part of this PR: 

https://github.com/jacoco/jacoco/pull/525/files#diff-1b87cb3f6fcae3ed6f0c4bc8cd4fbe6c

If you want I can revert the change here and create a separate PR? The fix obviously applies to Ant ant Maven too.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
marchof: @Godin Re navigation down to source level: I can't reproduce this or I do not understand what exactly you're observing. When I try the same I get methods linked to source files and also the source switch is provided. There is even a unit test which verifies that a source page is emitted without exec files `ReportTest.should_create_html_report_when_html_option_is_provided`

marchof: @Godin Re existing destination: As with Ant and Maven existing files are overwritten (if they are writable). Do you expect a different behavior?

org.jacoco.doc/xsl/cli.xsl
```diff
@@ -0,0 +1,113 @@
...
+			</body>
+		</html>
+	</xsl:template>
+	
+	<xsl:template match="command">
+		<h2><xsl:value-of select="@name" /></h2>
+		<pre class="source" style="white-space: pre-wrap">
+			<xsl:value-of select="usage" />
+		</pre>
+		<p><xsl:value-of select="description" /></p>
```
marchof: @Godin Added warning in f1d8d8088d002745b1afa143adb56a4360d87adf ✅ 

About merging of agent and cli: Do you have a specific idea for packaging?

* Add CLI to existing `jacocoagent.jar`?
* Separate, new delivery `jacoco.jar` which replaces the agent over time?
* What about Ant tasks? Should they also be included?
* What about Maven plug-ins?

org.jacoco.core/src/org/jacoco/core/tools/ExecDumpClient.java
Godin: @marchof understanding why real agent can close connection prior to answering dump does not seem so trivial, so IMO this definitely deserves separate PR. No need to revert change here - just create separate, once its merge into master branch, merge of master into this one will result in desired effect.

marchof: @Godin Separate PR #538 ✅

org.jacoco.doc/xsl/cli.xsl
Godin: @marchof to my eyes/ears `A Java VM version 1.5 or greater is required for execution.` expresses exactly what you want, while other variants look/sound awkward. Can ask native English speaker tomorrow :wink:

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
Godin: @marchof yep - I was exactly proposing double dash :thumbsup:

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
Godin: @marchof Re navigation down to source level: okay, most likely I was confused by classes in a report that really do not have corresponding sources at specified location

Re existing destination: just was wondering, indeed you are right - we do not remove destination in other integrations too, i.e. stale files are not removed automatically

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
marchof: @Godin Fixed in 494c8f5f8e422b1f0c3d56ed4775ba7f0db9ac00 ✅

org.jacoco.cli/src/org/jacoco/cli/internal/Command.java
```diff
@@ -0,0 +1,91 @@
...
+	/**
+	 * Prints textual help for this command.
+	 * 
+	 * @param writer
+	 *            output destination
+	 */
+	protected void printHelp(final PrintWriter writer) {
+		final CommandParser parser = new CommandParser(this);
+		writer.println("Usage: " + parser.getCommand().usage(parser));
+		writer.println(description());
```
marchof: @Godin Looks definitely better, lines adjusted in 1dd730d56a74aae2a02f224a37d03e4e4fba4287 ✅ 

The error message itself comes from args4j.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
marchof: @Godin Automatic removal is dangerous, think of `--html /` 😜

org.jacoco.cli/pom.xml
marchof: @Godin Ups, thanks for finding this flaw! Fixed in 920c3cf7fc8c2658d59081bb46a974f9ebab28be ✅

org.jacoco.cli/about.html
marchof: @Godin Opened a separate issue #539 for overall cleanup.

org.jacoco.doc/docroot/index.html
marchof: @Godin Created #540 for this.

marchof:
@Godin Thanks for taking the time to perform such a detailed review! You identified quite a few relevant issues. I hope for every review remark I either fixed the issue or left a comment when I was unsure how to proceed with it.

For me the remaining main issue is how the multi value arguments `execfiles` and `classfiles` should behave when no value is provided at all. There are several options:

* Fail (make them mandatory). This may exclude some use cases for the commands.
* Print warning (might be suppressed with `--quiet`
* Consider this as valid usage (status quo)

org.jacoco.doc/xsl/cli.xsl
Godin: @marchof quoting answer of @ganncamp :
> IMO "VM" is unnecessary: Java 1.5 or greater is required for execution.

😉

org.jacoco.doc/docroot/doc/license.html
Godin: I guess we'll handle this as part of #539

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
```diff
@@ -0,0 +1,142 @@
...
+		assertContains(
+				"Usage: java -jar jacococli.jar report [<execfiles> ...]", err);
+	}
+
+	@Test
+	public void should_print_number_of_analyzed_classes() throws Exception {
+		execute("report", "--classfiles", getClassPath());
+
+		assertOk();
+		assertContains("[INFO] Writing report with 14 classes.", out);
```
Godin: @marchof as discussed in https://github.com/jacoco/jacoco/pull/525#discussion_r117818267 - `Analyzing 14 classes.` sounds better

org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
Godin: @marchof for now for simplicity let's leave without integration test

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
Godin: @marchof ah, now I see - we traverse directories recursively here and not in `Instrumenter`. `dest` here is a file passed as function argument, which I confused with a field with same name that stores command line argument.

org.jacoco.doc/docroot/doc/integrations.html
```diff
@@ -40,8 +40,8 @@
     </tr>
     <tr>
       <td>Command Line</td>
-      <td><a href="agent.html">JaCoCo Manual</a></td>
-      <td></td>
+      <td><a href="agent.html">JaCoCo Java agent</a>, <a href="cli.html">JaCoCo Command Line Interface</a></td>
+      <td>Since version 0.8.0</td>
```
Godin: @marchof given that version is only about CLI, I think will be better to have separate entries for Agent and CLI

Godin:
@marchof Re multi value arguments: the use case that you described in https://github.com/jacoco/jacoco/pull/525#discussion_r117818663 doesn't work as is for me:
```
$ bash
$ java -jar jacococli.jar merge -destfile /tmp/jacoco.exec testclients/*.exec
[INFO] Loading execution data file /private/tmp/j/testclients/*.exec.
Exception in thread "main" java.io.FileNotFoundException: testclients/*.exec (No such file or directory)

$ exit

$ zsh
$ java -jar jacococli.jar merge --destfile /tmp/jacoco.exec testclients/*.exec
zsh: no matches found: testclients/*.exec

$ exit
```

And I personally was puzzled multiple times by empty output in console after forgetting to provide an argument:
```
$ java -jar jacococli.jar classinfo
$ java -jar jacococli.jar execinfo
```
And by an empty report after forgetting to provide exec files:
```
java -jar jacococli.jar report --classfiles jacococli.jar --html report
```
Hence warned that potentially this might become a source of common mistakes.
An empty report after merge also might become a source of common mistakes:
```
java -jar jacococli.jar merge --destfile merged.exec
java -jar jacococli.jar report --classfiles jacococli.jar --html report merged.exec
```
For the `instrument` there is at least output for the ones, who are capable of reading it 😆 
```
java -jar jacococli.jar instrument --dest instrumented
[INFO] 0 classes instrumented to /private/tmp/j/instrumented.
```
So printing to console sounds like a good option.
But I'm fine to leave this as is for now.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Instrument.java
Godin: But BTW even if there is no corrupted files, users shouldn't rely on an output in case of non-zero exit code as it is incomplete

org.jacoco.doc/docroot/doc/changes.html
Godin: @marchof okay

org.jacoco.doc/xsl/cli.xsl
Godin: @marchof Re merging: well, presence of such amount of question indicates that we'd better to do a baby step without thinking about merge 😉 

But for the record: I don't see any reason to merge wiht maven plugin and was mainly thinking about addition of CLI into an agent , where all classes are common except `org.jacoco.report` and `args4j`. With Ant tasks there is even more shared classes. However for both cases story is complicated as we embed `jacocoagent.jar` into Ant tasks, which in first case will anyway lead to greatly increased size of Ant tasks, and in second case also will greatly complicate packaging due to a kind of circularity of dependencies.

bjkail:
@Godin The `nullglob` shell option should make it work, but it's not the default.

org.jacoco.doc/xsl/cli.xsl
marchof: @Godin Done, thanks @ganncamp! ✅

org.jacoco.doc/docroot/doc/license.html
marchof: @Godin I already removed the version. The remaining license issues we clarify in #539.

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
marchof: @Godin Done in baaf53946929d4d61d8646bf6b0d40626f195dc5 ✅

org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java
marchof: @Godin Done in baaf53946929d4d61d8646bf6b0d40626f195dc5 ✅

marchof:
@Godin Added warning about not providing exec files or class files to `merge`, `report`, `classinfo` and `execinfo` in 4907b227a09602a2d462fc3bfe7cf135ec8db52f ✅

org.jacoco.doc/docroot/doc/integrations.html
marchof: @Godin Fixed in 8cd9627e1b297fbbfd59778db0894cb7c60e7635 ✅

marchof:
@Godin I think I now fixed all issues identified by you. Thx again for the detailed review!

PR #505, opened 2017-05-04
Preceeding if block leads to partial instruction coverage on while statement (ECJ only)

Godin:
While inserting compiler specific tests in our validation tests it turned out that for ECJ we get different coverage results for a while block when there is a preceeding if/else block.

Test target: https://github.com/jacoco/jacoco/blob/47094b8e7670994bc9dc6e2e31a4db6b343f5a63/org.jacoco.core.test/src/org/jacoco/core/test/validation/targets/Target01.java#L46

Test verification: https://github.com/jacoco/jacoco/blob/47094b8e7670994bc9dc6e2e31a4db6b343f5a63/org.jacoco.core.test/src/org/jacoco/core/test/validation/ControlStructuresTest.java#L44


Godin:
@marchof unfortunately I completely forgot that `break` and `continue` statements are compiled into `GOTO` 😞  , so that unlikely we'll be able to implement filter here. Most likely this also means that not safe to ignore `GOTO` in filter for `finally`. And anyway I wanna at least re-add case described here to validation tests.

PR #529, opened 2017-05-06
Exclude from a report private empty constructors that do not have arguments

Godin:
After reflecting for some time, I believe that we can unconditionally safely ignore private empty constructors that do not have arguments. Even despite the fact that they relate to a source code and not generated by compiler. Note that this will completely exclude from report unused classes that have only such constructor, but anyway revealing of those is not a job of code coverage tool.

This will also improve report for JaCoCo itself.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Constructor.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Constructor.java
@@ -1,0 +1,76 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.filter.targets;
+
+import static org.jacoco.core.test.validation.targets.Stubs.nop;
+
+/**
+ * This test target is a constructors.
+ */
+public class Constructor {
+
+	Constructor() { // $line-packageLocal$
+	}
+
+	private Constructor(Object arg) { // $line-arg$
+	}
+
+	private static class Super extends Constructor {
+		private Super() {
+			super(null); // $line-super$
+		}
+	}
+
+	private class Inner {
+		private Inner() { // $line-inner$
+		}
+	}
+
+	private static class InnerStatic {
+		private Object field = this;
+
+		private InnerStatic() { // $line-innerStatic$
+		}
+	}
+
+	public static class PublicDefault { // $line-publicDefault$
+	}
+
+	static class PackageLocalDefault { // $line-packageLocalDefault$
+	}
+
+	private static class PrivateDefault { // $line-privateDefault$
+	}
+
+	private static class PrivateNonEmptyNoArg {
+		private PrivateNonEmptyNoArg() {
+			nop(); // $line-privateNonEmptyNoArg$
+		}
+	}
+
+	private static class PrivateEmptyNoArg {
+		private PrivateEmptyNoArg() { // $line-privateEmptyNoArg$
+		} // $line-return$
+	}
+
+	public static void main(String[] args) {
+		new Super();
+		new Constructor().new Inner();
+		new InnerStatic();
+		new PublicDefault();
+		new PackageLocalDefault();
+		new PrivateDefault();
+		new PrivateNonEmptyNoArg();
+		new PrivateEmptyNoArg();
+	}
+
+}
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
@@ -1,0 +1,55 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+import org.objectweb.asm.tree.VarInsnNode;
+
+/**
+ * Filters private empty constructors that do not have arguments.
+ */
+public final class PrivateEmptyNoArgConstructorFilter implements IFilter {
+
+	public void filter(final String className, final String superClassName,
+			final MethodNode methodNode, final IFilterOutput output) {
+		if ((methodNode.access & Opcodes.ACC_PRIVATE) != 0
+				&& "<init>".equals(methodNode.name)
+				&& "()V".equals(methodNode.desc)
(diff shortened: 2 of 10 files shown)


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+
+	private static class Matcher extends AbstractMatcher {
+		private static final VarInsnNode THIS = new VarInsnNode(Opcodes.ASTORE,
+				0);
+
+		private boolean match(final MethodNode methodNode,
+				final String superClassName) {
+			vars.put("this", THIS);
```
marchof: The trick with `ASTORE` is to define a named argument. This looks surprising as there is no `ASTORE` involved in private empty constructors.

For better documentation I would therefore prefer to encapsulate this within `AbstractMatcher` and  provide a method for this:

    void defineLocal(int idx, String name)

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
```diff
@@ -0,0 +1,55 @@
...
+			cursor = methodNode.instructions.getFirst();
+			nextIsVar(Opcodes.ALOAD, "this");
+			nextIs(Opcodes.INVOKESPECIAL);
+			MethodInsnNode m = (MethodInsnNode) cursor;
+			if (m != null && superClassName.equals(m.owner)
+					&& "<init>".equals(m.name) && ("()V").equals(m.desc)) {
+				nextIs(Opcodes.RETURN);
+				return cursor != null;
```
Godin: @marchof then there will be temptation to use it with `&&` everywhere, but that is exactly what we were trying to avoid.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
Godin: @marchof I removed usage of `nextIsVar` - it does not work for a first instruction in absence of preceding label.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
marchof: @Godin It's internal API, I would try to keep the filter as readable as possible.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
marchof: @Godin Ok, I see. I would still prefer to implement semantical matching premitives in `AbstractMatcher`, so we can write e.g.

		private boolean match(final MethodNode methodNode,
				final String superClassName) {
			cursor = methodNode.instructions.getFirst();
			skipNonCodes();
			nextIsVar(Opcodes.ALOAD, 0);
			nextIsMethod(Opcodes.INVOKESPECIAL, superClassName, "<init>",
					"()V");
			return // true if no more  statements, see separate comment
		}

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Constructor.java
```diff
@@ -0,0 +1,67 @@
...
+	public static class PublicDefault { // $line-publicDefault$
+	}
+
+	static class PackageLocalDefault { // $line-packageLocalDefault$
+	}
+
+	private static class PrivateDefault { // $line-privateDefault$
+	}
+
+	private static class PrivateEmptyNoArg {
```
marchof: @Godin I wonder whether we should add another negative test for the case where the private non-arg constructor is not empty in source code: 

	private static class PrivateNonEmptyNoArg {
		private PrivateEmptyNoArg() {
			nop(); // $line-privateNonEmptyNoArg$
		}
	}

Godin: Done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
Godin: @marchof I added `skipNonOpcodes`, but `nextIsVar` anyway can't be used for a very first instruction, because after `skipNonOpcodes` we need to check **current instruction**, while semantic of `next...` methods is to move to a next instruction. Change of semantic to check of current instruction - first of all will affect a lot of existing code relying on it, but anyway will lead to the same problem, a kind of, but at the end of method. IMO the only way - is to insert an artificial label at the beginning of all methods, but IMO this is an overkill compared just to a condition in `if` here, which is IMO perfectly readable.

Replacement of another `if` on method `nextIsMethod` in abstract class IMO also an overkill, because  its only usage will be here, and there won't be code-sharing with other places where we already check `MethodInsnNode` (see references on it in `TryWithResourcesJavacFilter` and `TryWithResourcesEcjFilter`), and again IMO condition is perfectly readable.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
Godin: @marchof Then in case of preceding `if` statement containing `return` of a boolean, IDE start to suggest to combine them, making code unreadable. Also even if I don't see this in IntelliJ, static analyzers might warn about unused return value in other places.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Constructor.java
marchof: @Godin Thx! :+1:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
marchof: @Godin Thanks for the explations. This would require to keep a state in `AbstractMatcher` and replace the direct access to `cursor` with simething like `start(AbstractInsnNode)`.

But I agree that refactoring the matcher API soes not provide enough value for now.

alexanthony813:
Hi, 

Thanks for the change! How are we supposed to use this filter? I did not see any notation here

https://github.com/jacoco/jacoco/wiki/FilteringOptions

MonsieurBon:
Do you have an estimate about when this could be released?

marchof:
@MonsieurBon Hopefully this year.

You can use our snapshot builds in the meantime and help us testing!

saadlu:
@marchof 

Snapshot build is fine with me, but I can't seem to test this. 
This is what I have done

1. Checkout the latest master
    As of now, this is the master HEAD (your commit I assume :))
    
    ```
    commit 8f0460960967e15e46c9f2bf766c0bfe9e18e6b8
    Author: Marc R. Hoffmann <hoffmann@mountainminds.com>
    Date:   Thu Nov 2 22:07:24 2017 +0100

    Remove obsolete phase binding (#617)
    
    Since version 0.6.2 the "report" goal is bound to phase "verify".
    ```
1. built locally (with `mvn clean install`)
1. Using a gradle project to test out the local built jacoco

   ```
   apply plugin: 'jacoco'
   jacoco {
      toolVersion = "0.7.10-SNAPSHOT"
   }
   ```
   testing code:
   ```
   public class Empty {

       private Empty(){}

       public static String say(){
           return "say";
       }
    }

   ```
   with 
   ```
   public class EmptyTest {
      @Test
      public void say() {
        assertThat(Empty.say(), is("say"));
      }

   }
   ```
1. Using sonar to generate me the report 
1. Sonar says only 50% covered. Complaining that the private constructor is not convered

what am I doing wrong?

Godin:
@saadlu see https://github.com/jacoco/jacoco/pull/513#issuecomment-293176354

Issue #323, opened 2015-06-08
Race condition in integration test for maven goal "dump"


Godin:
This test constantly fails in https://jacoco.ci.cloudbees.com/job/jacoco-jdk8/

PR #530, opened 2017-05-08
Fix random test failures in TcpClientOutputTest and TcpConnectionTest

marchof:


marchof:
@Godin Can you please review this? Should fix some of our random build failures.

(Sorry, currently can't set tags or assign reviewer through GitHub web interface. Maybe some internet connectivity issue since I'm on the road).

marchof:
Ironically our other source of random just failed one of the builds (`it-dump/pom.xml`)

Godin:
@marchof BTW `testRemoteClose` also fails from time to time.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/TcpClientOutputTest.java
```diff
@@ -81,17 +82,25 @@ public void testRemoteClose() throws Exception {
...
+	public void should_log_exception_when_invalid_command_is_sent()
+			throws Exception {
 		remoteWriter.visitSessionInfo(new SessionInfo("info", 1, 2));
-		while (remoteReader.read()) {
+
+		try {
+			while (remoteReader.read()) {
+			}
+		} catch (EOFException e) {
+			// Might be thrown as the socket is directly closed by the client
```
marchof: @Godin Now thst you're asking I realize that I haven't fully understood the reason neither :flushed: 

I'll investigate further and provide a updated commit. At least I can reproduce it locally from time to time.

marchof:
The exception for `TcpClientOutputTest.testInvalidCommand` is:

	java.io.EOFException
	at java.io.DataInputStream.readChar(DataInputStream.java:333)
	at org.jacoco.core.data.ExecutionDataReader.readHeader(ExecutionDataReader.java:127)
	at org.jacoco.core.data.ExecutionDataReader.readBlock(ExecutionDataReader.java:109)
	at org.jacoco.core.runtime.RemoteControlReader.readBlock(RemoteControlReader.java:47)
	at org.jacoco.core.data.ExecutionDataReader.read(ExecutionDataReader.java:92)
	at org.jacoco.agent.rt.internal.output.TcpClientOutputTest.testInvalidCommand(TcpClientOutputTest.java:86)

marchof:
@Godin I think I can now explain the cause of the problem. We have two threads here:

- T1: The main thread executing the test. During `setup()` this thread synchronously created both ends of the connection. In `TcpClientOutput.startup()` the method `TcpConnection.init()` is called. This creates a new `RemoteControlWriter` which writes an exec file header to the socket.
- T2: This internal thread of `TcpClientOutput` also started during `setup()` and processes the received requests.
- T1: Sends a invalid request

Now the concurrency problems starts:

- T1: starts reading the exec file header from the buffer
- T2: sees an invalid request and closes the socket

Now if the socket is closed by T2 while T1 is reading the header we get the `EOFFileException`.

I'll rework the test.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/TcpConnectionTest.java
```diff
@@ -103,8 +104,8 @@ public Void call() throws Exception {
 	 * Remote endpoint is closed before even a valid header was send.
 	 */
 	public void testRemoteCloseWithoutHeader() throws Throwable {
```
Godin: @marchof a bit unrelated, but I'm wondering why `Test` annotation was removed from this method in - 54e674c0ab96b93668e6c0b036e7773e3ed7ff6d ?

marchof: @Godin Nice catch! Will fix this as part of this PR.

marchof: @Godin Fixed.

marchof:
@Godin Yes, the irony 😕 ... do we have access to the test results on Travis? Would be interested in the stack trace for this test failure.

marchof:
@Godin I can now reproduce the issue (by simply executing the test in a loop).

Godin:
@marchof 😆  I just did the same and was about to post a stack trace 😉

Godin:
@marchof answering question about tests results in Travis: Unfortunately once build completed we don't have more than its log as it is now. Investment of time into making logs more verbose in case of failures unfortunately was never in top priorities since normally they are green and obvious failures are catched prior to commit. And in this particular case we would need to wait next failure... Now since we've found a way to reproduce these random failures locally and will fix them, for me priority compared to other tasks is even lower than before...

marchof:
@Godin Thanks for the info! I hope I can stabilize the unit tests. Just with the latest build another common candidate failed one of the builds: `it-dump/pom.xml`

Godin:
@marchof BTW we have a ticket for `it-dump/pom.xml` - https://github.com/jacoco/jacoco/issues/323 Looking at the code of this test: seems that race condition - is that dump might be requested prior to start of `main` method.

marchof:
@Godin `TcpClientOutputTest.testRemoteClose()` should be fixed now (at least I can't reproduce race condition any more in a loop).

PR #538, opened 2017-05-22
ExecDumpClient should report error when no execution data is retrieved

marchof:
Ant and Maven `dump` commands use the `ExecDumpClient` class from the `org.jacoco.core` module. This utility does not evaluate the response status after sending a `dump` request. If the remote closes the connection without sending a response this is silently ignored.

Expected behavior is that a `IOException` is thrown when a `dump` request is not answered.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/tools/ExecDumpClientTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/tools/ExecDumpClientTest.java
@@ -101,7 +101,7 @@ public class ExecDumpClientTest {
 				// 3. Retry
 				"onConnectionFailure", "onConnecting")
 
-		, callbacks);
+				, callbacks);
 	}
 
 	@Test
@@ -124,6 +124,12 @@ public class ExecDumpClientTest {
 		client.dump((String) null, port);
 		assertFalse(dumpRequested);
 		assertTrue(resetRequested);
+	}
+
+	@Test(expected = IOException.class)
+	public void testNopServer() throws IOException {
+		int port = createNopServer();
+		client.dump((String) null, port);
 	}
 
 	private int getFreePort() throws IOException {
@@ -159,11 +165,27 @@ public class ExecDumpClientTest {
 				dumpRequested = dump;
 				resetRequested = reset;
 				if (dump) {
-					writer.visitSessionInfo(new SessionInfo("TestId", 100, 200));
+					writer.visitSessionInfo(
+							new SessionInfo("TestId", 100, 200));
 				}
 				writer.sendCmdOk();
 			}
 		});
 		reader.read();
 	}
+
+	private int createNopServer() throws IOException {
+		server = new ServerSocket(0, 0, InetAddress.getByName(null));
+		new Thread(new Runnable() {
+			public void run() {
+				try {
+					server.accept().close();
+				} catch (IOException e) {
+					// ignore
+				}
+			}
+		}).start();
+		return server.getLocalPort();
+	}
+
 }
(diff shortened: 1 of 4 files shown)


marchof:
@Godin No idea how the actual JaCoCo agent would behave like this. This PR is more about making the client bullet-proof especially about missconfiguration.

org.jacoco.core/src/org/jacoco/core/tools/ExecDumpClient.java
```diff
@@ -123,7 +123,10 @@ public ExecFileLoader dump(final InetAddress address, final int port)
 					.setExecutionDataVisitor(loader.getExecutionDataStore());
 
 			remoteWriter.visitDumpCommand(dump, reset);
-			remoteReader.read();
+
+			if (!remoteReader.read()) {
```
Godin: @marchof we have similar code in `org.jacoco.examples/src/org/jacoco/examples/ExecutionDataClient.java`. Later we can update it to use `ExecDumpClient`, but IMO anyway will be better to apply same modification in it with this PR.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -41,6 +41,9 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/529">#529</a>).</li>
   <li>Maven aggregated reports will now also include modules of runtime dependencies
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/498">#498</a>).</li>
+  <li><code>dump</code> commands now report errors when remote connections are
+      closed unexpectedly
```
marchof: @Godin Fixed in c9d9887758ff04627edd663dfffd0bf4f2e3c99a ✅

org.jacoco.core.test/src/org/jacoco/core/tools/ExecDumpClientTest.java
```diff
@@ -126,6 +126,12 @@ public void testReset() throws IOException {
 		assertTrue(resetRequested);
 	}
 
+	@Test(expected = IOException.class)
+	public void testNopServer() throws IOException {
```
marchof: @Godin Good catch! Needed to insert a Thread.sleep() to reproduce this, but I was able to reproduce it. Fixed in 42b4f9f57d4403778f49c89a5790f287a06f1572 ✅ 

I didn't change the method name originally, as the code change was a 1:1 patch from #525. Now as we needed to adjust the code I also change the method name as suggested. To avoid merge conflicts I will revert the corresponding change in #525, correct?

org.jacoco.core/src/org/jacoco/core/tools/ExecDumpClient.java
marchof: Fixed in acfb4b0add8620f20f904e9f4275ca68c3692903 ✅

Godin:
@marchof :thumbsup:

org.jacoco.core.test/src/org/jacoco/core/tools/ExecDumpClientTest.java
Godin: @marchof well, IMO no big difference whether to explicitly revert or resolve conflict during merge, but I see that you already did a revert

denisa:
Thanks for having that fixed, I just spend an hour tracking that down on 0.7.9.

My use case:
I have a docker image that contains wildfly and the jacobo agent jar file.
The docker image always exposes port 6300, but I decide upon launching the image if I want to enable the jacoco agent or not.

When I connect to a running docker instance that does not have jacoco enabled, ExecDumpClient.dump() returns an (empty) ExecFileLoader instead of failing as desired.

Godin:
@denisa could you please clarify just to be sure - do you mean that after this change in your use case you're getting `IOException` as expected?

denisa:
Yes, this bug is fixed: 0.7.10-SNAPSHOT throws an IOException as designed when 0.7.9 simply returned an object.

Godin:
@denisa cool! thank you for information :thumbsup:

PR #541, opened 2017-05-25
Method Analyzer.analyzeClass(InputStream, String) as others should provide context information when unable to read input during analysis

Godin:
This was forgotten in #400

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
@@ -41,9 +41,6 @@ public final class Java9Support {
 	 */
 	public static byte[] readFully(final InputStream is)
 			throws IOException {
-		if (is == null) {
-			throw new IllegalArgumentException();
-		}
 		final byte[] buf = new byte[1024];
 		final ByteArrayOutputStream out = new ByteArrayOutputStream();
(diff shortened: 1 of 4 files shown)


org.jacoco.core/src/org/jacoco/core/internal/Java9Support.java
```diff
@@ -41,9 +41,6 @@ private Java9Support() {
 	 */
 	public static byte[] readFully(final InputStream is)
 			throws IOException {
-		if (is == null) {
```
marchof: @Godin Why do we need this check specifically here? I think all other APIs like `analyzeAll()` will result in a NPE for a `null` `InputStream`.  Which looks ok for me.

Godin: @marchof I guess that you misread change: it doesn't add check, but removes it exactly because we don't need it - indeed in all places for a `null` `InputStream` we throw `NullPointerException` instead of `IllegalArgumentException`, also this condition wasn't covered by tests, so I took opportunity to remove it in this PR.

marchof: @Godin Sure, my bad. You actually fixed the issue - thanks!!!

PR #543, opened 2017-06-08
NullPointerException in CLI when instrumenting single class

marchof:
```
java -jar jacococli.jar instrument Example.class --dest instrumented
```

leads to

```
Exception in thread "main" java.lang.NullPointerException
	at org.jacoco.cli.internal.commands.Instrument.instrument(Instrument.java:78)
	at org.jacoco.cli.internal.commands.Instrument.instrumentRecursive(Instrument.java:72)
	at org.jacoco.cli.internal.commands.Instrument.execute(Instrument.java:56)
	at org.jacoco.cli.internal.Main.execute(Main.java:89)
	at org.jacoco.cli.internal.Main.main(Main.java:104)
```

while

```
java -jar jacococli.jar instrument Example.class --dest ./instrumented
```

works and produces file `/tmp/./instrumented`

```
[INFO] 1 classes instrumented to /tmp/./instrumented.
```

however description of `--dest` option:

```
--dest <dir>  : path to write instrumented Java classes to
```

### DIFF
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/InstrumentTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/InstrumentTest.java
@@ -54,7 +54,7 @@ public class InstrumentTest extends CommandTestBase {
 	}
 
 	@Test
-	public void should_instrument_class_files_and_copy_resources()
+	public void should_instrument_class_files_and_copy_resources_when_folder_is_given()
 			throws Exception {
 		File destdir = tmp.getRoot();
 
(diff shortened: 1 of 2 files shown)


Godin:
@marchof I came across the same problem as you: can't close review by approve - GitHub keeps saying
> You need to leave a comment indicating the requested changes.

Or at least one bizarre side effect is observable here - issue was created by me, but now first comment looks like was created by you.

In any case - review and manual test have been done, so I'll bypass review UI and will do merge.

marchof:
@Godin Thanks for merging!

PR #559, opened 2017-07-03
Regression in 0.7.9: "help" goal is not available in jacoco-maven-plugin

Godin:
```
mvn org.jacoco:jacoco-maven-plugin:0.7.8:help
```

works, while

```
mvn org.jacoco:jacoco-maven-plugin:0.7.9:help
```

fails as well as

```
mvn org.jacoco:jacoco-maven-plugin:0.7.10-SNAPSHOT:help
```

### DIFF
--- a/jacoco-maven-plugin.test/it/setup-parent/invoker.properties
+++ b/jacoco-maven-plugin.test/it/setup-parent/invoker.properties
@@ -1,0 +1,1 @@
+invoker.goals = org.jacoco:jacoco-maven-plugin:${project.version}:help install
--- a/jacoco-maven-plugin/pom.xml
+++ b/jacoco-maven-plugin/pom.xml
@@ -116,13 +116,9 @@
           <execution>
             <id>default-descriptor</id>
             <phase>process-classes</phase>
-            <goals>
-              <goal>descriptor</goal>
-            </goals>
           </execution>
           <execution>
-            <id>help-mojo</id>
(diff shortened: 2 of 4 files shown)


marchof:
@Godin Anyways, wouldn't it make sense to add a simple IT test to ensure there is a help goal which produces some output? We have the same for CLI.

Godin:
@marchof I already added -[ see new file `jacoco-maven-plugin.test/it/setup-parent/invoker.properties`](https://github.com/jacoco/jacoco/pull/559/files#diff-7842d0e44be15e48c4fe637b78434ed5)

Godin:
@marchof it just checks that there is `help` goal, but doesn't check content, what is IMO sufficient - we haven't seen problems with content so far, so why not trust generator

Godin:
@marchof on the other hand - I can add simple check of content, but anyway it won't catch complex regressions

marchof:
@Godin Ok, thanks. I was confused as I haven't seen a separate test for this. The setup parent is used by all tests. This means the help goal is now executed within all tests?

Godin:
@marchof no - this means that when preparing (`install`ing) `setup-parent` we'll test availability of `help` goal, i.e. only once. I did it this way to avoid addition of test, because each separate test is quite costly in terms of time.

marchof:
@Godin Ok, I see. Fine to me.

Issue #539, opened 2017-05-22
Consistent license infos


marchof:
We need to review and clean-up the license information we ship with JaCoCo. 

- [ ] Consistent license information in all published artifacts
- [ ] Check license information for `nodeps` JARs
- [ ] Use proper terms (e.g. our artifacts are referred to as "plug-ins" which technically is not true for some of them - see https://github.com/jacoco/jacoco/pull/525#discussion_r117550561)
- [ ] Ensure that we comply with license terms for all our 3rd party products
- [ ] Avoid duplicate/redundant license references (see https://github.com/jacoco/jacoco/pull/525#discussion_r117527532)
- [ ] Document best practices for future reference

PR #561, opened 2017-07-06
Relocate 3rd party libs in CLI jar

Godin:
To avoid any version conflicts we should relocate all 3rd party library dependencies as we do for the Ant jar with classifier `nodeps`.


Godin:

Godin:
@marchof I don't think that you forgot - we created #539 to clarify license issues

Godin:
@marchof in particular word `plug-in` in `about.html` was questionable (as well as maybe `about.html` in general), because it is nor OSGi bundle, nor Eclipse plug-in, nor any other plugin - see https://github.com/jacoco/jacoco/pull/525#discussion_r117550561

marchof:
@Godin You're right, I added check point for `nodeps` artifacts.

PR #572, opened 2017-08-02
Add Artifact.SCOPE_PROVIDED artifacts to aggregate report

chonton:
I have an aggregate maven project where one of the projects must be scoped as 'provided'.  I would like it to be still considered for inclusion in the aggregate report.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
@@ -42,8 +42,8 @@ import org.jacoco.report.IReportGroupVisitor;
  * </p>
  * 
  * <ul>
- * <li><code>compile</code>, <code>runtime</code>: Project source and execution
- * data is included in the report.</li>
+ * <li><code>compile</code>, <code>runtime</code>, <code>provided</code>:
+ * Project source and execution data is included in the report.</li>
  * <li><code>test</code>: Only execution data is considered for the report.</li>
  * </ul>
  * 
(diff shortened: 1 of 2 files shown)


marchof:
@chonton Thanks for the contribution! May I ask you for your use case? Especially it looks like my current assumptions do not hold true:

1) A separate project only for the purpose of creating the coverage report is required. In your case it looks like you can rely on a existing project.
2) My understanding of usage of scope PROVIDED is a dependency on a common API (like Servlet) which is typically not part of the reactor itself. Our goal can only create reports for projects from the reactor.

chonton:
My typical reactor project has three modules:
- ```api``` (interface jar)
- ```service``` (implementation delivered as a docker image)
-  ```test``` (integration tests in a docker image)

Both ```service``` and ```test``` have a **compile** dependency on ```api```.  Additionally, ```test``` has a particular dependency on ```service```;  the  ```service``` module must be built before ```test``` so that ```service``` docker image can be launched during  **pre-integration-test** phase, and the classes from ```service``` should not be included in the compile or runtime classpath of ```test```. 

To prevent the ```service``` classes from being included in the ```test``` docker image, the dependency cannot be **compile**  or **runtime**.  The jacoco aggregate goal uses **test** to distinguish the projects that provide coverage without providing code.

This leaves  **provided**, which seems correct, as the ```service``` is providing the implementation of ```api```.

marchof:
@chonton  Indeed, this is nice use case where the aggregate-report works well. Thx!

@Godin Do you agree to merge this? If so I we need to add a change log.

chonton:
@marchof and @Godin, anything else required to merge this PR?

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -46,7 +46,20 @@
  * Project source and execution data is included in the report.</li>
  * <li><code>test</code>: Only execution data is considered for the report.</li>
  * </ul>
- * 
+ *
+ * A typical use case might be in a reactor project which has three modules:
```
Godin: I don't think that we need this update of Javadoc.

chonton: removed in next commit.

chonton:
anything else needed to merge this PR?

Godin:
> anything else needed to merge this PR?

@chonton not so much - I just need to come from vacation back to work :wink:

Godin:
@chonton merged. Thank you for your contribution! :thumbsup:

PR #578, opened 2017-08-18
Add --verbose option to classinfo

marchof:
While analyzing reported issue I extended the `classinfo` command to provide details about every method and line. As this was useful to quickly spot an issue I decided to provide an official option for this.


marchof:
@Godin Can you please have a look if you find some time? Thx

PR #581, opened 2017-08-23
[dev] enable coverage reporting on a per-commit level

juliangamble:
This is a new Jacoco Report that shows coverage at a commit level. 

To see an example of this - please see the example project here:
https://github.com/juliangamble/jacoco-example-project

The usage scenario is for large teams on large projects. In this situation, individual team members care less about overall coverage and more about the coverage on the lines they have touched. This plugin enables that team members to say "I'm working on the overall coverage incrementally - for now my commits have 80% coverage."

I'm open to suggestions and feedback about how this could be improved. Please tell me what needs to happen to get this merged into the mainline.

### DIFF
--- a/jacoco-maven-plugin/META-INF/m2e/lifecycle-mapping-metadata.xml
+++ b/jacoco-maven-plugin/META-INF/m2e/lifecycle-mapping-metadata.xml
@@ -19,6 +19,7 @@
           <goal>report</goal>
           <goal>report-integration</goal>
           <goal>report-aggregate</goal>
+          <goal>report-on-commit-coverage</goal>
           <goal>check</goal>
           <goal>dump</goal>
           <goal>instrument</goal>
--- a/jacoco-maven-plugin/pom.xml
+++ b/jacoco-maven-plugin/pom.xml
@@ -84,6 +84,31 @@
       <!-- annotations are needed only to build the plugin: -->
       <scope>provided</scope>
     </dependency>
+
+
+    <dependency>
+      <groupId>org.eclipse.jgit</groupId>
+      <artifactId>org.eclipse.jgit.java7</artifactId>
+      <version>3.7.1.201504261725-r</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jacoco</groupId>
+      <artifactId>jacoco-maven-plugin</artifactId>
+      <!-- version for Java 7 -->
+      <version>0.6.5.201403032054</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>1.3.2</version>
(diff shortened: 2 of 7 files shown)


marchof:
@juliangamble Thanks for your contribution!

Some remarks:

1) Not sure whether I see this in scope for JaCoCo. There are great tools for this like SonarQube. We accumulating too many add ons here at the JaCoCo project.
2) Why is the implementation only in the Maven plugin? Why only git? We have different integrations for our functionality. If such functionality should be integrated in JaCoCo we need a common API which then is used by the different integrations
3) As you obviously care about test coverage, why do you think test coverage is not relevant for the JaCoCo project? :wink: 

But to clearly manage expectations I don't think we will extend JaCoCo into this direction. What about maintaining this as a separate project?

juliangamble:
Hi @marchof, 

A broader explanation of the usage scenarios are here:
https://groups.google.com/forum/#!topic/jacoco/Dat1WLBoDZU

Re (1) - I use sonarQube daily. The problem is that you can't know in advance what level of coverage your changes in your commit have prior to pushing. This is directed at solving that problem. 

Re (2a) It's in the maven plugin for convenience and simplicity. If there is a better way to do this let me know. 

Re (2b) Why only git? I'm not sure I understand the assumptions behind your question. Are you suggesting that mercurial, CVS or clearcase should be in scope? Largely pragmatism was the reason for this choice. Also because JGIT has a powerful API, and I'm not convinced that equivalent VCS client libraries exist for the other version control systems. 

Re (3) Fair call. I can get back to you on that. I wanted a minimum working proposal for feedback - and it sounds like you've provided the feedback you're going to give. 

If this were to be maintained as a separate project - could you give an example of something similar?

marchof:
@juliangamble I understand your use case and I fully buy the idea. My point it that Maven/Git is your specific scenario while JaCoCo is used by thousands of projects with different build tools and repositories. From the experience in the past we learned that if we implement a feature for a specific scenario only we imediately get requests for our other integrations. And I don't see dealing with different repository technolgies in scope of the JaCoCo project. We simply have no budget to maintain this. Therefore we focus on the code code coverage technology.

But JaCoCo provides well documented APIs and is released to Maven. I don't have a specific example at hand but there are [plenty of integrations](http://www.jacoco.org/jacoco/trunk/doc/integrations.html) and also in in [Maven Central](http://search.maven.org/#search%7Cga%7C1%7Cjacoco) you find many usages.

juliangamble:
hanks for your support. The best path forward was for this to go as a separate library. You can use it here:

https://github.com/juliangamble/commit-level-coverage-report

There is an example project here:

https://github.com/juliangamble/jacoco-example-project

Cheers
Julian

marchof:
@juliangamble Thanks for your understanding and for picking this ides up as a separate tool!

PR #596, opened 2017-09-21
Add filter for bytecode that javac generates for String in switch

Godin:
I propose to do one more baby step in story of filtering - add filter for bytecode that **javac** generates for String in switch. And modify it for ECJ separately a bit later since it is a bit more tricky.


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/StringSwitchTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.junit.Test;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class StringSwitchTest implements IFilterOutput {
```
marchof: Should be calles `StringSwitchFilterTest` as target is `StringSwitchFilter`. Or `StringSwitchFilterJavacTest`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchFilter.java
```diff
@@ -0,0 +1,73 @@
...
+import org.objectweb.asm.tree.LabelNode;
+import org.objectweb.asm.tree.LookupSwitchInsnNode;
+import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+import org.objectweb.asm.tree.TableSwitchInsnNode;
+
+/**
+ * Filters code that is generated by javac for a switch statement with a String.
+ */
+public final class StringSwitchFilter implements IFilter {
```
marchof: As this filter is specific for javac we should call it `StringSwitchJavacFilter`.

marchof:
@Godin BTW, +1 for moving on with baby steps, thanks!

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchFilter.java
Godin: Haven't used `Javac` prefix, thinking to update it for ECJ later, but you're right. Done.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/StringSwitchTest.java
Godin: This explains why switching between code and test wasn't working in IDE :laughing: Fixed.

Godin:
> I think this filter is not specific enough: It filters every switch on String.hashCode()
> I don't think it is very likely that we find such cases in real code but I think the filter could be more specific.

Well, was about to do push-back and say "if unlikely, then why complexify? This will make step bigger and harder than baby". But actually making it more specific wasn't that hard thanks to previously implemented `AbstractMatcher` and implementation with it is quite nice. So lets say that I was lazy :wink: 
Please re-review.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	private static class Matcher extends AbstractMatcher {
+		boolean match(final AbstractInsnNode start,
+				final AbstractInsnNode secondSwitchLabel) {
+			cursor = start;
+			for (int i = 0; cursor != null && i < 4; i++) {
+				cursor = cursor.getPrevious();
+			}
+			if (cursor == null || cursor.getOpcode() != Opcodes.ICONST_M1) {
```
marchof: Couldn't that be `nextIs(Opcodes.ICONST_M1)` followed by `if (cursor == null) return;`?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java
```diff
@@ -0,0 +1,97 @@
...
+		if (new Matcher().match(start, dflt)) {
+			output.ignore(start, dflt);
+		}
+	}
+
+	private static class Matcher extends AbstractMatcher {
+		boolean match(final AbstractInsnNode start,
+				final AbstractInsnNode secondSwitchLabel) {
+			cursor = start;
+			for (int i = 0; cursor != null && i < 4; i++) {
```
marchof: Would be nice to have a fluent API for this in Matcher

    moveBack(4);

or something like this.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java
Godin: @marchof Unfortunately no: `nextIs` first moves to next instruction and then checks, so can't be used to verify current instruction, and we can't place cursor before very first instruction of method.

Godin: @marchof this is kind of catch 22 - either we need to special case checking of beginning, or will have problem with end of methods, i.e. if we change semantic and cursor will be pointing to instruction for checking and moved after checking, then `null` can not be used to represent no-match as it is now, because `next` of last instruction is `null`. Maybe we can represent no-match by a special non-null value, but will be better to do this separately from this PR as it requires change of semantics `next` methods.

Godin: @marchof another option is to have fake instruction at the beginning:
```
private void moveBack(int num) {
	for (int i = 0; i < num; i++) {
		if (cursor.getPrevious() == null) {
			cursor = new InsnNode(Opcodes.NOP) {
				final AbstractInsnNode next = cursor;
				@Override
				public AbstractInsnNode getNext() {
					return next;
				}
			};
		} else {
			cursor = cursor.getPrevious();
		}
	}
}
```

But I doubt that this is a good idea in general and definitely too much of complexity if done only here, compared to current handling of beginning.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java
Godin: @marchof I would prefer to leave this as is and extract only if will be used somewhere else, otherwise IDE complains that value of parameter is always `4`.

Issue #553, opened 2017-06-25
generate report when sources not available


ggeorgovassilis:
### Steps to reproduce

JaCoCo version: jacoco-maven-plugin 0.7.9
Operating system: windows
Tool integration: Maven 3.3.9/Eclipse

Please see [test case](https://github.com/ggeorgovassilis/test-with-embedded-tomcat/tree/jacoco_bug_report)

`mvn clean install`

### Expected behaviour
An HTML report consistent to the entries in jacoco.exec is rendered

### Actual behaviour
1. observe that test-with-embedded-tomcat/webapp-integration-tests/target/jacoco.exec contains test coverage as expected
2. observe that test-with-embedded-tomcat/webapp-integration-tests/target/site/jacoco/index.html contains an empty report.
3. observe that maven will say that "Analyzed bundle 'webapp-integration-tests' with 0 classes".

ggeorgovassilis:
Please not that when a class with source code is included in the webapp-integration-test project, it (and only it) will show up fine in the report.

marchof:
@ggeorgovassilis A general remark about JaCoCo design:

* The report contains all classes which are provided to the analyzer as class files.
* Source files are optional (no source coloring without sources of source)
* The exec file does **not** define the scope of the report (execfiles contain all classes loaded by the JVM, which typically is too much, classes missed at all by tests are also not included)

Godin:
@ggeorgovassilis more precisely about your example: generation of report requires class files and 
 and `report` goal uses classes of module where it is executed and hence generates report containing them, so the behavior that you observe is a correct one, because module `webapp-integration-tests` has no classes. To have a report that includes classes of other modules within Maven reactor please use `report-aggregate` (http://www.jacoco.org/jacoco/trunk/doc/report-aggregate-mojo.html). In complex scenarios/setups to have a more fine grained control of what is used (classes, sources) for generation of report consider usage of JaCoCo Ant Tasks (http://www.jacoco.org/jacoco/trunk/doc/ant.html).

ggeorgovassilis:
Thank you @Godin

PR #556, opened 2017-09-25
Improve diagnostics for missing coverage report

marchof:
### Steps to reproduce

While debugging #553, the maven jacoco plugin logs (-X option) were not helpful. As a reminder: an extensive jacoco.exec was produced, but coverage reports were empty.

JaCoCo version: 0.7.9
Operating system: windows
Tool integration: Maven

### Expected behaviour

Logs should show:
- inventory of classes included in the report and the rule that applied for the inclusion (e.g. instrumented by agent based on <includes> rule r1)
- inventory of classes excluded from the report and the rule that lead to the exclusion (e.g. found class X in project but was not instrumented because of exclusion rule e1)
 

### Actual behaviour
Logs state nothing about why certain classes don't show up in coverage reports.


marchof:
@Godin We just added some diagnostics for report generation. What do think about this PR? Adding minimal INFO logs to the Maven goal where and whether classes are loaded will also help. I can refresh this PR if you agree. Otherwise I would like to close this.

Issue #487, opened 2017-02-08
NegativeArraySizeException


Godin:
```
java.io.IOException: Error while instrumenting class Example.
Caused by: java.lang.NegativeArraySizeException
	at org.objectweb.asm.Frame.merge(Frame.java:1447)
	at org.objectweb.asm.CurrentFrame.execute(CurrentFrame.java:52)
	at org.objectweb.asm.MethodWriter.visitInsn(MethodWriter.java:747)
	at org.objectweb.asm.ClassReader.readCode(ClassReader.java:1367)
	at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1032)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:708)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:521)
	at org.objectweb.asm.ClassWriter.toByteArray(ClassWriter.java:1003)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:91)
```

This happens when resizing jump instruction during instrumentation in a class without frames.

Corresponding ASM issue: https://gitlab.ow2.org/asm/asm/issues/317784

Godin:
Fixed by #600

Issue #584, opened 2017-08-30
getEnclosingClass() and getDeclaringClass() return null [reflection]


jvanheesch:
This is an issue tracker. Please use our mailing list for general questions: 
https://groups.google.com/forum/?fromgroups=#!forum/jacoco

Also check FAQ before opening an issue: http://www.jacoco.org/jacoco/trunk/doc/faq.html

### Steps to reproduce
quickstart example:
[quickstart.tar.gz](https://github.com/jacoco/jacoco/files/1264268/quickstart.tar.gz)

Run (`mvn test`) the quickstart example.
JaCoCo version: 0.7.9
Operating system:
Tool integration: Maven

### Expected behaviour
`SampleTest` doesn't fail - `getEnclosingClass()` and `getDeclaringClass()` return `SomeClass.class`.

### Actual behaviour
When running the test with Maven and JaCoCo (`mvn test`), the test fails.
If the JaCoCo plugin is removed, `mvn test` will successfully run the test.

Godin:
@jvanheesch  first of all - thank you for this report with a complete concise reproducer! :thumbsup:

In presence of Serializable lambdas javac will generate [method `$deserializeLambda$`](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/SerializedLambda.html). This method is big in given example due to amount of lambdas. So that JaCoCo instrumentation adds wide jumps into this method. They are handled in a specific way by ASM library that we use. And seems that ASM looses `InnerClasses` attribute during this process - https://gitlab.ow2.org/asm/asm/blob/ASM_5_2/src/org/objectweb/asm/ClassWriter.java#L991-1006

Workarounds:
* reduce size of a method by reducing number of lambdas - as far as I can see for given example will be enough by 1
* exclude class from instrumentation

Should be noted that we tend to not implement any exclusions of methods during instrumentation for various reasons. One of them - such changes break compatibility of exec-file. And better to get fix in ASM. So will try to craft reproducer without JaCoCo and report this to ASM developers.

Godin:
Created bug report in ASM library:
http://forge.ow2.org/tracker/?func=detail&aid=317792&group_id=23&atid=100023

jvanheesch:
@Godin Thank you very much, both for the quick response and the ASM reproducer!
I will definitely keep an eye on that ticket.

Godin:
Fixed by #600

Issue #544, opened 2017-05-31
NullPointerException when instrumenting module-info.class


Godin:
Attempt to instrument `module-info.class` produced by JDK 9 EA b171 leads to

```
Exception in thread "main" java.io.IOException: Error while instrumenting /tmp/original/module-info.class.
	at org.jacoco.core.instr.Instrumenter.instrumentError(Instrumenter.java:165)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:117)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:141)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:159)
	at org.jacoco.core.instr.Instrumenter.instrumentAll(Instrumenter.java:196)
	at org.jacoco.cli.internal.commands.Instrument.instrument(Instrument.java:83)
	at org.jacoco.cli.internal.commands.Instrument.instrumentRecursive(Instrument.java:72)
	at org.jacoco.cli.internal.commands.Instrument.instrumentRecursive(Instrument.java:68)
	at org.jacoco.cli.internal.commands.Instrument.execute(Instrument.java:56)
	at org.jacoco.cli.internal.Main.execute(Main.java:89)
	at org.jacoco.cli.internal.Main.main(Main.java:104)
Caused by: java.lang.NullPointerException
	at org.objectweb.asm.Item.set(Item.java:231)
	at org.objectweb.asm.ClassReader.copyPool(ClassReader.java:368)
	at org.objectweb.asm.ClassWriter.<init>(ClassWriter.java:666)
	at org.jacoco.core.instr.Instrumenter$1.<init>(Instrumenter.java:79)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:79)
	at org.jacoco.core.instr.Instrumenter.instrument(Instrumenter.java:109)
```

Workaround is to exclude it from instrumentation, however there are complications:
* in case of CLI, Maven and Ant there is no way to exclude files when instrumenting archive files (#187), so might require repackaging of archive files
* in case of CLI there is no way to specify exclusions, so in addition to repackaging of archive files also requires copying of files

Godin:
Fixed by #600

PR #600, opened 2017-09-27
Upgrade to ASM 6

andrioli:
Upgrade to new ASM release 6.0. This version supports Java 9 class files, so I also removed Java9Support class.

This ASM version comes with a module-info.class file (major 53) but the shade plugin don't like it. So I also updated the maven-shade-plugin.

The shade plugin will output some warnings:
```
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
```
I tried to apply a filter but the filter have no effect removing the warn message. Looking in the bundled JAR the module-info.class  looks to not be included. So I think is ok to ignored the warn.

I built with success in my machine with Maven 3.5.0 and JDK 1.7.0_80, 1.8.0_144 and 9+181. Also tried with Maven 3.2.5 and JDK 1.6.0_65


org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
```diff
@@ -124,9 +123,8 @@ public void analyzeClass(final ClassReader reader) {
 	public void analyzeClass(final byte[] buffer, final String location)
 			throws IOException {
 		try {
-			analyzeClass(
-					new ClassReader(Java9Support.downgradeIfRequired(buffer)));
-		} catch (final RuntimeException cause) {
+			analyzeClass(new ClassReader(buffer));
+		} catch (final Exception cause) {
```
andrioli: Note that here I changed from RuntimeException to Exception

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
```diff
@@ -143,13 +141,11 @@ public void analyzeClass(final byte[] buffer, final String location)
 	 */
 	public void analyzeClass(final InputStream input, final String location)
 			throws IOException {
-		final byte[] buffer;
 		try {
-			buffer = Java9Support.readFully(input);
-		} catch (final IOException e) {
-			throw analyzerError(location, e);
+			analyzeClass(new ClassReader(input));
+		} catch (final Exception cause) {
```
andrioli: And here from IOException to Exception

marchof:
@andrioli Thanks for this nice and clean contribution!

@Godin Beside change log entry this PR looks good to me. Local IDE and builds are ok.

marchof:
@andrioli Also nice that you managed to open PR #600 for ASM 6. 😄

Godin:
@andrioli thank you for your contribution :thumbsup: and congratulations with round number 600 😉 

> Upgrade to new ASM release 6.0.

This is definitely good thing since should bring fixes for bugs reported by us.

@marchof I believe that we would need to add tests for them to prevent regressions since they are not tested in ASM explicitly. Also maybe will be better to do this update after finishing work on "filtering" - to ease its integration into Eclipse EclEmma, recall that it consumes ASM from Eclipse Orbit.

> The shade plugin will output some warnings:
> ```
> [WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
> ```
> I tried to apply a filter but the filter have no effect removing the warn message. Looking in the bundled JAR the module-info.class looks to not be included. So I think is ok to ignored the warn.

This is unavoidable warning, meaning that initially non-exported packages became visible for classes in final JAR - see mention of this in announcement of release of `maven-shade-plugin` version 3.1.0 (http://markmail.org/message/s6ogzz6h2j72xno6).

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: Wondering why? how this is related to ASM 6? I'm asking because we had some improvements and quite some nasty regressions/missed cases of exception handling exactly here.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: Also wondering why? and how this is related to ASM 6?

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
```diff
@@ -132,13 +124,11 @@ protected String getCommonSuperClass(final String type1,
 	 */
 	public byte[] instrument(final InputStream input, final String name)
 			throws IOException {
-		final byte[] bytes;
 		try {
-			bytes = Java9Support.readFully(input);
+			return instrument(new ClassReader(input));
```
Godin: If instrumentation fails, then exception won't have location information due to this usage of `instrument(ClassReader)` instead of previously used `instrument(byte[], String)`. Maybe simply preserve `readFully`?

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
```diff
@@ -156,7 +146,11 @@ protected String getCommonSuperClass(final String type1,
 	 */
 	public void instrument(final InputStream input, final OutputStream output,
 			final String name) throws IOException {
-		output.write(instrument(input, name));
+		try {
+			output.write(instrument(input, name));
+		} catch (final RuntimeException e) {
+			throw instrumentError(name, e);
```
Godin: IMO addition of this is an indication of regression in `instrument(InputStream, String)`.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
Godin: This reminds me again recurring old wish to deprecate `instrument(ClassReader)` or at least add a note.

marchof:
@Godin As Eclipse platform is working towards Java 9 I hope ASM 6 will show up in orbit soon (AFAIK ASM is required for PDE).

You're right about regression tests. Looks like the primary purpose of JaCoCo is to serve as a regression test suite for ASM and the JDK. 😉  

But this is not something we can request from @andrioli .

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
marchof: @Godin IMHO we can even just remove it. But it's a separate issue though.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
andrioli: Initially, I got some tests with errors (the test were expecting a exception with a actual message. but null were given). 

Later I noticed that the method `Instrumenter. instrument(InputStream...)` need to be wrapped with a try-catch block (noticed by looking the previous state in the git history).

But I didn't know if it was related

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
andrioli: Same comment here, I got a test with an exception (I think was an `ArrayIndexOutOfBoundsException` in this case)

Godin:
> As Eclipse platform is working towards Java 9 I hope ASM 6 will show up in orbit soon (AFAIK ASM is required for PDE).

@marchof Don't know specifically about PDE, but AFAIK for Oxygen.1a, whose focus is Java 9, there were no interest in ASM 6 BETA (see https://dev.eclipse.org/mhonarc/lists/orbit-dev/msg04882.html) and AFAIK final version was released too late for inclusion into Oxygen.1a.

But that's doesn't matter: by saying "ease integration", I meant exactly "easier", otherwise will just need to work on its addition as did for the two previous versions. There are quite some mess in sources of ASM published to Maven Central, but maybe I'll find time to deal with this and raise CQ for addition.

> You're right about regression tests. Looks like the primary purpose of JaCoCo is to serve as a regression test suite for ASM and the JDK. 😉

😆 

> But this is not something we can request from @andrioli .

This wasn't a request, just a remark/reminder for us.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
andrioli: @Godin I double checked my changes.  Looks that this can be reverted.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
andrioli: But reverting this change I got an `Failed tests:   testAnalyzeClass_BrokenStream(org.jacoco.core.analysis.AnalyzerTest): expected:<Error while analyzing BrokenStream.> but was:<null>`

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
andrioli: And if we don't wrap it with a try-catch block we got `Tests in error: testInstrumentAll_BrokenClassFileInZip(org.jacoco.core.instr.InstrumenterTest): 67` that is the ArrayOutOfBoundsException

Godin: @andrioli if you fix the regression in `instrument(InputStream, String)` described in https://github.com/jacoco/jacoco/pull/600#discussion_r141427042 , then test won't be failing and try-catch won't be needed here.

andrioli:
Done. @Godin 

Now is even easier to understand the changes.  Unfortunately Travis is failing. Looks that the JDK download URL is not valid. Maybe you can fix the environment variables and re-run the jobs to keep everything green 😄

Godin:
@andrioli Thanks! We'll merge this later.

@marchof I raised CQ for ASM 6 and it is already approved, so I'll raise PR for addition to Eclipse Orbit soon. However migration from Hudson to Jenkins for Orbit has been requested today, so might be delays in merges of PRs and builds.

org.jacoco.core/src/org/jacoco/core/internal/Files.java
```diff
@@ -0,0 +1,49 @@
...
+package org.jacoco.core.internal;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * File handling utilities
+ */
+public final class Files {
```
Godin: This is not just about files, because method inside this class works for any `InputStream`, so better to change name. I'll do this during rebase.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ResizeInstructionsTest.java
```diff
@@ -59,8 +58,8 @@ public void teardown() {
 	@Test
 	public void should_not_loose_InnerClasses_attribute() throws Exception {
```
Godin: After upgrade this test fails on JDK 5 due to another ASM bug - https://gitlab.ow2.org/asm/asm/issues/317800 😆

Godin: However ASM 5.2 already had an issues in a similar scenario - see #487

Godin: Given that scenarios are limited to class files without frames and with big methods that require wide jumps, and that there was problem #487 in the exact same scenarios, I do not consider this as regression.

In these scenarios after update of ASM, we'll be producing class files with frames that are not supposed to be present. While this can cause problems of reading of produced classes by the ASM as described in https://gitlab.ow2.org/asm/asm/issues/317800 , seems that JVM can read such classes and I believe that doesn't perform validation of frames, so this should not cause execution problems.

And taking into account that this update will resolve #487, #544, #584 and #606, I propose to update test to make it pass, and merge update into master.

Also seems that we can implement removal of these unwanted frames after instrumentation if there were no frames in original class - see my comment in ASM issue (https://gitlab.ow2.org/asm/asm/issues/317800#note_9609). However I propose to defer this and merge update without this - can implement later and maybe soon there will be ASM 6.0.1 with fix (https://gitlab.ow2.org/asm/asm/merge_requests/46).

@marchof WDYT?

marchof: @Godin Let's move forward to 6.0 without the workaround -- a you propose. I assume that 6.0.1 will be there before we get the first bug report for old class files.

PR #601, opened 2017-09-28
Yak shaving: add ability to merge coverage information of several instructions

Godin:
I propose to do one more baby step in story of filtering - add ability to merge coverage information of several instructions. This is non-functional change (in terms of visibility for users), but required for implementation of filter for `finally` blocks. This change separated from filter to allow focus on actual implementation of filter rather than on internal mechanics.


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -638,15 +641,84 @@ public void testTryCatchBlockFullyCovered() {
...
+		createTryFinally();
+		probes[0] = true;
+		probes[3] = true;
+		runMethodAnalzer();
+		assertEquals(4, nextProbeId);
+
+		assertLine(1001, 2, 4, 2, 2);
+	}
+
+	private final IFilter FILTER = new IFilter() {
```
marchof: Should this be a static constant? Or a instance member?

In any case I would prefer to name it specific for this test or make it a local variable in `testTryFinallyMergeSameBranch()`.

Godin: > or make it a local variable in `testTryFinallyMergeSameBranch()`

it is also used in `testTryFinallyMergeDifferentBranches`

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
```diff
@@ -103,10 +107,17 @@
 	public MethodAnalyzer(final String className, final String superClassName,
 			final String name, final String desc, final String signature,
 			final boolean[] probes) {
+		this(className, superClassName, name, desc, signature, probes, FILTERS);
+	}
+
+	MethodAnalyzer(final String className, final String superClassName,
```
marchof: Now as we have a instance member with the filters I don't think this class is the right set to maintain the list of all filters.

I would prefer to move the `FILTERS` constant to a new class `AllFilters` in the filter package.

Also there is no need to have a filter array: We can have a aggregate Filter which combines a list of filters into one `IFilter` instance. This simplifies the code in this class.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
marchof: Then `tryFinallyFilter`?

Godin: Now `static` and `TRY_FINALLY_FILTER`. BTW note that it merges only branching instruction, but not whole duplicate.

Godin:
> `MethodAnalyzerTest` misses the case when two merges instructions are already in the same set

In addition to `merge(i1, i2)` can simply do `merge(i2, i1)` and/or `merge(i1, i1)`, but maybe you're thinking about something different?

Strictly speaking, I would say that target of this test is computation of coverage in `visitEnd` and this test is at too high level and loses precision in terms of testing of operations with disjoint-sets and might be better to extract them and test separately, but not sure that this worth it at this stage.

> Maybe not part of this PR but `MethodAnalyzerTest` should also test `ignore()`

Let's do separately.

> Instead of `throw new AssertionError()` I would prefer JUnit's API `fail()` (also for consistency with the rest of the test base)

Done.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
Godin: > Also there is no need to have a filter array

Presence of array simplifies disabling of filters in `MethodAnalyzerTest`, otherwise will also need `NopFilter`.

Godin: @marchof extracted, let's see if can pass review now :laughing:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
```diff
@@ -0,0 +1,40 @@
...
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.tree.MethodNode;
+
+/**
+ * Filter that combines all other filters.
+ */
+public final class Filters implements IFilter {
+
+	public static final IFilter ALL = new Filters();
```
marchof: This introduces the only warning in core: public field needs JavaDoc

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -638,15 +641,95 @@ public void testTryCatchBlockFullyCovered() {
...
+	@Test
+	public void testTryFinallyMergeDifferentBranches() {
+		createTryFinally();
+		probes[0] = true;
+		probes[3] = true;
+		runMethodAnalzer(TRY_FINALLY_FILTER);
+		assertLine(1001, 2, 3, 0, 2);
+	}
+
+	private static final IFilter NOP_FILTER = new IFilter() {
```
marchof: Idea: Moving as `NONE` to `Filters`?

marchof:
@Godin Regarding test of `merge()`: I came across this as the line

    if (i1 != i2) {

has one branch missing. As the map based implementation of disjoint sets is quite tricky I thought it might be worth covering this case.

marchof:
@Godin What about dropping this constructor at all?

	public MethodAnalyzer(final String className, final String superClassName,
			final String name, final String desc, final String signature,
			final boolean[] probes) {
		this(className, superClassName, name, desc, signature, probes, Filters.ALL);
	}

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
Godin: Done.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
Godin: Done.

marchof:
@Godin May I propose a alternative implementation for `Filters` which looks a bit readable and flexible to me: https://gist.github.com/marchof/2ef153ad02b208c398a928808539eb0d

marchof:
@Godin Forget my last comment about alternative implementation of `Filters`. I think your implementation perfectly makes sense. Sorry.

Godin:
> Regarding test of merge(): I came across this as the line `if (i1 != i2)` has one branch missing. As the map based implementation of disjoint sets is quite tricky I thought it might be worth covering this case.

Quite a usual of my knowledge - maybe you remember that initially even comment was omitted :laughing: 

But you right that implementation is not so obvious - extraction will allow to express far more operations in tests than just merge of set with itself, e.g. `merge(a, c)`  given `a -> b` and `c -> d`.

Deferring this extraction, to cover branch added simple additional `merge` for now.

> What about dropping this constructor at all?
> ```
> public MethodAnalyzer(final String className, final String superClassName,
>		final String name, final String desc, final String signature,
>		final boolean[] probes) {
>	this(className, superClassName, name, desc, signature, probes, Filters.ALL);
> }
> ```

Done.

> May I propose a alternative implementation for Filters

Well, they are not much different.

> Forget my last comment about alternative implementation of Filters. I think your implementation perfectly makes sense. Sorry.

Ok. Keeping current comments as well?

> Thanks for addressing the issues! Looks much cleaner now. I have two more remarks in the new code though... sorry.

No need to be sorry - indeed much cleaner now! :thumbsup:

PR #603, opened 2017-09-29
Clean-up MethodAnalyzerTest

marchof:
As a follow up to #601 this PR seeks to clean-up the `MethodAnalyzerTest`:

* Use new naming conventions
* cleanup and separate existing tests
* new tests for use cases not yet covered

There are still some opcodes (i.e. `visit()` methods) which are not yet covered.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
@@ -57,12 +57,13 @@ public class MethodAnalyzerTest implements IProbeIdGenerator {
 		return nextProbeId++;
 	}
 
-	// === Scenario: linear Sequence without branches ===
+	// === Scenario: linear Sequence with and without ignore filtering ===
 
 	private void createLinearSequence() {
 		final Label l0 = new Label();
 		method.visitLabel(l0);
 		method.visitLineNumber(1001, l0);
+		method.visitInsn(Opcodes.NOP);
 		method.visitInsn(Opcodes.NOP);
 		final Label l1 = new Label();
 		method.visitLabel(l1);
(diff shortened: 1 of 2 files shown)


org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Constructor.java
```diff
@@ -36,7 +36,8 @@ private Inner() { // $line-inner$
 	}
 
 	private static class InnerStatic {
-		private Object field = this;
+		@SuppressWarnings("unused")
+		private final Object field = this;
```
Godin: I did cherry-pick of this change separately into master as it is unrelated to `MethodAnalyzerTest` and `MethodAnalyzer`.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -57,47 +57,83 @@ public int nextId() {
...
+		assertLine(1001, 0, 2, 0, 0);
+		assertLine(1002, 0, 1, 0, 0);
+	}
+
+	/** Filters consecutive NOP instructions as ignored */
+	private static final IFilter NOP_FILTER = new IFilter() {
+		public void filter(final String className, final String superClassName,
+				final MethodNode methodNode, final IFilterOutput output) {
+			AbstractInsnNode node = methodNode.instructions.getFirst();
+			while (node != null) {
```
Godin: IMO tests should require as little time as possible on understanding, so I would prefer to make this filter a way simpler.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -705,14 +807,46 @@ public void testTryFinallyMergeSameBranch() {
...
+		method.visitInsn(Opcodes.NOP);
+		method.visitInsn(Opcodes.NOP);
+		final Label l2 = new Label();
+		method.visitLabel(l2);
+		method.visitLineNumber(1001, l2);
+		method.visitInsn(Opcodes.RETURN);
+	}
+
+	@Test
+	public void decending_line_numbers_should_report_lines_correctly() {
```
Godin: Out of curiosity I'm wondering what motivated you to add this test?

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
marchof: @Godin Agreed. The target was to have a filter which filters at least two instructions (to specify a range). Any other ideas for a simplified filter? E.g. just by hard-coded index?

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
marchof: @Godin To get full branch coverage on `MethodAnalyzer.visitLineNumber()`. Any proposal for a different testing approach or improved test name?

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
Godin: @marchof two hard-coded indexes seems ok - this is already the case in this test in `TRY_FINALLY_FILTER`.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
Godin: @marchof fine like it is, I was just curious.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
marchof: @Godin Simplifed filter with hardcoded instruction positions.

marchof:
@Godin This is a new version, rebased on master without the unrelated fix of warning in `Constructor.java`.

PR #604, opened 2017-10-04
Add filter for duplicates of finally blocks

Godin:
Finally filter for `finally`, hopefully finally final version 😆


org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/TryWithResourcesTest.java
```diff
@@ -44,7 +44,7 @@ public void test() {
 		// without filter next line has branches:
 		assertLine("test.close", ICounter.EMPTY);
 		assertLine("test.catch", ICounter.NOT_COVERED);
-		assertLine("test.finally", ICounter.PARTLY_COVERED);
+		assertLine("test.finally", ICounter.FULLY_COVERED);
```
marchof: @Godin Just out of curiosity: The new FinallyFilter now also applies to tryWithResouces?

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,115 @@
...
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.filter.targets;
+
+import static org.jacoco.core.test.validation.targets.Stubs.ex;
+import static org.jacoco.core.test.validation.targets.Stubs.f;
+import static org.jacoco.core.test.validation.targets.Stubs.nop;
+
+public class Finally {
```
marchof: @Godin Can we add a nested finally here to see how it works?

org.jacoco.core.test/src-java7/org/jacoco/core/test/filter/TryWithResourcesTest.java
Godin: @marchof `javac` translates `try-with-resources` into `try-catch-finally`, so for it for sure applies. However most likely not applies for bytecode that `ecj` generates for closing of resources in `try-with-resources`. But applies for `try-with-resources` with `finally` for both `ecj` and `javac` as this test shows. Can add more precise tests about this, but definitely would prefer to do this in a separate pull request.

marchof: @Godin Ok, makes sense. I was wondering whether our validadtion tests for filters should only use the filter under test.

Godin: @marchof we already discussed this: for start it is very good to see combined behavior - this way with low investment we observe what end-user will observe and we observe that filters correctly cooperate and do not interfere with each other in a bad way. However I agree that in long run ability to execute filters separately will be also valuable.

marchof: @Godin Sure. For I don't want to change test strategy.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,196 @@
...
+
+					i = next(i);
+					if (continues && !inside.contains(i)) {
+						merge(size, e, i);
+					}
+				}
+
+				if (t != finallyBlock && t.start == finallyBlock.start
+						&& t.end == finallyBlock.end) {
+					final AbstractInsnNode i = next(next(t.handler));
```
marchof: @Godin Potential NPE? `next(t.handler)` could possibly return `null`

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,196 @@
...
+			size++;
+			i = next(i);
+			if (i == null) {
+				// when always completes abruptly
+				return 0;
+			}
+		} while (!(Opcodes.ALOAD == i.getOpcode()
+				&& var == ((VarInsnNode) i).var));
+		i = next(i);
+		if (Opcodes.ATHROW != i.getOpcode()) {
```
marchof: @Godin Potential NPE? `i` could be `null` here -- but only for invalid class files. So probably not an issue.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
Godin: @marchof `t.handler` is a label that must have instruction after it, so `next(t.handler)` can't be `null`.

`next(next(t.handler))` potentially can be `null` only if handler has exactly one instruction that is the last one in method, but in this case safe to call `merge(size, e, null)` that safely calls `isSame(size, e, null)` and returns.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: @marchof done

marchof: @Godin Thanks! :white_check_mark:

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,129 @@
...
+		} finally { // $line-emptyTry.0$
+			nop(); // $line-emptyTry.1$
+		} // $line-emptyTry.2$
+	}
+
+	private static void alwaysCompletesAbruptly() {
+		try {
+			nop();
+		} finally { // $line-alwaysCompletesAbruptly.0$
+			return; // $line-alwaysCompletesAbruptly.1$
```
marchof: @Godin Here I get a warning in Eclipse, maybe annotate method with `@SuppressWarnings("finally")`?

marchof:
@Godin Looks like the latest update broke some tests.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
```diff
@@ -0,0 +1,196 @@
...
+	public void filter(final String className, final String superClassName,
+			final MethodNode methodNode, final IFilterOutput output) {
+		for (final TryCatchBlockNode tryCatchBlock : methodNode.tryCatchBlocks) {
+			if (tryCatchBlock.type == null) {
+				new Matcher(output, methodNode, tryCatchBlock).filter();
+			}
+		}
+	}
+
+	private static class Matcher {
```
marchof: @Godin The only state in `Matcher` class is member `inside`, which is used only locally within method `filter()`. Therefore should be better al local variable within this method. Then the `Matcher` class can be removed at all. Tried this here: https://gist.github.com/marchof/5adb1930188136c3ea340ff430a2cac5

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
```diff
@@ -0,0 +1,420 @@
...
+	 *       ...
+	 *       return;
+	 *     }
+	 *   } finally {
+	 *     ...
+	 *   }
+	 * </pre>
+	 */
+	@Test
+	public void should_analyze_control_flow() {
```
marchof: @Godin While I'm trying to wrap my mind around the filter implementation I was wondering why we need to loop over all tryCatchBlocks with the same handler. This seems to be the only test case which triggers this. Therefore I wonder whether it might make sense to add this scenario also to the validation test.

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
```diff
@@ -0,0 +1,129 @@
...
+				nop("finally"); // $line-breakStatement.1$
+			} // $line-breakStatement.2$
+		}
+	}
+
+	private static void catchNotExecuted() {
+		try {
+			nop("try");
+		} catch (Exception e) { // $line-catchNotExecuted$
+			nop("catch");
```
marchof: @Godin Add assertion that this line is `NOT_COVERED`? To ensure that no finally code sneaks in here.

Godin:
@marchof 
> Looks like the latest update broke some tests.

Ouch, forgot to update test that studies placement of `GOTO`s. Fixed.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
Godin: @marchof done

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: @marchof done

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
Godin: @marchof done

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
Godin: @marchof added this suppression, however warnings and suppressions are incredibly inconsistent in IDEs, e.g. even after addition I see warnings in this target file and in some other target files in IntelliJ. Ideally, I think, that our main code should not contain warnings and suppressions, but should be possible to write whatever in targets without bothering with warnings/suppressions.

marchof: We could simply add `@SuppressWarnings("all")` to our validation targets. Should work for all IDEs.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
marchof: @Godin When I remove the two loops like this:

		// for (final TryCatchBlockNode t : tryCatchBlocks) {
		{
			final TryCatchBlockNode t = finallyBlock;

the validation tests are still fully green (ECJ) and a different test fails for javac:

    emptyCatch(org.jacoco.core.test.filter.FinallyTest): Status in line 79:	nop("finally"); // $line-emptyCatch.1$ expected:<[FUL]LY_COVERED> but was:<[PART]LY_COVERED>

But new test `twoRegions()` stays green. What am I missing here?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java
marchof: @Godin Thanks! :white_check_mark:

org.jacoco.core.test/src/org/jacoco/core/test/filter/targets/Finally.java
marchof: @Godin Thanks! :white_check_mark:

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
```diff
@@ -23,17 +23,6 @@
  */
 public class ExceptionsTest extends ValidationTestBase {
 
-	/**
-	 * https://bugs.openjdk.java.net/browse/JDK-8180660
-	 */
-	private static final boolean isJDK8u152;
-
-	static {
-		final Matcher m = Pattern.compile("1\\.8\\.0_(\\d++)(-ea)?")
```
marchof: @Godin Deletion leads to unused imports warning for `java.util.regex.*`.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
```diff
@@ -0,0 +1,420 @@
...
+		assertEquals("ignored", toIndexes(expectedIgnored),
+				toIndexes(actualIgnored));
+		assertEquals("merged", toIndexes(expectedMerged),
+				toIndexes(actualMerged));
+	}
+
+	private Set<Integer> toIndexes(Set<AbstractInsnNode> set) {
+		final Set<Integer> result = new HashSet<Integer>();
+		for (final AbstractInsnNode i : set) {
+			result.add(m.instructions.indexOf(i));
```
marchof: @Godin Eclipse shows a boxing warning here.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
Godin: @marchof after this modification of both loops
* `emptyCatch` fails because condition of `if` statement in second loop always `false`
* unit test for the case of `twoRegions` shows that merged too much, but validation test has assertion only for `final`, for same reason `ecj` stays green

Missing assertions added.

Also this test demonstrates https://bugs.openjdk.java.net/browse/JDK-7008643

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
Godin: @marchof fixed

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
Godin: @marchof if replaced by `Integer.valueOf(m.instructions.indexOf(i)))`, then no warning in Eclipse, but warning in IntelliJ with suggestion to remove unnecessary boxing, so added `@SuppressWarnings("boxing")` to this method

org.jacoco.core.test/src/org/jacoco/core/test/validation/ExceptionsTest.java
marchof: @Godin Thx! ✅

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FinallyFilterTest.java
marchof: @Godin Ok. But should give you many warnings at other places then in IntelliJ? ✅

Godin: @marchof yes, I have quite some warnings in other places in IntelliJ, but we can put this aside

Godin:
@marchof description and comments added, could you please re-review? If all fine, I'll squash and merge.

Issue #545, opened 2017-05-31
IllegalAccessError when executing Java 9 modules after offline instrumentation


Godin:
```
$ cat src/module-info.java
module org.example {
}

$ cat src/org/example/HelloWorld.java
package org.example;

class HelloWorld {
  public static void main(String[] args) {
    System.out.println("Hello, World!");
  }
}

$ javac -d mods/org.example \
        src/module-info.java \
        src/org/example/HelloWorld.java
```

```
$ java -jar jacococli.jar \
       instrument mods/org.example/org \
       --dest mods-instrumented/org.example/org
```

```
$ java \
       -cp mods-instrumented/org.example:jacocoagent.jar \
       org.example.HelloWorld
Hello, World!
```

```
$ cp mods/org.example/module-info.class mods-instrumented/org.example

$ java --version
java 9-ea
Java(TM) SE Runtime Environment (build 9-ea+171)
Java HotSpot(TM) 64-Bit Server VM (build 9-ea+171, mixed mode)

$ java \
       --module-path mods-instrumented
       -cp jacocoagent.jar
       --module org.example/org.example.HelloWorld
```

leads to

```
Exception in thread "main" java.lang.IllegalAccessError: class org.example.HelloWorld (in module org.example) cannot access class org.jacoco.agent.rt.internal_c42d03f.Offline (in unnamed module @0x15975490) because module org.example does not read unnamed module @0x15975490
	at org.example/org.example.HelloWorld.$jacocoInit(HelloWorld.java)
	at org.example/org.example.HelloWorld.main(HelloWorld.java)
```

as well as

```
java \
       --module-path mods-instrumented \
       -javaagent:jacocoagent.jar=excludes=org.example.HelloWorld \
       -m org.example/org.example.HelloWorld
```

Godin:
Some thoughts:

* If agent will have an explicit module name, then maybe requirement of an explicit `--add-reads` will not be a big issue even after Java 9 GA. But to address this question we need to keep an eye on integration with build tools.

* Offline instrumentation can update `module-info.class`, but this requires changes in/for ASM. And this will not work in cases when for some reasons instrumentation of `module-info.class` was not performed.

[...]

PR #608, opened 2017-10-08
Add test of offline instrumentation of Java 9 module-info

Godin:


marchof:
@Godin I wonder how this test relates to the Maven plug-in. It's more like a JaCoCo code issue. Maybe it is time to create a new source folder

    org.jacoco.core.test/src-java9

and move the test there?

Godin:
@marchof you're right that test is not directly related to Maven plugin, however we can't simply add `module-info.java` into `org.jacoco.core.test` because then compilation will fail with many errors - for example one of them
```
org.jacoco.core.test/src/org/jacoco/core/test/validation/SourceTest.java:[14,24] package org.junit does not exist
```
i.e. dependencies should be modules too, which is not the case at least for `junit`, not counting other issues with modularization such as "package exists in another module: org.jacoco.core".

Also runtime issues such as #545 won't be visible if we'll verify only offline instrumentation of `module-info.class` without execution, whereas I planned to update this end-to-end test to also demonstrate #545 and how to deal with it in Maven.

So not counting commit of already compiled `module-info` that I don't like for three reasons: addition of binary file, will not be testing further versions of compiler and will not be testing end-to-end user experience that includes execution. The only simple way to test instrumentation of compiled `module-info` that I found - is to add integration test of Maven plugin.

Godin:
@marchof 

> I wonder whether there should be some some verification? Or is it that there is no error?

without
```xml
              <!-- FIXME remove this workaround after update of ASM to 6.0 -->
              <excludes>
                <exclude>module-info.class</exclude>
              </excludes>
```
in `pom.xml` build will fail with `NullPointerException` as described in #544. Thus, yes, verification - is that build completes successfully.

Godin:
To be even more precise: currently this test doesn't demonstrate #545, because use of `--module-path` not yet implemented in `maven-surefire-plugin` (https://issues.apache.org/jira/browse/SUREFIRE-1262), but as was said before - demonstrates #544.

marchof:
@Godin Ok, thanks for the explanation!

PR #605, opened 2017-10-09
Unclosed streams in CoreTutorial

marchof:
Will be better to clarify that caller is responsible for closing of streams passed to methods of `Analyzer` and `Instrumenter` - at least by fixing this in `CoreTurorial` example, and ideally also by update of Javadocs.


Godin:
LGTM. Thanks!

PR #610, opened 2017-10-10
Filter for Groovy generated methods

datoma:
Hackergarten Bern Session, co-hacked with @aalmiray and @marchof  

highres_465229986

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
@@ -1,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2017 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Marc R. Hoffmann - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import java.util.List;
+
+import org.objectweb.asm.tree.AnnotationNode;
+import org.objectweb.asm.tree.MethodNode;
+
+/**
+ * Filters annotated methods.
+ */
+abstract class AbstractAnnotatedMethodFilter implements IFilter {
+	private final String descType;
+
+	/**
+	 * Configures a new filter instance.
+	 * 
+	 * @param annotationType
+	 *            VM type of the annotation
+	 */
+	protected AbstractAnnotatedMethodFilter(final String annotationType) {
+		this.descType = "L" + annotationType + ";";
+	}
+
+	public void filter(final String className, final String superClassName,
+			final MethodNode methodNode, final IFilterOutput output) {
+		if (hasAnnotation(methodNode)) {
+			output.ignore(methodNode.instructions.getFirst(),
+					methodNode.instructions.getLast());
+		}
+	}
+
+	private boolean hasAnnotation(final MethodNode methodNode) {
+		final List<AnnotationNode> annotations = getAnnotations(methodNode);
+		if (annotations != null) {
+			for (final AnnotationNode annotation : annotations) {
+				if (descType.equals(annotation.desc)) {
+					return true;
+				}
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Retrieves the annotations to search from a method. Depending on the
+	 * retention of the annotation this is either
+	 * <code>visibleAnnotations</code> or <code>invisibleAnnotations</code>.
+	 * 
+	 * @param methodNode
+	 *            method to retrieve annotations from
+	 * @return list of annotations
+	 */
+	abstract List<AnnotationNode> getAnnotations(final MethodNode methodNode);
+
+}
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -35,8 +35,12 @@
       try-with-resources statement
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/500">#500</a>).</li>
   <li>Exclude from a report methods which are annotated with <code>@lombok.Generated</code>.
-      Initial analysis and contribution by Rüdiger zu Dohna.
+      Initial analysis and contribution by Rüdiger zu Dohna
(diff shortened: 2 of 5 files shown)


aalmiray:
Here's the Groovy side of the equation https://github.com/apache/groovy/pull/617

marchof:
@aalmiray Thanks! Two questions:

- Is there a build or SNAPSHOT available of this Branch so we can test integration? JaCoCo build of this branch is [here](https://ci.appveyor.com/project/JaCoCo/jacoco/build/1.0.1442/artifacts).
- What is the Groovy version which will include this feature? So we can make a reference to the required version in our change log.

Godin:
@datoma @aalmiray cool! thank you for contribution! 👍 

@aalmiray impatient to try full end-to-end example, so same questions as from @marchof plus maybe you can advise some Groovy-based project where JaCoCo is already used, so that we'll be able to use it as such end-to-end example? And in any case it can be a nice addition for page https://github.com/jacoco/jacoco/wiki/Projects-that-use-JaCoCo 😉 

Also wondering if you know examples of generated code that might require filtering and is not represented by generated methods, but is inside of non-generated methods?

aalmiray:
@marchof this feature currently targets Groovy 2.6.0

@Godin I use JaCoCo on the Griffon project. Trust me when I say I'll be the first to try out this new feature 😄

marchof:
@Godin Groovy side has been merged on master 2_5_x and 2_6_x. Can you review our side?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
```diff
@@ -0,0 +1,52 @@
...
+	public void filter(final String className, final String superClassName,
+			final MethodNode methodNode, final IFilterOutput output) {
+		if (hasAnnotation(methodNode)) {
+			output.ignore(methodNode.instructions.getFirst(),
+					methodNode.instructions.getLast());
+		}
+	}
+
+	private boolean hasAnnotation(final MethodNode methodNode) {
+		final List<AnnotationNode> runtimeInvisibleAnnotations = methodNode.invisibleAnnotations;
```
Godin: In contrast to Lombok, Groovy adds visible annotation, while here we check only `invisibleAnnotations`, so that currently filter doesn't work.

marchof: @Godin Excellent catch! Fixed ✅

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -37,6 +37,9 @@
   <li>Exclude from a report methods which are annotated with <code>@lombok.Generated</code>.
       Initial analysis and contribution by Rüdiger zu Dohna.
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/513">#513</a>).</li>
+  <li>Exclude from a report methods which are annotated with <code>@groovy.transform.Generated</code>
+      to better integrate with Groovy >= 2.6.0. Thanks to Andres Almiray to make this possible!
```
Godin: I'm not a native English speaker, but "**to** make this possible" looks/sounds weird for me.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
```diff
@@ -0,0 +1,55 @@
...
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import java.util.List;
+
+import org.objectweb.asm.tree.AnnotationNode;
+import org.objectweb.asm.tree.MethodNode;
+
+/**
+ * Filters which filters annotated methods.
```
Godin: "Filter**s** which filters" ?

Godin: Also I'm wondering why abstracting instead of having simply one class that will handle both annotations?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
Godin: @marchof nothing extraordinary - I'm simply doing end-to-end testing ;)

org.jacoco.doc/docroot/doc/changes.html
Godin: Also I believe that `>= 2.6.0` should be replaced by `>= 2.5.0` since change on Groovy side was merged not just into `master` and `2_6_x` branches, but also into `2_5_x`.

bjkail: I guess "Thanks to Andres Almiray **for making** this possible!" would work, but perhaps just rephrase to match the preceding bullet or say "for the contribution" or something.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractAnnotatedMethodFilter.java
marchof: @Godin I would prefer to have a separate class for every filter:

* Separate testing of different concerns
* Possibly enable/disable filters separately
* I could imagine that we add meta methods/annotations to filters for generating documentation

org.jacoco.doc/docroot/doc/changes.html
marchof: @bjkail Thanks for your input! What I want to express here that Andres triggered the introduction of the annotation in the Groovy compiler.

bjkail: Ah, then "for making this possible" would work, but "this" seems ambiguous, so I would say "for adding the annotation to Groovy, which made this possible".

marchof: @Godin @bjkail Fixed wording and version range :white_check_mark:

Godin:
End-to-end test - [example.zip](https://github.com/jacoco/jacoco/files/1392543/example.zip)
Note that class is `abstract`, otherwise generated and annotated methods will be marked as synthetic and thus filtered even without this change.

Without this change using JaCoCo 0.7.9:
before

After this change:
after

aalmiray:
Have to check again with @blackdrag why the `GroovyObjectMethods` are marked as synthetic if the class is not abstract (which it shouldn't be).

Godin:
@aalmiray well, in any case they are filtered :wink: I was more wondering why generated setters and getters do not have nor line numbers (was expecting line of a field), nor annotation? But that's not so important for end-to-end test here and now that's entirely up to you, Groovy-guys, to mark whatewer you want :wink: Thank again! :thumbsup: Great collaboration! :heart:

PR #618, opened 2017-11-03
Cleanup AgentTest

marchof:
* Don't write to file system
* Avoid DNS lookups (slows down builds)
* Use new test naming conventions

### DIFF
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -146,14 +147,20 @@ public class AgentTest implements IExceptionLogger, IAgentOutput {
 		agent.shutdown();
 	}
 
-	@Test(expected = InstanceNotFoundException.class)
+	@Test
 	public void startup_should_not_register_mbean_when_disabled()
 			throws Exception {
 		Agent agent = createAgent();
+
 		agent.startup();
 
 		ObjectName objectName = new ObjectName("org.jacoco:type=Runtime");
-		ManagementFactory.getPlatformMBeanServer().getMBeanInfo(objectName);
+
+		try {
+			ManagementFactory.getPlatformMBeanServer().getMBeanInfo(objectName);
+			fail("InstanceNotFoundException expected");
+		} catch (InstanceNotFoundException e) {
+		}
 	}
 
 	@Test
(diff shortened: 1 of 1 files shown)


org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
```diff
@@ -16,176 +16,239 @@
...
+		ObjectName objectName = new ObjectName("org.jacoco:type=Runtime");
+		final MBeanServer server = ManagementFactory.getPlatformMBeanServer();
+		assertEquals(JaCoCo.VERSION,
+				server.getAttribute(objectName, "Version"));
 
+		// cleanup as MBean is registered globally
+		agent.shutdown();
+	}
+
+	@Test(expected = InstanceNotFoundException.class)
```
Godin: I prefer `try` over `@Test(expected = ... )` as it allows more precisely specify place where exception will be thrown.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
```diff
@@ -16,176 +16,239 @@
...
+		};
 		agent.startup();
 
-		assertTrue(exception instanceof IOException);
+		agent.shutdown();
+
+		assertSame(expected, loggedException);
+	}
+
+	@Test(expected = InstanceNotFoundException.class)
```
Godin: Same.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
```diff
@@ -16,176 +16,239 @@
...
+	public void shutdown_should_deregister_mbean_when_enabled()
+			throws Exception {
+		options.setJmx(true);
+		Agent agent = createAgent();
+		agent.startup();
+
+		agent.shutdown();
+
+		ObjectName objectName = new ObjectName("org.jacoco:type=Runtime");
+		ManagementFactory.getPlatformMBeanServer().getMBeanInfo(objectName);
```
Godin: Combine with `startup_should_register_mbean_when_enabled` ?

marchof: I personally prefer to test one target/aspect at a time in unit tests. In case of test failures you see which aspects are broken.

marchof:
@Godin Addressed exception checking.

As a side note: I added a new commit for the review issues. During the filtering stories I sometimes got lost what has been addressed. As we squash anyways (and GitHub should allow to remove the additional commit messages) I try to make it more traceable during the review.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentTest.java
Godin: @marchof in case of failure of MBean startup or shutdown, both tests most likely will be failing, and that's why I asked about combining 😉

marchof: @Godin Sure. Mostlikely. But could also be that registration works but deregistration does not. I simply want to avoid `should_do_this_and_that` unit tests 😉

Issue #265, opened 2014-12-16
Probe inserted at the beginning of a try-catch block should be included in the block.


allenhair:
We insert a probe before a label if it is a successor and also a target. If the label is also used to define the start of an try-catch block, the inserted probe will not be included in the block.

This causes some issues when dealing with synchronized blocks. Normally, the compiler adds a catch-all entry to the exception table which covers the synchronized block in order to call monitorexit if any exceptions are thrown. If we need to insert a probe before the label which is used to define the start of the try-catch block, the catch-all entry will no longer cover all code between the monitorenter and monitorexit calls. While the probe is unlikely to cause an exception, this approach is not strictly correct.

We should probably split labels if they define a jump target and the start of a try-catch block. That way, the probe can be inserted after the start of the try-catch block, but before the jump target.

marchof:
Fixed on master.

PR #627, opened 2017-12-02
End try-catch blocks before inserted probes.

### LINKED ISSUE #626: Unbalanced monitorenter / monitorexit instructions in instrumented bytecode.
Jacoco will sometimes insert a probe before the end label of a try-catch block. When this happens, the probe will be covered by the range of the catch clause. This can lead to code paths where monitorexit can be called twice.

For example, consider the following:
```
  Code:
          ...
     9: monitorenter
    10: iload_1
    11: ifeq          23
    14: aload_2
    15: monitorexit
    16: aload_0
          ...
    34: astore_3
    35: aload_2
    36: monitorexit
    37: aload_3
          ...
  Exception table:
     from    to  target type
        10    16    34   any
        34    37    34   any
```

The range of the catch clauses are inclusive on the "from" end and exclusive on the "to" end, so that all instructions after monitorenter are covered, including the final monitorexit (but not the following aload_0). If any of these throw, we'll jump to PC 34 so that monitorexit is executed no matter what.

However, when instrumented, we get the following:
```
  Code:
          ...
    13: monitorenter
    14: aload_2
    15: iconst_3
    16: iconst_1
    17: bastore
    18: iload_1
    19: ifne          29
    22: aload_2
    23: iconst_4
    24: iconst_1
    25: bastore
    26: goto          47
    29: aload_3
    30: monitorexit
    31: aload_2
    32: iconst_5
    33: iconst_1
    34: bastore
    35: aload_0
          ...
    58: astore        4
    60: aload_3
    61: monitorexit
    62: aload         4
          ...
  Exception table:
     from    to  target type
        14    35    58   any
        58    62    58   any
```

In this case, the probe inserted before the aload_0 (PC 35) instruction becomes covered by the catch clause. As a result, it is possible to execute monitorexit twice (on PC 30, and then again on PC 61) if any of the probe instructions (PC 31-34) throw.

### Steps to reproduce

JaCoCo version: 0.7.10-SNAPSHOT
Operating system: Linux
Tool integration: CLI

Compile and instrument the attached [Check.java](https://github.com/jacoco/jacoco/files/1523374/Check.java.txt)
class, then inspect the bytecode:
1. javac Check.java
2. java -jar jacococli.jar instrument Check.class --dest instrumented
3. javap -c instrumented/Check.class

### Expected behaviour

The catch clause does not cover any instructions after monitorexit. That is, the exception table entry should be:
```
  from    to  target type
     14    31    58   any
```

### Actual behaviour

The exception table entry covers the probe instructions that occur after monitorexit succeeds. The exception table entry is:
```
  from    to  target type
     14    35    58   any
```

allenhair:
Fixes #626


org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java
```diff
@@ -28,7 +28,7 @@
 
 	static {
 		// Runtime initialize to ensure javac does not inline the value.
-		FORMAT_VERSION = 0x1007;
+		FORMAT_VERSION = 0x1008;
```
Godin: I'm wondering whether version increment is really required, if the only change - is an end of try-catch blocks and not number and placement of probes? Otherwise I'm afraid of consequences of this incompatible change for downstream projects as already mentioned in https://github.com/jacoco/jacoco/issues/626#issuecomment-349073113 - when we did increment last time several years ago, it was quite disruptive.

Also in backlog we have some other changes that require increment, but were considered minor to justify it. So if underlying issue here justifies increment, then we'd better do them all together.

allenhair: To be honest, I'm not quite sure whether the increment is necessary. I remember @marchof said it was needed for [#265](https://github.com/jacoco/jacoco/issues/265#issuecomment-68149109) but you are correct that the number of probes and their placement doesn't really change.

Godin: Actually not obvious that number and placement do not change - I was not and is not sure about this and that was a reason to ask.

Also I have feeling that our unit tests unfortunately are not exhaustive in regards to number and placement of probes and especially in case of complex CFGs involving try-catch blocks, and unfortunately we don't have assertions about number of probes in validation tests and they are probably not exhaustive too.

And since you're also unsure, we'll need to study this more carefully, e.g. by counting number of probes on a big set of class files before and after this change.

Godin: @allenhair in fact after merge of #266 @marchof reverted increment of version - 82f40d7151ef0f99e18573a4e6b248eba6abc3ad :wink:

And here I come to following conclusions: move of start and end labels of try-catch blocks between end and beginning of probe doesn't affect original labels and so doesn't affect MethodInstrumenter, i.e. placement and number of probes, as well as MethodAnalyzer, i.e. result of analysis, and affects only final class after instrumentation. I guess this is exactly what is described as "leads to semantically identical probes" in 82f40d7151ef0f99e18573a4e6b248eba6abc3ad.

@marchof Am I right? WDYT?

marchof: @Godin @allenhair If the change in #266 did not change the number of probes this new handling for the end labels will also not change the number of probes. So no need to change exec version.
Also if we would do so we should combine this with other changes in our backlog (see @Godin 's comment).

org.jacoco.core.test/src/org/jacoco/core/internal/flow/MethodProbesAdapterTest.java
```diff
@@ -320,52 +320,87 @@ public void testVisitTryCatchBlockNoProbe() {
...
+		adapter.visitTryCatchBlock(start, end, handler2, "java/io/IOException");
+		adapter.visitLabel(start);
+		adapter.visitInsn(Opcodes.NOP);
+		adapter.visitLabel(end);
 
-		expectedVisitor.visitTryCatchBlock(start, end, handler1,
+		expectedVisitor.visitTryCatchBlock(probe, end, handler1,
 				"java/lang/Exception");
-		expectedVisitor.visitTryCatchBlock(start, end, handler2,
+		expectedVisitor.visitTryCatchBlock(probe, end, handler2,
```
Godin: This test verifies that in case of probe at the beginning two invocations of `visitTryCatchBlock` will use the same first label, i.e. second invocation uses label created during first invocation. But there is no test to verify second label in case of probe at the end.

Godin:
@marchof I restored version of exec-file, polished tests and comments and added changelog entry. Could you please re-review? If all fine, I'll squash and merge.

Godin:
@allenhair hope you don't mind that we took over finalization of this, thanks again for your valuable contribution! ❤️

allenhair:
No problem @Godin, thanks for pushing this through.

PR #634, opened 2017-12-26
Offline instrumentation should not damage module-info

Godin:
Given

```
$ mkdir -p src/org/example

$ cat <<END > src/module-info.java
module org.example {
}
END

$ cat <<END > src/org/example/HelloWorld.java
package org.example;

class HelloWorld {
    System.out.println("Hello, World");
  }
}
END

$ javac -d mods/org.example \
        src/module-info.java \
        src/org/example/HelloWorld.java
```

after #600 we are able to instrument `module-info.class`

```
$ java -jar jacococli.jar \
       instrument mods \
       --dest mods-instrumented
```

and after #565 execute instrumented module using `--add-reads`

```
$ java \
       --module-path mods-instrumented:jacocoagent.jar \
       --add-modules org.jacoco.agent.rt \
       --add-reads org.example=org.jacoco.agent.rt \
       --module org.example/org.example.HelloWorld
```

however test #608 lacks such execution and following exception went unnoticed

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Error reading module: mods-instrumented/org.example
Caused by: java.lang.module.InvalidModuleDescriptorException: Bad #fields
```

cause of exception

```
$ javap -p mods-instrumented/org.example/module-info.class
Compiled from "module-info.java"
module org.example {
  requires java.base;
  private static transient boolean[] $jacocoData;
  private static boolean[] $jacocoInit();
}
```

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactoryTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactoryTest.java
@@ -211,6 +211,19 @@ public class ProbeArrayStrategyFactoryTest {
 				true, 0);
 	}
 
+	@Test
+	public void testModule() {
+		final ClassWriter writer = new ClassWriter(0);
+		writer.visit(Opcodes.V9, Opcodes.ACC_MODULE, "module-info", null, null,
+				null);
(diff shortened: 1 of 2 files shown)


Godin:
@marchof could you please review? If all fine, I'll squash and merge.

PR #636, opened 2017-12-29
Restore exec files compatibility regarding Java 9 class files

Godin:
https://groups.google.com/d/msg/jacoco-dev/uPwM10Ia4Ks/x4CC4hd0CwAJ

```
$ cat <<END > Example.java
class Example {
  }
}
END

$ javac -source 9 -target 9 Example.java

$ java -javaagent:jacoco-0.7.9/lib/jacocoagent.jar=includes=Example,destfile=1.exec Example

$ java -javaagent:jacoco-0.7.10-20171220.225235-84/lib/jacocoagent.jar=includes=Example,destfile=2.exec Example

$ java -jar jacoco-0.7.10-20171220.225235-84/lib/jacococli.jar execinfo 1.exec 2.exec
[INFO] Loading exec file 1.exec.
CLASS ID         HITS/PROBES   CLASS NAME
Session "Godins-iMac.lrc.sonarsource.com-2c84a6d3": Thu Dec 21 20:53:29 CET 2017 - Thu Dec 21 20:53:29 CET 2017
2cc2366f70618d57    1 of   2   Example

[INFO] Loading exec file 2.exec.
CLASS ID         HITS/PROBES   CLASS NAME
Session "Godins-iMac.lrc.sonarsource.com-4cd2ef2e": Thu Dec 21 20:53:24 CET 2017 - Thu Dec 21 20:53:24 CET 2017
d9c2366f35618d57    1 of   2   Example
```

before #600 computation of Java 9 class identifier was happening after modification of version from 53 on 52 in bytecode,  this was unfortunately overlooked/underestimated in #406.

---

Even if our changelog only claimed "experimental Java 9 support", this support [leaked into downstream projects](https://jira.sonarsource.com/browse/SONARJAVA-1716). So change of class identifiers will affect at least users of SonarQube and Jenkins plugins in a way similar to change of exec-file version, but limited by Java 9 class files: in order to get report for Java 9 class files will be required to produce exec file with the same version of JaCoCo that is embedded in SonarQube and Jenkins plugins.

In case of a real change of exec file version such situation will be easy to diagnose, but this will affect all users and not only ones who use Java 9 class files, which will negatively impact adoption of JaCoCo version 0.8.0.

And maybe will be hard to diagnose such situation in absence of a real change of exec file version and given absence of messages about class versions and [mismatch of class IDs in SonarQube](https://jira.sonarsource.com/browse/SONARJAVA-2601) and Jenkins.

So for now I would prefer to keep compatibility and to not change Java 9 class IDs.

Code to keep compatibility is well localized, can be removed after change of exec file version or maybe even just after JDK 10 release - according to http://www.oracle.com/technetwork/java/eol-135779.html Java 9 is not an LTS release and users advised to immediately transition to the next version once it will be released:

> Java SE 9 will be a short term release, and users should immediately transition to the next release

as another example - https://www.elastic.co/blog/this-week-in-elasticsearch-and-apache-lucene-2017-12-18 :

> When JDK 9 is end-of-life in March 2018, releases of Elasticsearch will stop supporting JDK 9

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
@@ -15,7 +15,10 @@ import static org.junit.Assert.assertEquals;
 
 import java.io.UnsupportedEncodingException;
 
+import org.jacoco.core.data.ExecutionDataWriter;
 import org.junit.Test;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
 
 /**
  * Unit tests for {@link CRC64}.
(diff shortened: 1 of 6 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
```diff
@@ -15,16 +15,43 @@
...
 /**
  * Unit tests for {@link CRC64}.
  */
 public class CRC64Test {
 
+	@Test
+	public void testJava9() {
+		// should remove workaround for Java 9
+		// during change of exec file version
+		assertEquals(0x1007, ExecutionDataWriter.FORMAT_VERSION);
```
marchof: Automatic reminder. Nice.

PR #644, opened 2018-02-01
Simplify conditions that use "java.version" property in tests

Godin:


Godin:

namannigam:
For JDK9+, you might want to look at the implementation of [`Runtime.Version`](https://docs.oracle.com/javase/9/docs/api/java/lang/Runtime.Version.html).  Here's the doc for [10-ea](http://download.java.net/java/jdk10/docs/api/java/lang/Runtime.Version.html) with further updates to the APIs exposed.

marchof:
@Godin Nice! I wonder we should encapsulate more version logic in the new utility:

- Moving `ValidationTestBase.JAVA_VERSION` to `JavaVersion.CURRENT`
- Adding comparison Methods to `JavaVersion`
- (Implementation Detail: Make `JavaVersion` `Comparable`)?

Then the test code could look like like:

    if (JavaVersion.CURRENT.isBefore(8, 152)) ...

or maybe this reads better:

    if (JavaVersion.CURRENT.isBefore("1.8.0_152")) ...

What do you think?

marchof:
@namannigam Thanks for the pointer! Just remember that the JaCoCo code base is still compatible with Java 1.5 ;)

marchof:
@Godin As a side note. I just realized that we actually use the version in two contexts:

- Java runtime version (`BadCycleInterfaceTest`)
- Java compiler version (`FinallyTest`, `TryWithResourcesTest`)

Our current build setup ensures both are the same. To make the difference more explicit we could have two constants `JavaVersion.RUNTIME` and `JavaVersion.COMPILER`. Plus we would need a new `vendor` property. The tricky part here is that the compiler version cannot be fully derived as runtime (minor version required in `TryWithResourcesTest`). Maybe we can add this information at build time (`compilerversion.properties`).

Sorry for the noise, just thinking loud... ;)

Godin:
@namannigam Thanks for the comment. I know about `Runtime.Version` and associated JEPs [223](http://openjdk.java.net/jeps/223) and [322](http://openjdk.java.net/jeps/322): Terms `feature-release counter` and `update-release counter` for javadoc were actually taken from JEP 322. If we could use this API (Java >= 9), then to me seems that all conditions can be removed :wink:

@marchof

> The tricky part here is that the compiler version cannot be fully derived as runtime (minor version required in TryWithResourcesTest). Maybe we can add this information at build time (compilerversion.properties).

The really tricky part is to determine version of compiler when running tests from IntelliJ IDEA and not from command-line :wink: In https://github.com/jacoco/jacoco/pull/500#discussion_r109605564 we agreed that for now will be simpler if `javac` version matches runtime version. At the moment I would prefer not to investigate this and keep it as it is. Let's maybe just add a comment about this as javadoc of `ValidationTestBase.JAVA_VERSION`?

> Adding comparison Methods to `JavaVersion`

Good that for now all ranges can be declared without starting point (e.g. `TryWithResourcesTest` not executed on Java < 7) and they do not have holes, so indeed `isBefore` brings even more simplicity :thumbsup:

> Make `JavaVersion` `Comparable`

Referring to YAGNI & KISS, I think that `Comparable` is unnecessary.

> Moving `ValidationTestBase.JAVA_VERSION` to `JavaVersion.CURRENT`

I propose to keep it easily accessible to tests and close to `ValidationTestBase.isJDKCompiler`.

Have a look at addition of `isBefore` and let me know your new loud thoughts, or if I can merge this 😆

PR #645, opened 2018-02-02
Update version of commons-collections in jacoco-maven-plugin

### LINKED ISSUE #641: Update version of commons-collections dependency in jacoco-maven-plugin
[plugin.xml.txt](https://github.com/jacoco/jacoco/files/1616456/plugin.xml.txt)
The jacoco-maven-plugin uses apache commons collections 3.2, which has been determined to have a security vulnerability.  (Commons-collections-3.2.2 and higher are patched for the vulnerability.)  Our goal is to remove all vulnerable libraries from our local servers, build servers and artifactory, which means we will need to discontinue use of the jacoco maven plugin, unless it is patched.

**The dependency is injected via the \META-INF\maven\plugin.xml file.**

<dependency>
<groupId>commons-collections</groupId>
<artifactId>commons-collections</artifactId>
<type>jar</type>
<version>3.2</version>
</dependency>

JaCoCo version:  jacoco-maven-plugin-0.8.0 and below
Operating system:  Linux
Tool integration: Maven/Ant/API/Other

We need the dependency to point to version 3.2.2 or higher, otherwise, our applications that use the jacoco maven plugin will require the vulnerable library to be present in order to run builds.

===================================================

Here is information outlining the security vulnerability.

WebSphere vulnerability report:
https://www-01.ibm.com/support/docview.wss?uid=swg21970575

Determining if your application uses the Apache Commons InvokerTransformer and vulnerable to CVE-2015-7450
https://www.ibm.com/developerworks/community/blogs/aimsupport/entry/determining_if_your_application_uses_apache_commons_invokertransformer_and_vulnerable_to_cve20157450?lang=en

According to this document, anything prior to Apache Commons  v3.2.2 is vulnerable to the issue.
https://commons.apache.org/proper/commons-collections/javadocs/api-3.2.2/org/apache/commons/collections/functors/InvokerTransformer.html

This is the original publication of the vulnerability:
https://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

Godin:
Fixes #641

### DIFF
--- a/jacoco-maven-plugin/pom.xml
+++ b/jacoco-maven-plugin/pom.xml
@@ -30,6 +30,22 @@
     <maven>2.2.1</maven>
   </prerequisites>
 
+  <dependencyManagement>
+    <dependencies>
+      <!-- maven-reporting-impl and slight update of version of its transitive dependency on commons-collections -->
+      <dependency>
+        <groupId>org.apache.maven.reporting</groupId>
+        <artifactId>maven-reporting-impl</artifactId>
+        <version>2.1</version>
+      </dependency>
+      <dependency>
+        <groupId>commons-collections</groupId>
+        <artifactId>commons-collections</artifactId>
+        <version>3.2.2</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
   <dependencies>
     <dependency>
       <groupId>org.apache.maven</groupId>
@@ -60,7 +76,6 @@
     <dependency>
       <groupId>org.apache.maven.reporting</groupId>
       <artifactId>maven-reporting-impl</artifactId>
-      <version>2.1</version>
     </dependency>
 
     <dependency>


jacoco-maven-plugin/pom.xml
```diff
@@ -30,6 +30,16 @@
     <maven>2.2.1</maven>
   </prerequisites>
 
+  <dependencyManagement>
+    <dependencies>
```
marchof: @Godin Should we leave a comment here, why this redundant/explicit dependency has been added?

Godin: @marchof `dependencyManagement` is to explicitly declare version of transitive dependency, so strictly speaking can't say that it is redundant, and some even advice to fully control transivite dependencies and always explicitly declare their versions. And seems that `commons-collections` uses semantic versioning, so update from `3.2` to `3.2.2` (on its own) seems like a good idea to grab bug-fixes. Commit message and so `git blame` will anyway be linked with this PR and hence with initial ticket. But also can add link to a ticket as a comment, or you was thinking about some other comment?

marchof: @Godin Someone looking at it might ask: why do we care about this version? I was thinking about a in-line comment like:

> Avoid transitive dependency on version with security issue CVE-2015-7450

Godin: @marchof I still don't believe that serialization can be involved during execution of plugin, and so don't believe that security issue can be exploited here, also because

```diff
--- jacoco-maven-plugin.test/pom.xml
+++ jacoco-maven-plugin.test/pom.xml
@@ -31 +31 @@
-    <jacoco.includes>org.jacoco.maven.*</jacoco.includes>
+    <jacoco.includes>*</jacoco.includes>
```

```
$ mvn clean install
$ java -jar /tmp/jacoco-0.8.0/lib/jacococli.jar execinfo jacoco-maven-plugin.test/target/jacoco.exec | grep InvokerTransformer | wc -l
0
```

insecure `org/apache/commons/collections/functors/InvokerTransformer` is not touched during execution of our tests that IMO quite exhaustive. And so didn't wanted to scare by comment users that browse sources or follow commits (see for example https://github.com/jacoco/jacoco/commit/6a0c097ff42ff114b1efb0a7d264aaf1b290345e#commitcomment-26640407) 😉  But ok - can add such message and let you handle user questions about where is immediate release to fix security vulnerability 😆 or maybe add such comment just before release to scare users of old versions 😈

Godin: BTW latest version of `org.apache.maven.reporting:maven-reporting-impl:3.0.0` brings `3.2.1` as transitive dependency, so I'm pretty sure that many Maven plugins do not have `3.2.2`.

marchof: @Godin After this discussion I feel like we should close #641 as ´Other Project´.

* The only sustainable fix is bumping the dependency in the `maven-reporting-impl` project.
* There are [many other plugins](https://mvnrepository.com/artifact/org.apache.maven.reporting/maven-reporting-impl/usages) depending on it, so why adding a hack in JaCoCo?
* As you wrote the vulnerable part is not used.
* Builds need excessive local privileges like access to the file system, network etc anyways. So either you trust your build or you run it in a sandboxed environment.

marchof:
@Godin What is your preference about this issue?

1) Merge as is or
2) do nothing about it in JaCoCo?

After the discussion my personal preference would be 2.

Godin:
@marchof well, looks like we will need third opinion 😆 
While I agree that appearance of transitive dependency of `maven-reporting-impl` in `dependencyManagement` looks hacky and potentially might be forgotten in future with bad consequences, and that the best place to spread fix is in `maven-reporting-impl`, and that builds should be done in a sandboxed environment, and truly believe that there is no vulnerability and so even hard to classify such change, I also can understand that premise of #641 is a wish to simply globally ban dependency to get sandbox, and who knows when `maven-reporting-impl` will be updated and why not use mechanism to manage dependency. Solution described in #641 about update of dependencies of plugins without their releases can also be considered as error-prone "hacky" dependency management, but not on our side 😉

Maybe as a trade-off can propose to update this PR with comment next to dependency on `maven-reporting-impl` to not forget its connection with `dependencyManagement`, but why someone will not check all tree of dependencies during updates without this comment?

Or we can flip a coin to decide - [right now Google says that upper side is "heads"](https://www.google.com/search?q=flip+a+coin) 😆 In any case I will love reduction of open tickets and PRs 😉

marchof:
@Godin My personal preference is to do nothing here. But I can life with a documented declaration.

👍  for reduction of open tickets

Godin:
@marchof please check - IMO looks better now

marchof:
@Godin Looks better now, except that there is no mention _why_ the "slight update" is done. But good to merge for me -- at least I will probably remember why.

PR #649, opened 2018-02-10
Add filter for empty constructor without parameters in enum

### LINKED ISSUE #640: Unnecessary empty Enum Contructor shows up on report
JaCoCo reports missing coverage to enum constructor on empty enum.

### Steps to reproduce
Create a enum like this:
```java
enum EnumExample {
  ;
  public static final String CONST = "This is an empty enum";
}
```
The report shows a Element "EnumExample(String, int)" without any coverage.

JaCoCo version: 0.8.0
Operating system: Linux (Ubuntu 16.04)
Tool integration: Maven
Compiler: javac from OpenJDK Runtime Environment (build 1.8.0_151-8u151-b12-0ubuntu0.16.04.2-b12) OpenJDK 64-Bit Server VM (build 25.151-b12, mixed mode)

### Expected behaviour
In this case the constructor should be ignored and no Element should shown on the report.

### Actual behaviour
The report shows a Element for the enum constructor.

Godin:
Fixes #640

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.tree.AbstractInsnNode;
 import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
 import org.objectweb.asm.tree.VarInsnNode;
 
 abstract class AbstractMatcher {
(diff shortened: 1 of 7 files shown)


Godin:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/EnumEmptyConstructorFilter.java
```diff
@@ -0,0 +1,59 @@
...
+				&& new Matcher().match(methodNode, superClassName)) {
+			output.ignore(methodNode.instructions.getFirst(),
+					methodNode.instructions.getLast());
+		}
+	}
+
+	private static class Matcher extends AbstractMatcher {
+		private boolean match(final MethodNode methodNode,
+				final String superClassName) {
+			firstIsLoadThis(methodNode);
```
marchof: +1 for this very readable reader declaration ;)

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
```diff
@@ -25,6 +26,35 @@
 
 	AbstractInsnNode cursor;
 
+	/**
+	 * Sets {@link #cursor} to first instruction of method if it is
+	 * <code>ALOAD 0</code>, otherwise sets it to <code>null</code>.
+	 */
+	void firstIsLoadThis(final MethodNode methodNode) {
```
marchof: Strictly speaking the method name only applies to  instance methods. For static methods it would be the first argument.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
```diff
@@ -17,6 +17,7 @@
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.tree.AbstractInsnNode;
 import org.objectweb.asm.tree.MethodInsnNode;
+import org.objectweb.asm.tree.MethodNode;
 import org.objectweb.asm.tree.VarInsnNode;
 
 abstract class AbstractMatcher {
```
marchof: I wonder whether it would be worth to add unit tests for AbstractMatcher. Especially as we miss several branches with the rest of our test suite.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
Godin: @marchof renamed into `firstIsALoad0`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
Godin: @marchof Let's do this separately.

marchof: 👍

PR #652, opened 2018-02-21
Cleanup internal XML generation framework

marchof:
I started a SVG badge prototype these days based on our existing XML generator.  Hereby I identified  a major design flaw with our XML generator (root element and children have different class hierarchy) and several minor issues.

* Combine XMLElement and XMLDocument to allow single schema specific
subclass
* Add subclass for JaCoCo report XML schema
* Remove mostly unused return values
* Remove unused Writer output
* Cleanup unit tests
* Use new test naming conventions


marchof:
@Godin PR is now rebased

I agree that badges are primary for marketing purposes (people love it). Not saying that it will be a percentage badge. Could be a "365 Tests Missing" badge ;)

Godin:
LGTM 👍 Thanks!

PR #658, opened 2018-03-15
Maven goal `report-aggregate` should consider dependencies specified using version range

### LINKED ISSUE #657: report-aggregate goal doesn't work with deps defined using version ranges
### Rationale

I have a project where in a single reactor I have multiple independently versioned modules. I'm trying to create an aggregate coverage report over all of them. This is awkward because in the aggregate pom for the coverage I need to manually specify the concrete snapshot versions of all projects in the reactor because there is no universal `${project.version}` I could use.

I tried to use a version range `[0,)`  for the dependencies (i.e. just anything that is in the reactor) but that doesn't work and the aggregate report is empty.

### Steps to reproduce

JaCoCo version: 0.8.0
Operating system: any
Tool integration: Maven

aggregate/pom.xml:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <groupId>test</groupId>
    <artifactId>coverage</artifactId>
    <version>0</version>
    <packaging>pom</packaging>

    <dependencies>
        <dependency>
            <groupId>test</groupId>
            <artifactId>artifact</artifactId>
            <version>[0,)</version>
        </dependency>
    </dependencies>

    <build>
        <plugins>
            <plugin>
                <groupId>org.jacoco</groupId>
                <version>0.8.0</version>
                <executions>
                    <execution>
                        <phase>verify</phase>
                        <goals>
                            <goal>report-aggregate</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

### Expected behaviour
If the dependency in the reactor matches the version range specified in the aggregate report module, the dependency should be considered for the aggregate report.

### Actual behaviour
All dependencies specified using a version range are not included in the aggregate report.

metlos:
Fixes #657.


marchof:
@Godin What do you think about this PR? Beside adding a test this looks good to me.

metlos:
I've added an integration test heavily based on the `it-report-aggregate`. It should check that the jacoco files are loaded from locations of the projects in reactor that conform to the version range.

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -178,10 +181,19 @@ public String getName(final Locale locale) {
 	}
 
 	private MavenProject findProjectFromReactor(final Dependency d) {
+		VersionRange depVersionAsRange;
+		try {
+			depVersionAsRange = VersionRange.createFromVersionSpec(d.getVersion());
+		} catch (InvalidVersionSpecificationException e) {
+			throw new IllegalStateException("Could not parse the version of the dependency even though Maven" +
+					" could: " + d);
```
Godin: Can this ever happen?

metlos: Should not, really, but `InvalidVersionSpecificationException` is a checked exception, so there has to be something done about it...

metlos:
> I'm wondering why instead not simply update/extend it-report-aggregate?

Well, I've seen another version of the report aggreate it-test - it-report-aggregate-customization and so I thought this was the approach you take. I can merge it into the main it-test if you wish..

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
Godin: If I'm not mistaken, then we use `throw new AssetionError(e)` for such things - e.g. https://github.com/jacoco/jacoco/blob/ce7de98c76b8bd107c05a061193de76d21c4c205/org.jacoco.core/src/org/jacoco/core/data/ExecutionDataWriter.java#L123

Godin:
For the record: during testing I stumbled upon a fact that when reactor contains two  modules with same `groupId:artifactId` but of different versions (exactly like integration test), then first dependency which satisfies range will be picked. I.e. in case of `<version>[0,2]</version>`
* if version `1` is before version `2` in reactor, then version `1` will be picked
* and vise versa - if version `2` is before version `1` in reactor, then version `2` will be picked

This looks acceptable to me - case seems rare, range can be narrowed/specified to select correct version from reactor.

Godin:
@metlos thank you for your contribution which is now recorded in changelog ❤️

PR #629, opened 2018-03-18
Support JDK 10

Godin:
Classes of JDK 10 EA b35 use new classfile version number ( https://bugs.openjdk.java.net/browse/JDK-8188870 ) and so JaCoCo Agent 0.7.9 as well as agent from current build of master branch fail to start:
```
$ java -version
java version "10-ea"
Java(TM) SE Runtime Environment (build 10-ea+35)
Java HotSpot(TM) 64-Bit Server VM (build 10-ea+35, mixed mode)

$ java -javaagent:jacoco-0.7.9/lib/jacocoagent.jar     
Exception in thread "main" java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:564)
        at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:510)
        at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:522)
Caused by: java.lang.RuntimeException: Class java/util/UUID could not be instrumented.
        at org.jacoco.agent.rt.internal_8ff85ea.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:140)
        at org.jacoco.agent.rt.internal_8ff85ea.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:101)
        at org.jacoco.agent.rt.internal_8ff85ea.PreMain.createRuntime(PreMain.java:55)
        at org.jacoco.agent.rt.internal_8ff85ea.PreMain.premain(PreMain.java:47)
        ... 6 more
Caused by: java.lang.NoSuchFieldException: $jacocoAccess
        at java.base/java.lang.Class.getField(Class.java:1958)
        at org.jacoco.agent.rt.internal_8ff85ea.core.runtime.ModifiedSystemClassRuntime.createFor(ModifiedSystemClassRuntime.java:138)
        ... 9 more
FATAL ERROR in native method: processing of -javaagent failed
```

So to support JDK 10 we'll need to do similar workarounds that were done for JDK 9 in #406 and removed in #600 or wait for ASM release containing https://gitlab.ow2.org/asm/asm/merge_requests/74

### DIFF
--- a/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AnnotationOnLocalVariableTest.java
+++ b/org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AnnotationOnLocalVariableTest.java
@@ -1,0 +1,35 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation;
+
+import org.jacoco.core.analysis.ICounter;
+import org.jacoco.core.test.validation.targets.AnnotationOnLocalVariable;
+import org.junit.Test;
+
+/**
+ * Test of ASM bug
+ * <a href="https://gitlab.ow2.org/asm/asm/issues/317815">#317815</a>
+ */
+public class AnnotationOnLocalVariableTest extends ValidationTestBase {
+
+	public AnnotationOnLocalVariableTest() {
+		super("src-java8", AnnotationOnLocalVariable.class);
+	}
+
+	@Test
+	public void testCoverageResult() {
+
+		assertLine("var", ICounter.FULLY_COVERED);
+
+	}
+
+}
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/BytecodeVersionTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/BytecodeVersionTest.java
@@ -1,0 +1,53 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal;
+
+import org.junit.Test;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+
+public class BytecodeVersionTest {
+
+	@Test
+	public void should_return_original_when_not_java10() {
+		final byte[] originalBytes = createClass(Opcodes.V9);
+
+		final byte[] bytes = BytecodeVersion.downgradeIfNeeded(Opcodes.V9,
+				originalBytes);
+
+		assertSame(originalBytes, bytes);
+	}
+
+	@Test
+	public void should_return_copy_when_java10() {
+		final byte[] originalBytes = createClass(BytecodeVersion.V10);
+
+		final byte[] bytes = BytecodeVersion
+				.downgradeIfNeeded(BytecodeVersion.V10, originalBytes);
+
+		assertNotSame(originalBytes, bytes);
+		assertEquals(Opcodes.V9, BytecodeVersion.read(bytes));
+		assertEquals(BytecodeVersion.V10, BytecodeVersion.read(originalBytes));
+	}
+
+	private static byte[] createClass(final int version) {
+		final ClassWriter cw = new ClassWriter(0);
+		cw.visit(version, 0, "Foo", null, "java/lang/Object", null);
+		cw.visitEnd();
+		return cw.toByteArray();
+	}
+
+}
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertEquals;
 import java.io.UnsupportedEncodingException;
 
 import org.jacoco.core.data.ExecutionDataWriter;
+import org.jacoco.core.internal.BytecodeVersion;
 import org.junit.Test;
 import org.objectweb.asm.ClassWriter;
 import org.objectweb.asm.Opcodes;
@@ -27,6 +28,9 @@ public class CRC64Test {
 
 	@Test
 	public void testJava9() {
+		assertEquals(0x589E9080A572741EL,
+				CRC64.classId(createClass(BytecodeVersion.V10)));
+
 		// should remove workaround for Java 9
 		// during change of exec file version
 		assertEquals(0x1007, ExecutionDataWriter.FORMAT_VERSION);
--- a/org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
+++ b/org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
@@ -22,6 +22,7 @@ import java.util.zip.ZipInputStream;
 
 import org.jacoco.core.data.ExecutionData;
 import org.jacoco.core.data.ExecutionDataStore;
+import org.jacoco.core.internal.BytecodeVersion;
 import org.jacoco.core.internal.ContentTypeDetector;
 import org.jacoco.core.internal.InputStreams;
 import org.jacoco.core.internal.Pack200Streams;
@@ -106,8 +107,16 @@ public class Analyzer {
 	 *            reader with class definitions
 	 */
 	public void analyzeClass(final ClassReader reader) {
-		final ClassVisitor visitor = createAnalyzingVisitor(
-				CRC64.classId(reader.b), reader.getClassName());
+		analyzeClass(reader.b);
+	}
+
+	private void analyzeClass(final byte[] source) {
+		final long classId = CRC64.classId(source);
+		final int version = BytecodeVersion.read(source);
+		final byte[] b = BytecodeVersion.downgradeIfNeeded(version, source);
+		final ClassReader reader = new ClassReader(b);
+		final ClassVisitor visitor = createAnalyzingVisitor(classId,
+				reader.getClassName());
 		reader.accept(visitor, 0);
 	}
 
@@ -124,7 +133,7 @@ public class Analyzer {
 	public void analyzeClass(final byte[] buffer, final String location)
 			throws IOException {
 		try {
-			analyzeClass(new ClassReader(buffer));
+			analyzeClass(buffer);
 		} catch (final RuntimeException cause) {
 			throw analyzerError(location, cause);
 		}
--- a/org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
+++ b/org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
@@ -21,9 +21,11 @@ import java.util.zip.ZipEntry;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
 
+import org.jacoco.core.internal.BytecodeVersion;
 import org.jacoco.core.internal.ContentTypeDetector;
 import org.jacoco.core.internal.InputStreams;
 import org.jacoco.core.internal.Pack200Streams;
+import org.jacoco.core.internal.data.CRC64;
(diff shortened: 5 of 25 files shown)


Godin:

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AnnotationOnLocalVariableTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.jacoco.core.analysis.ICounter;
+import org.jacoco.core.test.validation.targets.AnnotationOnLocalVariable;
+import org.junit.Test;
+
+/**
+ * Test of ASM bug
+ * <a href="https://gitlab.ow2.org/asm/asm/issues/317815">#317815</a>
+ */
+public class AnnotationOnLocalVariableTest extends ValidationTestBase {
```
marchof: I wonder whether this test should go into a separate commit as it is not strictly related with the rest of the changes (ok, we do downgrade/upgrade because new ASM release is broken...).

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal;
+
+import org.objectweb.asm.Opcodes;
+
+/**
+ * Utilities to read and modify bytecode version in bytes of class.
+ */
+public final class BytecodeVersion {
```
marchof: ASM has an API to read/write bytecode versions, therefore we should give a reason here why we have this class. Maybe add some Javadoc: "The purpose of this class is to deal with new Java versions which are not yet supported by ASM".

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
```diff
@@ -0,0 +1,78 @@
...
+
+	/**
+	 * Sets bytecode version in given bytes of class.
+	 *
+	 * @param b
+	 *            bytes of class
+	 * @param version
+	 *            version of bytecode to set
+	 */
+	public static void set(final byte[] b, final int version) {
```
marchof: For symmetry reasons I would prefer read/write (or get/set) for the method names.

org.jacoco.core.test/src/org/jacoco/core/internal/BytecodeVersionTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.junit.Test;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+
+public class BytecodeVersionTest {
```
marchof: Add explicit tests for read/write methods?

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
```diff
@@ -106,8 +107,16 @@ public void visitEnd() {
 	 *            reader with class definitions
 	 */
 	public void analyzeClass(final ClassReader reader) {
-		final ClassVisitor visitor = createAnalyzingVisitor(
-				CRC64.classId(reader.b), reader.getClassName());
+		analyzeClass(reader.b);
+	}
+
+	private void analyzeClass(final byte[] source) {
```
marchof: Move this private method below the public analyze* methods?

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
```diff
@@ -0,0 +1,78 @@
...
+	 * otherwise copy where version set to {@link Opcodes#V9}.
+	 * 
+	 * @param version
+	 *            version of bytecode
+	 * @param source
+	 *            bytes of class
+	 * @return given bytes of class if version is less than {@link #V10},
+	 *         otherwise copy where version set to {@link Opcodes#V9}
+	 */
+	public static byte[] downgradeIfNeeded(final int version,
```
marchof: From usage pattern it appears that a overloaded version `downgradeIfNeeded(byte[] source)` would be nice. So most call sited don't have to read version themself.

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
```diff
@@ -76,6 +78,15 @@ public void setRemoveSignatures(final boolean flag) {
 	 * 
 	 */
 	public byte[] instrument(final ClassReader reader) {
+		return instrument(reader.b);
+	}
+
+	private byte[] instrument(final byte[] source) {
```
marchof: Move the private method below the public `instrument()` methods?

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactory.java
```diff
@@ -30,19 +29,21 @@ private ProbeArrayStrategyFactory() {
...
 	 *            accessor to the coverage runtime
 	 * @return strategy instance
 	 */
-	public static IProbeArrayStrategy createFor(final ClassReader reader,
+	public static IProbeArrayStrategy createFor(final long classId,
+			final ClassReader reader,
 			final IExecutionDataAccessorGenerator accessorGenerator) {
 
 		final String className = reader.getClassName();
 		final int version = getVersion(reader);
```
marchof: Use new `BytecodeVersion` API instead.

org.jacoco.core.test/src-java8/org/jacoco/core/test/validation/AnnotationOnLocalVariableTest.java
Godin: @marchof extracted - https://github.com/jacoco/jacoco/pull/662

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
Godin: @marchof renamed `read` into `get`

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
Godin: @marchof To me seems logical that this method is at the beginning, being the main part of process of instrumentation. Also didn't know that we have convention that `public`s should be before `private`s, this is already not respected - private method `instrumentError` is before public method `instrumentAll`. And most importantly - movement will affect diff in a way that will be harder to understand what was changed. If strictly needed (?) - will prefer to do this reordering afterwards separately.

org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java
Godin: @marchof same as https://github.com/jacoco/jacoco/pull/629#discussion_r175766878

org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java
marchof: @Godin The private method is now called from the public method above and from another public `instrument()` method below. Without the history the placement in the middle looks a bit random to me. But not strictly needed as all.

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
Godin: @marchof This method must be used with caution - original version must be restored in main code in 2 out of 3 call sites (needed in `ModifiedSystemClassRuntime` and `Instrumenter`, not needed only in `Analyzer`). So by not providing overloaded version and by forcing read of version, this in my opinion forces to think twice about proper usage. Other calls are in tests and mainly only for some of them overloaded version will be useful. Thus not sure that addition of overloaded version is a good idea. WDYT?

org.jacoco.core.test/src/org/jacoco/core/internal/BytecodeVersionTest.java
Godin: @marchof while adding test, I realized that this is only about major segment of version, i.e.

```
BytecodeVersion.get(createClass(Opcodes.V1_1)) != Opcodes.V1_1
```

this is consistent with your request https://github.com/jacoco/jacoco/pull/629#discussion_r175668010 to use `BytecodeVersion.get` in `ProbeArrayStrategyFactory` because reading of both major and minor will break comparisons in it - counterintuitively `Opcodes.V1_1` is greater than `Opcodes.V9` and `BytecodeVersion.V10`.

I think that for the time being will be enough just to document this and reflect in test. WDYT?

marchof: @Godin Just add to JavaDoc that get/set is for major version only. The day when OpenJDK releases weekly with minor versions we think about it again ;)

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
marchof: @Godin I was also looking for better ways to make down-/upgrade fool proof. One idea was to create a stateful sinstance of the utility:

	FutureBytecodeSupport future = new FutureBytecodeSupport();
	byte[] instrumented = future.upgrade(instrument(future.downgrade(original)));

But still you could forget to apply `upgrade()`. With Java 8 we could wrap an `Function`. For now I would simply add Javadoc to our utility how to use it properly.

org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
```diff
@@ -27,6 +28,9 @@
 
 	@Test
 	public void testJava9() {
+		assertEquals(0x589E9080A572741EL,
```
marchof: Separate test per Java version?

org.jacoco.core.test/src/org/jacoco/core/internal/BytecodeVersionTest.java
Godin: @marchof test added, JavaDoc updated

org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeArrayStrategyFactory.java
Godin: @marchof done

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
Godin: @marchof so let's keep it as is

org.jacoco.core.test/src/org/jacoco/core/internal/data/CRC64Test.java
Godin: @marchof I would prefer to keep single test, so that values are written next to each other to make clear that they are all different except for Java 9 that is equal to Java 8. So just renamed test.

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
Godin: @marchof JavaDoc updated

Godin:
@marchof I think all comments were addressed, so could you please re-review?

marchof:
@Godin Done, thanks!

olamy:
thanks guys for your hard work I just tested a build and everything works fine!!
Thanks again!!

Godin:
Now that JaCoCo 0.8.1 with Java 10 support has been released, can think of Java 11. And here is question to you guys @don-vip @ijuma @sormuras : do you compile into bytecode version 55 (Java 11) or just run under JDK 11 EA ? The reason for this question is that the second case is easier and maybe could be possible to implement before ASM release.

Godin:
@sormuras just to be sure that there is no misunderstanding: JaCoCo 0.8.1 supports bytecode versions from 45 (Java 1.1) to 54 (Java 10) and runs under JDKs from 5 to 10, and doesn't run under JDK 11 EA. My question about 11 EA is for our future work.

ijuma:
@Godin We use `--release 7` (soon planning to switch to `--release 8`) in Kafka. We just want to be able to compile and run the tests with Java 9, 10 and (soon) 11 to ensure that we catch any problems early.

sormuras:
> @sormuras just to be sure that there is no misunderstanding: JaCoCo 0.8.1 supports bytecode versions from 45 (Java 1.1) to 54 (Java 10) and runs under JDKs from 5 to 10, and doesn't run under JDK 11 EA. My question about 11 EA is for our future work.

Understood.

We're using OracleJDK 9 and OpenJDK 10 to build and test JUnit 5 at the moment: https://travis-ci.org/junit-team/junit5/builds/355963543 -- JaCoCo is only activated when running with 9. This might change tonight, after upgrading to JaCoCo 0.8.1

11-ea is blocked by https://github.com/gradle/gradle/issues/4515 ... was blocked! Waiting for Gradle 4.7-RC1 now, and JaCoCo 0.8.2? 😉

sormuras:
fyi, JaCoCo 0.8.1 works like a charm on OpenJDK 10: https://codecov.io/gh/junit-team/junit5/list/0bb65c5b945e3cfbb7c99ded92b80b96642e96e6/

PR #668, opened 2018-04-03
Filter out synthetic classes during generation of report

Godin:


marchof:
@Godin For now I think this can be hardcoded in `Analyzer`. In case we need to extend the filtering API to classes for other use cases we can still move it to the new API.

Godin:
@marchof in this case could you please approve and I will do squash and merge 😉

org.jacoco.core.test/src/org/jacoco/core/test/filter/EnumSwitchTest.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.filter;
+
+import org.jacoco.core.analysis.ICounter;
+import org.jacoco.core.test.filter.targets.EnumSwitch;
+import org.jacoco.core.test.validation.JavaVersion;
```
marchof: Unused import gives warning.

org.jacoco.core.test/src/org/jacoco/core/test/filter/EnumSwitchTest.java
```diff
@@ -0,0 +1,39 @@
...
+public class EnumSwitchTest extends ValidationTestBase {
+
+	public EnumSwitchTest() {
+		super(EnumSwitch.class);
+	}
+
+	@Test
+	public void testCoverageResult() {
+		if (isJDKCompiler && JAVA_VERSION.isBefore("1.6")) {
+			assertLine("switch", ICounter.PARTLY_COVERED, 0, 2);
```
marchof: Should we leave a comment why this is the case?

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -20,6 +20,12 @@ <h1>Change History</h1>
 
 <h2>Snapshot Build @qualified.bundle.version@ (@build.date@)</h2>
 
+<h3>New Features</h3>
+<ul>
+  <li>Synthetic classes are filtered out during generation of report
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/668">#668</a>).</li>
```
marchof: What is our rule here: Do we refer to the PR or to the issue?

Godin: @marchof I prefer to refer to actual changes, so to PR,  and was following this principle.

org.jacoco.core.test/src/org/jacoco/core/test/filter/EnumSwitchTest.java
Godin: @marchof added comment

org.jacoco.core.test/src/org/jacoco/core/test/filter/EnumSwitchTest.java
Godin: @marchof unused import removed

PR #669, opened 2018-04-03
Update TryWithResourcesJavacFilter for javac 11

Godin:
`javac 11 EA` starting from `b7` and up to latest (as of today) `b24` contains http://hg.openjdk.java.net/jdk/jdk/rev/c2a3a2aa2475 ( [JDK-8194978](https://bugs.openjdk.java.net/browse/JDK-8194978) ), which changes bytecode generation for `try-with-resource`. Our `TryWithResourcesJavacFilter` does not detect this new bytecode.

Based on analysis of change, description of desugaring in [javadoc of method `makeTwrTy` in class `com/sun/tools/javac/comp/Lower` from `b24`](http://hg.openjdk.java.net/jdk/jdk/file/jdk-11+24/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/Lower.java#l1558), and practical observations:
* there is no `null` check of `primaryExc` anymore
* as with `javac 9` when resource is initialized with `new`, then `null` check of resource is omitted
* synthetic method `$closeResource` (that was introduced in `javac 9`) not generated anymore

To execute our validation tests with `javac 11 EA` in absence of full support of JDK 11:
```
$ javac -version
javac 11-ea

$ mvn clean verify \
    -pl org.jacoco.core,org.jacoco.core.test.validation.java7 \
    -Djacoco.skip -Dbytecode.version=10
```


Godin:
@marchof We need to continue work towards JDK 11 support - [AFAIK initial release candidate is planned on 2018/08/16](http://openjdk.java.net/projects/jdk/11/). So could you please have a look on this new filter? While it is similar to existing, I think that better to have them separate - existing one is already quite complex.

marchof:
@Godin Would it be already possible to add an JDK 11 build to Travis?

Godin:
@marchof 

> change log should be updated

Done, please check.

> I would love to have a more fluent matcher API. 

Same from my side, but as usual - baby steps, baby steps :wink:

> Would it be already possible to add an JDK 11 build to Travis?

Definitely! Was planning and will do so right after merge of this PR, so that javac 11 passes our tests 😆  and doesn't lead to build failure.

PR #672, opened 2018-04-13
Question mark in filter expressions should match exactly one character

### LINKED ISSUE #671: Incorrect filter convertor in WildcardMatcher
There is a bug in https://github.com/jacoco/jacoco/blob/6d4d8530f972162cc3a58e2161cf6e01cc13123e/org.jacoco.core/src/org/jacoco/core/runtime/WildcardMatcher.java#L49-L50

It incorrectly converts wildcard `?` into `.?` but it should be just `.`

marchof:
Fixes #671


Godin:
LGTM :thumbsup: for readability only changed sign "?" on words "question mark" in changelog, hope you don't mind.

marchof:
@Godin Thanks!

PR #675, opened 2018-04-18
Add method filter ExcludeFromCodeCoverage

philieu:
This PR is to allow filter methods that are annotated with a class named ExcludeFromCodeCoverage. The Java package of the class is not important. 
It will allow the end users the flexibility to create this annotation class themselves and use it to filter out methods from code coverage. 

Ref. #14

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/ExcludeFromCodeCoverageFilterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/ExcludeFromCodeCoverageFilterTest.java
@@ -1,0 +1,85 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Phi Lieu - add ExcludeFromCodeCoverage annotation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.jacoco.core.internal.instr.InstrSupport;
+import org.junit.Test;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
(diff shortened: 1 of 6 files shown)


jhulfikarali:
Is this PR planned for next Jacoco release? This will allow us to focus more on Service testing.

philieu:
Hi @Godin,

Is there anything I can do to move this PR forward?

As mentioned in another thread, it was created with the following scenario in mind:

* A build would fail if the code coverage is not 100%
* There is an ability to exclude certain parts of the code via annotations or comments
* What is being excluded is clear and right there in pull requests
* The team can then discuss the merit and reason why certain code is being excluded

In our case, the issue with percentage-based code coverage quality gate is that it is often not clear where the non-covered code is (unless one go through the report in details) and people can often trick coverage (e.g. write tests for the trivial parts but ignore the important bit).
Also, we can already "cheat" by using the lombok.Generated or other "well-known" annotations. It's just not nice having to use that for the above purpose.

Cheers.

anthonypt87:
Upvoting this. Getting this out would be great. People are misusing lombok.generated to get this functionality (https://stackoverflow.com/questions/47824761/how-would-i-add-an-annotation-to-exclude-a-method-from-a-jacoco-code-coverage-re)

arouel:
@philieu can you update the PR and resolve the conflicts?
@Godin what is blocking you to bring that into a release soon?

philieu:
@arouel I'm happy to spend time to update the PR and resolve the conflicts. However, given that this PR is from April 2018, I don't think the author is interested in this feature or intent to merge this PR.

arouel:
@philieu thank you for the update. I'm really curious what @Godin thinks about it?

ocind:
@Godin @marchof I was hoping to have this feature for Jacoco! Do we have some design concerns regarding this feature? Or you are currently simply too busy to look into this? Do let us know how we can help!

marchof:
@philieu Thanks for this pull request! After internal discussions we decided that we will not support manual exclusion. We assume that this has a long tail of more feature requests and bugs:

* Users will expect to have tags for more places: Classes, parts of methods (e.g. catch blocks)
* Users will ask for show excluded code separately in reports
* The implementation proposed by you does not work for lambda bodies in the annotated methods.

philieu:
@marchof Thanks for the message. I can understand where you are coming from.
I can see you already had / will have a long tail of user requests regarding Lombok, Groovy or whatever code generation tools... for now and in the future. 
My point is that instead of that, why don't we have the flexibility to specify the annotation exclusion at project config and support anything in the code that can be annotated. Something like this: https://github.com/tonerdo/coverlet/blob/master/Documentation/MSBuildIntegration.md#attributes

Anyway, thanks a lot for you and everyone's effort in building the tool. Much appreciated.

PR #680, opened 2018-04-28
Add current module to aggregated report

gslowikowski:
Previously only coverage data from dependent modules participated in aggregated report. After this change current module data is added as well.

In some cases this will simplify multi-module project configuration. Dedicated module for aggregated report will not be required. There will be possibility to generate aggregated report for the module depending on all other (if there is such).


marchof:
@gslowikowski This is included unconditionally, right? If the aggregate module contains code this will become part of the report. I assume we make other users unhappy with this change. So probably we need at least a way to make this behavior configurable.

PR #682, opened 2018-05-07
CLI tests should not depend on current locale

marchof:
As mentioned in out mailing list our CLI tests fail for locales other than english:

https://groups.google.com/d/msg/jacoco/zOw0llfraAM/wMiYu5CUEAAJ

The build should therefore configure english as the default locale.

### DIFF
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
@@ -24,7 +24,7 @@ public class MainTest extends CommandTestBase {
 
 		assertFailure();
 		assertNoOutput(out);
-		assertContains("Argument \"<command>\" is required", err);
+		assertContains("\"<command>\"", err);
 		assertContains("Usage: java -jar jacococli.jar --help | <command>",
 				err);
 		assertContains("Command line interface for JaCoCo.", err);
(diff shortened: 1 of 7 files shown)


marchof:
@Godin This is a bug in our version of args4j which has been fixed later:

https://github.com/kohsuke/args4j/commit/8bf2b9ec70be6016175bf76c4d423c428ade3378

Unfortunatelly we can't upgrade as our version is the last one with support for Java 5. What are the options:

1. Remove respective text assertions in `should_print_usage_when_no_arguments_given` tests
2. Use args4j message bundle to localize expected test
3. Replace args4j with a small custom framework specific to our needs

For the quick fix I would vote for 1). For the long run 3) is probably our only option if we want to continue support  for old Java versions

Godin:
@marchof :bulb:

4. remove `Messages_de_DE.properties` and `Messages_ru_RU.properties` during shading and depend on shaded artifact in tests - need to check if IDEs can understand such dependency, AFAIR in past there were problems with things like that, one of workarounds - repackage and publish jar with removed files using our groupId
5. invoke `org.jacoco.cli.internal.Main` in tests via ClassLoader with restrictions

marchof:
@Godin To keep things simple for now I just pushed version 1. I don't think there is a big value in verifying error messages in negative tests.

Godin:
> depend on shaded artifact in tests

* Seems that you already tried this in past and Eclipse wasn't happy - https://github.com/jacoco/jacoco/pull/525#discussion_r114322837
* IntelliJ behaves interestingly - it considers such dependency as dependency without classifier.
* Maven is fine with this as long as at least `mvn package` is used, but not just `mvn test` - I guess simply because shading is bound to `package` phase.

😞 

> To keep things simple for now
> I don't think there is a big value in verifying error messages in negative tests.

as long as we don't update `args4j` and this update doesn't bring us crazy change of messages :wink: I suppose both are quite unlikely in near future :laughing:

So let's use this option 👍 

And in any case I propose to fix

> awkward that some messages are localized, whereas others not

by

> remove `Messages_de_DE.properties` and `Messages_ru_RU.properties` during shading

Issue #654, opened 2018-02-25
Kotlin inline functions are not marked as covered


cplain:
A similar issue was raised here for Jetbrains coverage tools:
https://youtrack.jetbrains.com/issue/KT-12605
with the ultimate resolution being tracked here:
https://youtrack.jetbrains.com/issue/IDEA-185485

Although the former ticket mentions raising an issue on this project as far as I could see none has been

### Steps to reproduce
```
fun contrivedExample(condition: Boolean) {
    anInlineMethod(condition) { printLn("A thing happened") }
}

inline fun anInlineMethod(condition: Boolean, callback: () -> Unit) {
  if (condition) {
    callback.invoke()
  }
}
```

### Expected behaviour
If I write a test with contrivedExample(condition = true), anInlineMethod should be considered covered

### Actual behaviour
It isn't

JaCoCo version: 0.7.9
Operating system: macOS 10.13.2
Tool integration: Gradle

marchof:
Some remarks:

1. When the compiler assignes correct line numbers to the inlined code source highlighting should work. `javac` does so for lambda bodies
2. If the inlined method is still present in the compiled class file we need to get some hint that the method should be excluded (See #552).
3. If the method is inlined in multiple methods and not fully executed in one of the methods this will result in partial coverage.

yole:
Just to clarify: The main issue with line numbers and inline functions is that inline functions can be declared in different source files, so the .class file contains line numbers from different source files. This can't be expressed in a `LineNumberTable`, so Kotlin creates synthetic line numbers for lines that were inlined from other source files; those line numbers go beyond the line numbers in the original source file. The actual mapping between line numbers and original source files is stored in JSR-45 strata, which the compiler also generates.

JaCoCo needs to understand this; it's not something that can be changed on the Kotlin compiler side.

snowe2010:
Is this any closer to being resolved?

mformetal:
This is also an issue for me. Has there been any further steps to address this issue?

snowe2010:
... Why were our comments marked as spam? There have been no updates on this issue for a year, it's not unreasonable to ask for an update from the maintainers. I'm not saying you have to fix anything, but marking stuff as spam seems very uncalled for.

Godin:
[...]

PR #689, opened 2018-06-05
Add filter for methods that Kotlin compiler generates

goodwinnk:
Filtering for Kotlin generated methods based on line numbers absence among method instructions.


marchof:
@goodwinnk Are generated method already without line numbers? Or what would be the minimum required Kotlin compiler version?

goodwinnk:
Generated methods are without line numbers since Kotlin 1.0.0 till the latest one.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinNoSourceLinesFilter.java
```diff
@@ -0,0 +1,59 @@
...
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.LineNumberNode;
+import org.objectweb.asm.tree.MethodNode;
+
+import java.util.ListIterator;
+import java.util.Set;
+
+public class KotlinNoSourceLinesFilter implements IFilter {
```
marchof: JavaDoc missing.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilter.java
```diff
@@ -28,12 +30,17 @@
...
+	 *            annotations qualified name on class
+	 * @param sourceFileName
+	 *            file source name
 	 * @param methodNode
 	 *            method to inspect
 	 * @param output
 	 *            callback to report filtering results to
 	 */
-	void filter(String className, String superClassName, MethodNode methodNode,
+	void filter(String className, String superClassName,
```
marchof: Now we're definitely beyond the limit of extra parameters. Probably we will see even more in future. I would propose to introduce a new interface

    public interface org.jacoco.core.internal.analysis.filter.IFilterContext{
        String getClassName();
        String getSuperClassName();
        Set<String> getClassAnnotations();
        String getSourceFileName();
    }

For now this can be directly implemented by `ClassAnalyzer`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinNoSourceLinesFilter.java
Godin: In addition to a missing Javadoc - maybe also rename into `KotlinGeneratedFilter` by analogy with existing `GroovyGeneratedFilter` and `LombokGeneratedFilter` ?

marchof:
@Godin I tried to address your issues in first step. I will work on my findings tonight.

marchof:
@goodwinnk Silly question from my side: Have you done some integration testing with a Kotlin project? Are the results as expected?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilter.java
goodwinnk: I thought about introducing something like `IFilterContext` but decided that it's a bit more structural change than this PR needs.

goodwinnk:
@marchof, @Godin Thank you for the review. 
@marchof I didn't notice your changes and prepared some fixes, but they are probably the same (https://github.com/goodwinnk/jacoco/commits/kotlin-generated-filter-fixes).

I would still like to ask you fill `changes.html` in the way you prefer it.

> @goodwinnk Silly question from my side: Have you done some integration testing with a Kotlin project? Are the results as expected?

I'm using the sample project https://github.com/goodwinnk/kotlin-features-coverage and have just updated readme with the results from jacoco with `KotlinGeneratedFilter` enabled. 

All results are as expected. 
Inlines are still uncovered because of the https://github.com/jacoco/jacoco/issues/654. Function cannot be considered `generated` as it written in user code.
There're also problems with coroutines feature (there're uncovered branches in coroutines), but there would be an additional investigation, because coroutines are not finalized feature in Kotlin.

marchof:
@goodwinnk Thanks for the deedback! Sure, I will finalize this PR.

marchof:
@Godin I tried to address all issues . Can you please re-review?

hleb-albau:
Hi, thanks !
When new release is planned?

NickButcher1:
I have pointed my app at 0.8.2-SNAPSHOT.  The fix is working nicely, thanks.

hleb-albau:
@marchof I also moved to SNAPSHOT release, no issues, thanks.

PS: Sorry, I should read docs first.

krobitai:
So I just pulled down 0.8.2 but i'm still seeing the generated methods in my data classes (equals, to string, component etc...).  Is there a property I need to turn on to enable the filtering?

Godin:
@krobitai No there is no property. Just several messages above yours there is `example.zip` that contains complete working example.

Barryrowe:
Is there a configuration option that needs to be set to take advantage of this with 0.8.2? 
I have updated a project from 0.8.0 --> 0.8.2 but see no difference in my coverage and data classes still calculate coverage on generated toString()/hashCode() etc. functions.

Godin:
No - all filters so far implemented in JaCoCo are enabled unconditionally and do not require any configuration.

Also by carefully reading this thread - you can find above [example.zip](https://github.com/jacoco/jacoco/files/2078539/example.zip) which contains complete example for Maven and Gradle demonstrating that `toString` and `hashCode` are perfectly filtered out in data classes as shown on corresponding screenshots. This example was used as end-to-end test of unreleased version, however replacement of `0.8.2-SNAPSHOT` on `0.8.2` in `pom.xml` and `build.gradle` doesn't change result.

Please also pay attention on what is stated in each release announcement [starting from 0.8.0](https://groups.google.com/d/msg/jacoco/e2PVMeMvA8k/uHsJbi4iAQAJ) up to [0.8.2](https://groups.google.com/d/msg/jacoco/aMi1SkcfLZc/_UWgrC9xAQAJ) and in [changelog](https://www.jacoco.org/jacoco/trunk/doc/changes.html):

filtering happens at a time of generation of reports, so that

> Integrations developed as part of JaCoCo project by us (Ant Tasks, Maven Plugin and Command Line Interface) provide and new filters.

while

> Tools that directly read exec files (which is not a final report) and embed JaCoCo for generation of report will provide filtering functionality only after they updated to this version of JaCoCo.

If the above doesn't help, then please use [our mailing list](https://groups.google.com/forum/#!forum/jacoco) to provide **complete example** that demonstrates your problem.

PR #702, opened 2018-06-26
Add new attribute "sourcefilename" to "class" element in XML report

### LINKED ISSUE #701: Matching classes with source files from the XML report
Here is an example JaCoCo XML report:

        <report name="example">
            <package name="org/example">
                <class name="org/example/AnotherTopLevelClass">
                    ...
                </class>
                <class name="org/example/Person">
                    ...
                </class>
                <sourcefile name="Person.java">
                    ...
                </sourcefile>
            </package>
            ...
        </report>

The file `Person.java` holds two classes: `Person` and `AnotherTopLevelClass`.

In this case, there is no way to link the class coverage information (the `<class/>` elements) with the file coverage information (the `<sourcefile/>` elements) for cases in which multiple top level classes are implemented in a single Java file.

This problem does not occur if each `.java` file has only one top class, because Java requires the public class and the name to have the same name.

Here is a report format that would solve the issue:

        <report name="example">
            <package name="org/example">
                <sourcefile name="Person.java">
                    <class name="org/example/AnotherTopLevelClass">
                        ...
                    </class>
                    <class name="org/example/Person">
                        ...
                    </class>
                    ...
                </sourcefile>
            </package>
            ...
        </report>

How would one link the `<class/>` and `<sourcefile/>` instances from the XML reports?

### Version details

JaCoCo version: 0.8.0
Operating system: Linux
Tool integration: Gradle

marchof:
A new attribute should be added to allow mapping source files with secondary classes.

Fixes #701

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -45,6 +45,14 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/703">#703</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+  <li>The XML report now has an optional attribute <code>sourcefilename</code>
+      on the <code>class</code> element to allow unambiguously relate classes
+      to source files. The JaCoCo DTD version has been updated to 1.1. 
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/701">#701</a>).</li>
+</ul>
+
 <h2>Release 0.8.1 (2018/03/21)</h2>
 
 <h3>New Features</h3>
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -49,8 +49,8 @@
 <ul>
   <li>The XML report now has an optional attribute <code>sourcefilename</code>
       on the <code>class</code> element to allow unambiguously relate classes
-      to source files. The JaCoCo DTD version has been updated to 1.1. 
-      (GitHub <a href="https://github.com/jacoco/jacoco/issues/701">#701</a>).</li>
+      to source files. The JaCoCo DTD version has been updated to 1.1
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/702">#702</a>).</li>
 </ul>
 
 <h2>Release 0.8.1 (2018/03/21)</h2>
--- a/org.jacoco.report/src/org/jacoco/report/xml/XMLFormatter.java
+++ b/org.jacoco.report/src/org/jacoco/report/xml/XMLFormatter.java
@@ -32,7 +32,7 @@ import org.jacoco.report.internal.xml.XMLGroupVisitor;
  */
 public class XMLFormatter {
 
-	private static final String PUBID = "-//JACOCO//DTD Report 1.0//EN";
+	private static final String PUBID = "-//JACOCO//DTD Report 1.1//EN";
 
 	private static final String SYSTEM = "report.dtd";
 
--- a/org.jacoco.report/src/org/jacoco/report/xml/report.dtd
+++ b/org.jacoco.report/src/org/jacoco/report/xml/report.dtd
@@ -13,7 +13,7 @@
 <!-- This DTD describes the JaCoCo XML report format. It is identified by the
      following identifiers:
 
-        PUBID  = "-//JACOCO//DTD Report 1.0//EN"
+        PUBID  = "-//JACOCO//DTD Report 1.1//EN"
         SYSTEM = "report.dtd"
 -->
 
@@ -44,6 +44,8 @@
 <!ELEMENT class (method*, counter*)>
   <!-- fully qualified VM name -->
   <!ATTLIST class name CDATA #REQUIRED>
+  <!-- name of the corresponding source file -->
+  <!ATTLIST class sourcefilename CDATA #IMPLIED>
 
 <!-- representation of a method -->
 <!ELEMENT method (counter*)>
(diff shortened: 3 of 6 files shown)


org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -38,6 +38,14 @@ <h3>Fixed Bugs</h3>
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/672">#672</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+  <li>The XML report now has an optional attribute <code>sourcefilename</code>
+      on the <code>class</code> element to allow unambiguously relate classes
+      to source files. The JaCoCo DTD version has been updated to 1.1. 
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/701">#701</a>).</li>
```
Godin: As we discussed in https://github.com/jacoco/jacoco/pull/668#discussion_r178962005 - existing links point to PRs, not on original issue.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -38,6 +38,14 @@ <h3>Fixed Bugs</h3>
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/672">#672</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+  <li>The XML report now has an optional attribute <code>sourcefilename</code>
+      on the <code>class</code> element to allow unambiguously relate classes
+      to source files. The JaCoCo DTD version has been updated to 1.1.
```
Godin: Currently existing items consisting of multiple sentences use only one dot after link to GitHub in the last sentence.

org.jacoco.report/src/org/jacoco/report/xml/report.dtd
```diff
@@ -44,6 +44,8 @@
 <!ELEMENT class (method*, counter*)>
   <!-- fully qualified VM name -->
   <!ATTLIST class name CDATA #REQUIRED>
+  <!-- name of the corresponding source file -->
+  <!ATTLIST class sourcefilename CDATA #IMPLIED>
```
Godin: My eyes are bleeding a bit as a reaction on readability of this attribute name. But this is consistent with already existing other attributes that don't use CamelCase, what a pity that XML is case sensitive.

Wondering if better to keep it consistent with Java API or simplify into e.g. `source`, `sourcefile`? As a note - attribute of class denoting name of source file has name `SourceFile` in JVMS.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -38,6 +38,14 @@ <h3>Fixed Bugs</h3>
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/672">#672</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+  <li>The XML report now has an optional attribute <code>sourcefilename</code>
```
Godin: What do you think about addition of note that for better forward compatibility we recommend consumers of XML report to not fail (ignore / warn) on unknown attributes / nodes ?

org.jacoco.report/src/org/jacoco/report/xml/XMLFormatter.java
```diff
@@ -32,7 +32,7 @@
  */
 public class XMLFormatter {
 
-	private static final String PUBID = "-//JACOCO//DTD Report 1.0//EN";
+	private static final String PUBID = "-//JACOCO//DTD Report 1.1//EN";
```
Godin: FYI @marchof I triggered works on changing SonarQube integration from reading of binary to XML - https://jira.sonarsource.com/browse/MMF-1362 And for the time being seems that no changes will be required on JaCoCo side. Was thinking that maybe with increasing adoption of XML report, we'll need to introduce something like checksum of source file so that consumer can check it, however as of now I'm not aware of problems with mismatches in case of imports from other coverage tools.

marchof: Agreed. Can we handle this in a separate PR once someone comes up with a real use case?

org.jacoco.doc/docroot/doc/changes.html
marchof: Not sure whether our change log is the right place to advocate robust interface design line *tolerant reader*.

org.jacoco.report/src/org/jacoco/report/xml/report.dtd
marchof: I had the same thoughts but then decided for a consistent naming:

1) In our API the attribute is also called sourceFileName (ok in camel case)
2) Also it is conistent with the XML schema itself: It makes a reference to sourcefile/@name

org.jacoco.doc/docroot/doc/changes.html
marchof: Thx, fixed.

org.jacoco.doc/docroot/doc/changes.html
marchof: Thx, fixed.

org.jacoco.report/src/org/jacoco/report/xml/report.dtd
Godin: Ok

org.jacoco.report/src/org/jacoco/report/xml/XMLFormatter.java
Godin: This was not request of change / addition - just wanted to share thought with you. Let's handle this later when / if we'll see emerging need for real 😉

org.jacoco.doc/docroot/doc/changes.html
Godin: Well, you right - not really a correct place.

marchof:
@Godin Rebased on master. So should me mergeable now.

Godin:
Was about to resolve conflict by myself, but you've been faster 😆 Thanks!

PR #705, opened 2018-07-02
Update asm version to 6.2; support JDK 11

mmoayyed:
Please see https://asm.ow2.io/versions.html

> experimental support for the upcoming JDK 11 features (nest mates and ConstantDynamic)


Godin:
Please do not open PRs for changes that don't even pass unit tests.
Also should be noted that update of ASM alone is not enough to fully support JDK 11.
I'm closing this PR in favor of #706

PR #706, opened 2018-07-02
Upgrade ASM to 6.2

Godin:


Godin:
@marchof could you please have a look? I didn't update changelog to first discuss below points.

----

All changes in our tests, except one about `ClassWriter.COMPUTE_FRAMES`, are because starting from version 6.1 ASM is more strict about allowed arguments.

----

Should be noted that some regressions were already found in ASM 6.2, however to me seems that JaCoCo can't be affected by them, but might be worth double-checking - AFAIK as of 2 July 2018 from the list of issues (https://gitlab.ow2.org/asm/asm/issues) and changes after ASM 6.2 (https://gitlab.ow2.org/asm/asm/compare/ASM_6_2...master):
* https://gitlab.ow2.org/asm/asm/issues/317835 - we don't use `ClassWriter.COMPUTE_MAXS` and therefore `MethodWriter.COMPUTE_MAX_STACK_AND_LOCAL_FROM_FRAMES`
* https://gitlab.ow2.org/asm/asm/issues/317837 - we use `GeneratorAdapter` only in tests

And seems that while https://gitlab.ow2.org/asm/asm/issues/317833 was fixed after 6.2, problem was already there with ASM 6.0, also we don't use `ClassWriter.COMPUTE_MAXS` and don't change number of arguments of methods.

----

Also should be noted that ASM 6.2 contains experimental support for the upcoming JDK 11 features (nest mates [JEP-181](http://openjdk.java.net/jeps/181) and ConstantDynamic [JEP-309](http://openjdk.java.net/jeps/309) - see https://mail.ow2.org/wws/arc/asm/2018-05/msg00004.html Furthermore `ClassReader` doesn't reject JDK 11 classes even if flag `Opcodes.ASM7_EXPERIMENTAL` is not used - see https://mail.ow2.org/wws/arc/asm/2018-05/msg00000.html

While seems that this allows to use our `ModifiedSystemClassRuntime` under JDK 11 and perform on-the-fly instrumentation of classes compiled into bytecode version 55 without other changes (`ClassReader` reads new ConstantDynamic constant pool entries and new nest mates attributes, `ClassWriter` copies them), just an update of ASM alone is clearly not enough for proper full support of JDK 11:

* `ContentTypeDetector` should be updated for generation of report and offline instrumentation for classes compiled into bytecode version 55
* `TryWithResourcesJavacFilter` also requires update - need to finish work started in #669
* need to have a look on JEP 12 ( http://openjdk.java.net/jeps/12 ) as it changes meaning of a minor segment of bytecode version

Also even if [JDK 11 is now in Rampdown Phase One](http://mail.openjdk.java.net/pipermail/jdk-dev/2018-June/001509.html), AFAIK there is no strong guarantees that bytecode won't change, and this is also why support for JDK 11 features is experimental in ASM 6.2. IMO the best what we can do - is also "experimental support", i.e. without any guarantees. WDYT? In any case, I would prefer to separate ASM update from JDK 11 support.

marchof:
@Godin +1 for adding experimental JDK 11 support in a separate PR

Godin:
@marchof and just to be sure - what about known regressions in ASM 6.2? do you agree with my analysis that they don't affect us? any concerns about possible others, given that 6.1 was a massive refactoring?

marchof:
@Godin I agree with you that the regressions do not affect usage in JaCoCo. And hey, our tests are green ;)

Godin:
@marchof well, as [history shows](https://github.com/jacoco/jacoco/commit/04e13929cff687d829c7704e345ccabb85ee6723) - our tests are not totally ehaustive and bulletproof :wink:

I updated changelog and will do the merge. Thank you for review!

Godin:
@marchof in fact while

> `ClassReader` reads new ConstantDynamic constant pool entries and new nest mates attributes, `ClassWriter` copies them

is correct, [`ClassVisitor` without `Opcodes.ASM7_EXPERIMENTAL` throws `UnsupportedOperationException`](https://gitlab.ow2.org/asm/asm/blob/ASM_6_2/asm/src/main/java/org/objectweb/asm/ClassVisitor.java#L157) when [`ClassReader` encounters/visits nest mates](https://gitlab.ow2.org/asm/asm/blob/ASM_6_2/asm/src/main/java/org/objectweb/asm/ClassReader.java#L541) :disappointed:

Godin:
For the record: ticket to upgrade ASM in Eclipse Orbit - https://bugs.eclipse.org/bugs/show_bug.cgi?id=536922

PR #708, opened 2018-07-05
Remove unused constants in XMLFormatter

marchof:
It was probably to late in the night when I rebased #652 on master.


marchof:
@Godin May I ask you to have a look again?

> Reminder to myself: Don't rebase after midnight.

Godin:
@marchof do not reproach yourself, also probably I was too fast with review :wink: seen unused fields, but right after merge.

BTW looking at #702 and #652, seems that while we do check correctness of `DOCTYPE` in html report, we do not have such test for xml report, IMO worth to be added.

PR #710, opened 2018-07-13
Add Filter for Methods generated by AspectJ

larsgrefer:
All methods with the custom attribute `org.aspectj.weaver.AjSynthetic` are fully ignored.
This includes:
* `private static void ajc$preClinit()`
* `private static void ajc$postClinit()`
* `public static boolean hasAspect()`
* `public static <AspectClass> aspectOf()`

The calls to `ajc$preClinit()` and `ajc$postClinit()` which are inserted in the static initializer `<clinit>` are ignored.

The `$AjcClosure`-classes generated for _around_-advices are fully ignored.


marchof:
@larsgrefer Thanks for the contribution! Do you have some background information about this filter? We're not experts in AspectJ  (and I have no practical experience at all):

- In what specific situations does it improve the coverage report? Or is it a generic solution which works for extra code generated by AspectJ?
- Is it specific to a certain AspectJ use case?
- Is it specific to a AspectJ version or does it work for all versions?
- Is the a link where we can understand the background of this filter?
- It can probably only used with compile time weaving, right?

Godin:
@larsgrefer in addition to answers on @marchof questions, could you please also provide small example of project to be able to see end-to-end experience before/after this change?

larsgrefer:
For each class, which contains AOP advices, `ajc` generates the following additional bytecode:

* The methods `aspectOf()`, `hasAspect()` and `ajc$postClinit()` are created
* The method `<clinit>()` is created or modified and contains this at its end:
```java
        try {
            ajc$postClinit();
        } catch (Throwable var1) {
            ajc$initFailureCause = var1;
        }
```
For each class, which contains methods, that are advised by an advice, the following is generated:

* The method `ajc$preClinit()` is created
* The method `<clinit>()`is created or modified and contains a call to `ajc$preClinit()` as first instruction.
* The bytecode of the advised method is modified to call the advices

Some of the created methods have the `ACC_SYNTHETIC` flag and some have a custom attribute named `org.aspectj.weaver.AjSynthetic`.

This filter does the following:

* If the method has the `org.aspectj.weaver.AjSynthetic` attribute -> ignore the complete method
(In combination with the existing SyntheticFilter, all ajc-specific created methods should be ignored)
* If the method `<clinit>()` exists and contains a call to `ajc$preClinit()` -> ignore everything up to this call.
  * If the call is also the last real instruction -> ignore the complete `<clinit>()`-method
* If the method `<clinit>()` exists and contains a call to `ajc$postClinit()` -> ignore the call and everything after it.
  * If the call is also the first real instruction -> ignore the complete `<clinit>()`-method
  * If the method also had a call to `ajc$preClinit()` and there is no user code beween the calls -> ignore the complete `<clinit>()`-method

>In what specific situations does it improve the coverage report? Or is it a generic solution which works for extra code generated by AspectJ?
>Is it specific to a certain AspectJ use case?

It improves the coverage reports for classes with advice implementations and for classes with advised methods.
For advice implementations, this filter should be sufficient.
For advised methods, this filter works best for _before_ and _afterReturning_ advices

>Is it specific to a AspectJ version or does it work for all versions?

I've tested this with ajc 1.9.1, but it should work for older versions too.

>Is the a link where we can understand the background of this filter?

What exactly do you mean?

>It can probably only used with compile time weaving, right?

This filter is intended to be used with _compile-time_ or _post-compile_ weaving

Godin:
> If the method also had a call to `ajc$preClinit()` and there is no user code beween the calls -> ignore the complete <clinit>()-method

This sounds really suspicious: Can there be non `ajc` code before/after these calls?

>> Is the a link where we can understand the background of this filter?
>
> What exactly do you mean?

Link on some documentation that explains constructions which you wanna ignore.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private void checkStaticInitializer(MethodNode methodNode, IFilterOutput output) {
+        AbstractInsnNode preClinitNode = null;
+
+        for (AbstractInsnNode node = methodNode.instructions.getFirst();
+             node != null;
+             node = node.getNext()) {
+
+            if (node.getOpcode() == Opcodes.INVOKESTATIC) {
```
Godin: Inversion of condition and usage of `continue` will reduce nesting of this code and hence might improve its readability. Also note that we have `AbstractMatcher` which simplifies matching of bytecode. Maybe it can be used here?

larsgrefer:
> This sounds really suspicious: Can there be non ajc code before/after these calls?

The call to `ajc$preClinit()` is always inserted at the beginnging of the `<clinit>()`-method, and the `try { ajc$postClinit()}` is always inserted at the end.

I added this to my classes for testing:
```java
    private static final Object foo = new Object();
    static {
        System.out.println("bar");
    }
```

this is the Bytecode:

```
  static {};
    descriptor: ()V
    flags: ACC_STATIC
    Code:
      stack=2, locals=0, args_size=0
         0: invokestatic  #117                // Method ajc$preClinit:()V
         3: new           #3                  // class java/lang/Object
         6: dup
         7: invokespecial #1                  // Method java/lang/Object."<init>":()V
        10: putstatic     #4                  // Field foo:Ljava/lang/Object;
        13: getstatic     #5                  // Field java/lang/System.out:Ljava/io/PrintStream;
        16: ldc           #6                  // String bar
        18: invokevirtual #7                  // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        21: return
      LineNumberTable:
        line 9: 3
        line 11: 13
        line 12: 21
```

```
  static {};
    descriptor: ()V
    flags: ACC_STATIC
    Code:
      stack=2, locals=1, args_size=0
         0: new           #11                 // class java/lang/Object
         3: dup
         4: invokespecial #1                  // Method java/lang/Object."<init>":()V
         7: putstatic     #12                 // Field foo:Ljava/lang/Object;
        10: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
        13: ldc           #13                 // String bar
        15: invokevirtual #14                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        18: nop
        19: invokestatic  #114                // Method ajc$postClinit:()V
        22: goto          30
        25: astore_0
        26: aload_0
        27: putstatic     #107                // Field ajc$initFailureCause:Ljava/lang/Throwable;
        30: return
      Exception table:
         from    to  target type
            19    25    25   Class java/lang/Throwable
      StackMapTable: number_of_entries = 2
        frame_type = 89 /* same_locals_1_stack_item */
          stack = [ class java/lang/Throwable ]
        frame_type = 4 /* same */
      LineNumberTable:
        line 33: 0
        line 35: 10
        line 36: 18
```

I think everything that might be before the `ajc$preClinit`-call or after the `ajc$postClinit`-call, doens't have corresponding source code but must be generated in some way

Godin:
@larsgrefer in https://github.com/jacoco/jacoco/pull/710#issuecomment-404803569 :

> if the method <clinit>() exists and contains a call to ajc$postClinit()
> If the method also had a call to ajc$preClinit() and there is no user code beween the calls -> ignore the complete <clinit>()-method

in https://github.com/jacoco/jacoco/pull/710#issuecomment-404816305 you show **two** `clinit` methods, and none of them has both methods at the same time. Do I miss something?

> I think everything that might be before the `ajc$preClinit`-call or after the `ajc$postClinit`-call, doens't have corresponding source code but must be generated in some way

and similarly: why in this comment you say "before or after", while in previous as well as in implementation - "complete method" ?

Also from this comment - this is just an assumption and not a practical observation?

> I added this to my classes for testing

Sounds like something that can be shared as a response on https://github.com/jacoco/jacoco/pull/710#issuecomment-404797545 😉 

Executable example will definitely be helpful for understanding.

larsgrefer:
I've created an executable example project: https://github.com/larsgrefer/aspectj-jacoco-example

The project requires this branch to be installed in your local maven repository

You can run the whole project with the command `./gradlew test jacocoTestReport`
When adding `-Dorg.gradle.debug=true` as additional argument, you can debug the filter on port 5005

- `build/javac/main` will contain the class files after compilation with javac, but before weaving with ajc
- `build/classes/main` will contain the class files after weaving with ajc (this will be the input for jacoco)
- `reports/jacoco/test` will contain the coverage report

larsgrefer:
@Godin I've added unit tests and provided an example project as you requested

Also I've rebased this branch on the current master branch in order to resolve the merge conflicts.

Are there any open questions or TODO's left, which prevent this PR from being merged?

larsgrefer:
@Godin @marchof Could you please review this PR again?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
```diff
@@ -0,0 +1,206 @@
...
+     */
+    private void checkStaticInitializer(MethodNode methodNode, IFilterOutput output) {
+        MethodInsnNode preClinitNode = null;
+        MethodInsnNode postClinitNode = null;
+
+        for (AbstractInsnNode node = methodNode.instructions.getFirst();
+             node != null;
+             node = node.getNext()) {
+
+            if (node.getOpcode() != Opcodes.INVOKESTATIC) {
```
Godin: Can you finally answer question asked during previous review ( https://github.com/jacoco/jacoco/pull/710#discussion_r202324700 ):
Why not use `AbstractMatcher`? which in particular has `skipNonOpcodes`

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    private boolean isWellKnownSyntheticMethod(MethodNode methodNode) {
+        if (methodNode.name.equals("aspectOf") && isAjSynthetic(methodNode)) {
+            return true;
+        }
+        if (methodNode.name.equals("hasAspect") && isAjSynthetic(methodNode)) {
+            return true;
+        }
+        if (methodNode.name.equals("ajc$postClinit") && isAjSynthetic(methodNode)) {
```
Godin: What's the point of this?

According to [your example](https://github.com/larsgrefer/aspectj-jacoco-example/tree/afcea2a2a0d7ab78f5fef26d7ec26c7f7f0a129e) (https://github.com/jacoco/jacoco/pull/710#issuecomment-404971814) `ajc$postClinit` has `ACC_SYNTHETIC` flag

```
$ javap -v -p build/classes/java/main/de/larsgrefer/demo/Combined.class | grep -A 2 "void ajc\$postClinit"
  private static void ajc$postClinit();
    descriptor: ()V
    flags: ACC_PRIVATE, ACC_STATIC, ACC_SYNTHETIC
```

and therefore already filtered by `SyntheticFilter`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
```diff
@@ -0,0 +1,206 @@
...
+            return true;
+        }
+        if (methodNode.name.equals("ajc$preClinit") && ((methodNode.access & Opcodes.ACC_SYNTHETIC) != 0)) {
```
Godin: Methods with `ACC_SYNTHETIC` flag already handled by `SyntheticFilter`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java
```diff
@@ -44,4 +50,14 @@ public void filter(final MethodNode methodNode,
 				methodNode.instructions.getLast());
 	}
 
+	private static final Pattern ASPECTJ_AROUND_BODY_PATTERN = Pattern.compile(".*_aroundBody\\d+");
+
+	/**
+	 * @see AspectjFilter
+	 */
+	private boolean isHandledByAspectJFilter(MethodNode methodNode) {
```
Godin: Can't find anything about `_aroundBody` in `AspectjFilter`.

larsgrefer: When a method `foo()` is adivsed with an around advice, the bytecode of the method implementation is moved to `foo_aroundBody0()`. This method has the `ACC_SYNTHETIC` flag, but must not be filtered because it contains actual user code.

I'll add a comment to this method to make this more clear.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
larsgrefer: You're right. I've simplified the filter to only handle the custom `org.aspectj.weaver.AjSynthetic` attribute

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
larsgrefer: You're right. This method is already filtered. I've simplified the AspectJ filter accordingly.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AspectjFilter.java
larsgrefer: The detection of calls to `ajc$preClinit` and `ajc$postClinit` now uses `AbstractMatcher` subclasses.

larsgrefer:
@Godin I've adressed your feedback

larsgrefer:
@Godin Any chance to bring this forward?

Issue #312, opened 2015-05-20
Use standard Maven layout?


marchof:
Investigate whether can use a standard Maven layout, especially joining main and test into a single project. There might be issues with running JaCoCo tests with code coverage.

Godin:
This indeed doesn't seem to be possible since will be impossible to use JaCoCo which is result of the build itself - chicken and egg problem...

Godin:
Also current layout is not that "non standard", actually not uncommon.

PR #711, opened 2018-07-14
Extract "validation" tests into separate modules

Godin:
## Defaults

* module `org.jacoco.core.test.validation.java5` now contains what previously was in `org.jacoco.core.test/src/org/jacoco/core/test/(validation|filter)`
* module `org.jacoco.core.test.validation.java7` what previously was in `org.jacoco.core.test/src-java7`
* module `org.jacoco.core.test.validation.java8` what previously was in `org.jacoco.core.test/src-java8`

Previously manual configuration was needed for inclusion of the last two sources into build and to open them in IntelliJ IDEA. Now by default all modules included into build and so showup in IntelliJ IDEA.

By default the last two will be compiled into bytecode version 51 (Java 7) and 52 (Java 8) respectively, all other into 49 (Java 5).

And build is simple as

```
$ java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

$ mvn clean verify
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 8                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 49

$ javap -v org.jacoco.core.test.java7/target/classes/org/jacoco/core/test/filter/TryWithResourcesTest.class | grep major
  major version: 51

$ javap -v org.jacoco.core.test.java8/target/classes/org/jacoco/core/test/validation/LambdaExpressionsTest.class | grep major
  major version: 52
```

Should be noted that JDKs 9 - 12 can't compile into bytecode versions lower than 50 (Java 6) - see [JEP-182](http://openjdk.java.net/jeps/182). So when they are used modules `org.jacoco.core.test.java7` and `org.jacoco.core.test.java8` will be compiled as above, however all other modules will be compiled into bytecode version 50 (Java 6):

```
$ java -version
openjdk version "9.0.4"
OpenJDK Runtime Environment (build 9.0.4+11)
OpenJDK 64-Bit Server VM (build 9.0.4+11, mixed mode)

$ mvn clean verify
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 8                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 50

$ javap -v org.jacoco.core.test.java7/target/classes/org/jacoco/core/test/filter/TryWithResourcesTest.class | grep major
  major version: 51

$ javap -v org.jacoco.core.test.java8/target/classes/org/jacoco/core/test/validation/LambdaExpressionsTest.class | grep major
  major version: 52
```

And JDK 12 soon will stop compiling into bytecode versions lower than 51 (Java 7) - see [JDK-8028563](https://bugs.openjdk.java.net/browse/JDK-8028563), from which seems that change will be in build right after EA b2 (jdk-12+2) that is currently available.

Previously manual configuration was required to do build with JDK version greater than 8.

Also could be noted that builds with JDK greater than 10 still fail with the same reasons as before, which are unrelated to these changes.

## Property `jdk.version`

Property `jdk.version` can be specified to compile and run tests using JDK that is different from current.

As before location of JDK will be selected from Maven `toolchains.xml` with respective value of `version` tag, however now values do not have `1.` prefix.

As before JDK 5 will be used for snapshots and releases.

When value of this property is `5`, `6`, `7` or `8`, defaults are the same as described above for JDK 8,
when value is `9`, `10`, `11` or `12`, defaults are the same as described above for JDKs 9 - 12:

```
$ java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

$ mvn clean verify -Djdk.version=9
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 8                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...
[INFO] Toolchain (jdk) matched:JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in compiler-plugin: JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in maven-javadoc-plugin: JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in surefire-plugin: JDK[/home/godin/.java-select/versions/9]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 50

$ javap -v org.jacoco.core.test.java7/target/classes/org/jacoco/core/test/filter/TryWithResourcesTest.class | grep major
  major version: 51

$ javap -v org.jacoco.core.test.java8/target/classes/org/jacoco/core/test/validation/LambdaExpressionsTest.class | grep major
  major version: 52
```

Additionally, when value of this property is `5` or `6`, modules `org.jacoco.core.test.java7` and `org.jacoco.core.test.java8` will be excluded:

```
$ java -version
java version "1.8.0_172"
Java(TM) SE Runtime Environment (build 1.8.0_172-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.172-b11, mixed mode)

$ mvn clean verify -Djdk.version=5
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...
[INFO] Toolchain (jdk) matched:JDK[/home/godin/.java-select/versions/5]
...
[INFO] Toolchain in compiler-plugin: JDK[/home/godin.mandrikov/.java-select/versions/5]
...
[INFO] Toolchain in maven-javadoc-plugin: JDK[/home/godin/.java-select/versions/5]
...
[INFO] Toolchain in surefire-plugin: JDK[/home/godin/.java-select/versions/5]
...
```

When value is `7`, module `org.jacoco.core.test.java8` will be excluded:

```
$ mvn clean verify -Djdk.version=7
...
Java version: 1.8.0_172, vendor: Oracle Corporation
Java home: /home/godin/.java-select/versions/8
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...
[INFO] Toolchain (jdk) matched:JDK[/home/godin/.java-select/versions/7]
...
[INFO] Toolchain in compiler-plugin: JDK[/home/godin/.java-select/versions/7]
...
[INFO] Toolchain in maven-javadoc-plugin: JDK[/home/godin/.java-select/versions/7]
...
[INFO] Toolchain in surefire-plugin: JDK[/home/godin/.java-select/versions/7]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 49

$ javap -v org.jacoco.core.test.java7/target/classes/org/jacoco/core/test/filter/TryWithResourcesTest.class | grep major
  major version: 51
```

## Property `bytecode.version`

Property `bytecode.version` can be specified to compile modules into bytecode version different from their defaults described above.

Now values of this property should not have `1.` prefix.

Modules `org.jacoco.core.test.java7` and `org.jacoco.core.test.java8` will be excluded when value of this property is `5` or `6`:

```
$ mvn clean verify -Dbytecode.version=6
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 50
```

Module `org.jacoco.core.test.java8` will be excluded when value of this property is `7`:

```
$ mvn clean verify -Dbytecode.version=7
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 51
```

Additionally, it can be combined with property `jdk.version`:

```
$ mvn clean verify -Dbytecode.version=5 -Djdk.version=9
...
[ERROR] error: Source option 1.5 is no longer supported. Use 1.6 or later.
[ERROR] error: Target option 1.5 is no longer supported. Use 1.6 or later.
...
```

Also should be noted that `maven-shade-plugin` and `maven-plugin-plugin` can't process bytecode versions 54 (Java 10), 55 (Java 11) and 56 (Java 12). So when value of this property is `10`, `11` or `12`, only test modules (`org.jacoco.*.test.*`) will be compiled into specified bytecode version and others will be compiled into 53 (Java 9):

```
$ java -version
openjdk version "12-ea" 2019-03-19
OpenJDK Runtime Environment 19.3 (build 12-ea+2)
OpenJDK 64-Bit Server VM 19.3 (build 12-ea+2, mixed mode)

$ mvn clean verify -Dbytecode.version=10

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 53

$ javap -v org.jacoco.core.test/target/classes/org/jacoco/core/JaCoCoTest.class | grep major
  major version: 54
```

```
$ java -version
openjdk version "12-ea" 2019-03-19
OpenJDK Runtime Environment 19.3 (build 12-ea+2)
OpenJDK 64-Bit Server VM 19.3 (build 12-ea+2, mixed mode)

$ mvn clean verify -Dbytecode.version=11

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 53

$ javap -v org.jacoco.core.test/target/classes/org/jacoco/core/JaCoCoTest.class | grep major
  major version: 55
```

```
$ java -version
openjdk version "12-ea" 2019-03-19
OpenJDK Runtime Environment 19.3 (build 12-ea+2)
OpenJDK 64-Bit Server VM 19.3 (build 12-ea+2, mixed mode)

$ mvn clean verify -Dbytecode.version=12

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 53

$ javap -v org.jacoco.core.test/target/classes/org/jacoco/core/JaCoCoTest.class | grep major
  major version: 56
```

## Property `ecj`

Property `ecj` can be specified to compile all modules using ECJ.

For the time being with ECJ we compile all modules into bytecode version 52 (Java 8) unconditionally (and we were testing only this configuration in Travis even before these changes), however it can be combined with properties `bytecode.version` and `jdk.version` for selection of modules and JDK for execution of tests:

```
$ java -version
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)

$ mvn clean verify -Decj -Dbytecode.version=7 -Djdk.version=9
...
[INFO] Reactor Build Order:
...
[INFO] JaCoCo :: Test :: Core                                             [jar]
[INFO] JaCoCo :: Test :: Core :: Validation                               [pom]
[INFO] JaCoCo :: Test :: Core :: Validation Java 5                        [jar]
[INFO] JaCoCo :: Test :: Core :: Validation Java 7                        [jar]
[INFO] JaCoCo :: Test :: Report                                           [jar]
...
[INFO] Toolchain (jdk) matched:JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in compiler-plugin: JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in maven-javadoc-plugin: JDK[/home/godin/.java-select/versions/9]
...
[INFO] Toolchain in surefire-plugin: JDK[/home/godin/.java-select/versions/9]
...

$ javap -v org.jacoco.core/target/classes/org/jacoco/core/JaCoCo.class | grep major
  major version: 52

$ javap -v org.jacoco.core.test/target/classes/org/jacoco/core/JaCoCoTest.class | grep major
  major version: 52

$ javap -v org.jacoco.core.test.java8/target/classes/org/jacoco/core/test/validation/LambdaExpressionsTest.class | grep major
  major version: 52
```


Godin:
@marchof could you please have a look on these changes? if all fine, I will also update documentation.

As expected individual profiles are now shorter (https://github.com/jacoco/jacoco/pull/709#issuecomment-404887586), but there are anyway some duplication in `pom.xml`s (https://github.com/jacoco/jacoco/pull/709#issuecomment-404637295) as a necessary evil to have good defaults and customizations. This solves quite some headaches that I had during testing of different JDKs and filters.

marchof:
@Godin Thanks for taking up this effort!

I wonder how the dependencies should look like: Currently the base infrastructure is part of Java5 (like source). This would mean that e.g. Kotlin will have a dependency on Java5. I would prefer to move all common infraststructure to `org.jacoco.core.test`. Some tests can then be extended. E.g. `FramesTest` should get a common abstract base class. Then we can also do tests on Java7 and so on. But I can work on this in a separate PR.

Godin:
@marchof 

> I would prefer to move all common infraststructure to `org.jacoco.core.test`.

Good point - done.

> Some tests can then be extended. E.g. FramesTest should get a common abstract base class. Then we can also do tests on Java7 and so on.

You mean to run `FramesTest` on targets from `java7` module?

> But I can work on this in a separate PR.

Indeed will be better to do after this PR in order to unblock addition of tests for Kotlin.

I added update for documentation, so could you please do re-review?

marchof:
@Godin Thanks for fixing dependencies!

> You mean to run FramesTest on targets from java7 module?

This is what I wanted to propose. Separate PR fine to me as this would be a new anyways.

org.jacoco.build/pom.xml
```diff
@@ -708,37 +709,67 @@
       </build>
     </profile>
 
-    <!-- This profile is used to launch tests with compilation into specific bytecode version. -->
+    <!-- -->
```
marchof: Is this empty comment on purpose?

org.jacoco.build/pom.xml
```diff
@@ -776,12 +809,14 @@
       </build>
     </profile>
 
+    <!-- -->
```
marchof: Empty comment?

org.jacoco.core.test.java7/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+  <parent>
+    <groupId>org.jacoco</groupId>
+    <artifactId>org.jacoco.tests</artifactId>
+    <version>0.8.2-SNAPSHOT</version>
+    <relativePath>../org.jacoco.tests</relativePath>
+  </parent>
+
+  <artifactId>org.jacoco.core.test.java7</artifactId>
+
+  <!--<name></name>-->
```
marchof: Commented tag? Remove?

org.jacoco.core.test.java7/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+
+  <properties>
+    <jacoco.includes>org.jacoco.core.*</jacoco.includes>
```
marchof: I would prefer to not record coverage for integration tests at all. In this case this property is not relevant.

org.jacoco.tests/pom.xml
```diff
@@ -27,13 +27,12 @@
 
   <modules>
     <module>../org.jacoco.core.test</module>
+    <module>../org.jacoco.core.test.java5</module>
     <module>../org.jacoco.report.test</module>
     <module>../org.jacoco.agent.rt.test</module>
     <module>../org.jacoco.agent.test</module>
     <module>../org.jacoco.ant.test</module>
     <module>../org.jacoco.cli.test</module>
-    <module>../org.jacoco.examples.test</module>
```
marchof: Why do we need to move these tests?

marchof:
@Godin Do we still need the constructor `ValidationTestBase(final String srcFolder, final Class<?> target)`?

marchof:
@Godin Can we add Eclipse configs to the new projects? At least as "plain project" so it can be easily imported.

org.jacoco.tests/pom.xml
Godin: Because otherwise they will be executed before modules from profiles.

marchof: Sorry for my stupid question: And why would this be a problem?

org.jacoco.core.test.java7/pom.xml
Godin: They are not included into final report.

All classes are included into instrumentation and agent is enabled by default - see https://github.com/jacoco/jacoco/blob/753d6532d56892d037ff7958ea30c9bc27f6e8e4/org.jacoco.tests/pom.xml#L49-L70

Thus if we want to exclude all classes from instrumentation, then this should be replaced by `jacoco.excludes`? Or if we want to completely disable agent, then `jacoco.skip`?

org.jacoco.tests/pom.xml
Godin: Because `org.jacoco.examples.test` and especially `jacoco-maven-plugin.test` are the slowest ones, while perform tests on a way higher level than `org.jacoco.core.test.*`.

org.jacoco.core.test.java7/pom.xml
marchof: Yep, I think JaCoCo agent should not be in the game for integration tests. Targets are instrumented programmatically anyways during the test. So `jacoco.skip` would be the best fit here (and `jacoco.includes` can be removed at all).

org.jacoco.core.test.java7/pom.xml
marchof: Right, then the new modules should also have proper names. What about this?

    JaCoCo :: Test :: Core :: Integration Java 5

org.jacoco.build/pom.xml
Godin: Yep - for visual separation of profiles that follow this comment from ones that are above.

org.jacoco.build/pom.xml
Godin: Yep - for visual separation of profile that follows this comment from one that is above.

org.jacoco.tests/pom.xml
marchof: So you mean like "fail fast"? Where is that important? For local develop/test iterations I mostly run specific modules only anyways. For our CI all tests are executed anyways.

I just find it rather confusing that these modules are now included in specific JDK profiles.

org.jacoco.build/pom.xml
marchof: Some profiles are separated but some are not. I assume these are groups of profiles, right? Maybe add a short description into the comment for the profile group like with the profiles at the top?

org.jacoco.tests/pom.xml
Godin: > So you mean like "fail fast"?

Yes.

> Where is that important?

For develop/test iterations I run specific tests in IDE, but then before commit I do verification build in command-line and it better fail on one of `org.jacoco.core.test.*` than on `jacoco-maven-plugin.test`. In later case, I'll be forced to manually select modules in command-line or in IDE.

> I just find it rather confusing that these modules are now included in specific JDK profiles.

I see only two options
1. add comment
2. introduce one more module with packaging `pom` to group `jdk` specific modules

org.jacoco.core.test.java7/pom.xml
Godin: Done.

org.jacoco.core.test.java7/pom.xml
Godin: Replaced `jacoco.includes` by `jacoco.skip`.

org.jacoco.tests/pom.xml
marchof: For clean dependencies I was thinking about option 2) anyways. Resulting in the following modules:

* `org.jacoco.core.test` - pure unit tests for `org.jacoco.core`
* `org.jacoco.core.test.validation` - base infrastructure used by the specific tests
* `org.jacoco.core.test.validation.xxx` - language version specific tests, depends on `org.jacoco.core.test.validation`

Does this make sense?

Godin:
> Do we still need the constructor `ValidationTestBase(final String srcFolder, final Class<?> target)`?

Removed.

org.jacoco.tests/pom.xml
Godin: Unfortunately we will anyway need one more module, because only modules with packaging `pom` can include other modules.

org.jacoco.doc/docroot/doc/build.html
```diff
@@ -80,67 +79,34 @@ <h2>Testing with different JDKs</h2>
...
-      &lt;version&gt;1.9&lt;/version&gt;
-      &lt;vendor&gt;sun&lt;/vendor&gt;
+      &lt;id&gt;java6&lt;/id&gt;
+      &lt;version&gt;6&lt;/version&gt;
     &lt;/provides&gt;
     &lt;configuration&gt;
-      &lt;jdkHome&gt;/usr/lib/jvm/sun-jdk-1.9&lt;/jdkHome&gt;
+      &lt;jdkHome&gt;/usr/lib/jvm/java-6-openjdk-amd64&lt;/jdkHome&gt;
     &lt;/configuration&gt;
   &lt;/toolchain&gt;
```
marchof: I was a bit confused when reading this. Maybe a comment at the end would make clear that the file is not complete om respect to the description that is following:

org.jacoco.tests/pom.xml
marchof: Ok, I see. Then we would end up with something like:

* `org.jacoco.core.test.validation` - aggregator of type `pom`
* `org.jacoco.core.test.validation.base` - base infrastructure
* `org.jacoco.core.test.validation.xxx` - language version specific tests

Also looks good to me. Especially as we can separate the profile magic into the new module. Maybe it can also serve as a parent POM and we can e.g. move `jacoco.skip` to it.

Godin:
> Can we add Eclipse configs to the new projects? At least as "plain project" so it can be easily imported.

@marchof you know that for Java development I mostly use IntelliJ IDEA and nowadays don't use even it, because mostly do development in C++ :wink:

I added files, hopefully with correctly linked settings and correctly specified JRE. Could you please check import into IDE?

org.jacoco.doc/docroot/doc/build.html
Godin: I wanted to reduce boilerplate/duplication that we had here before. Comment sounds good to me.

Godin: Added
```
<!-- you can specify other JDKs in this file -->
```

marchof:
@Godin Thanks for the Eclipse settings! I can now import but not compile because the shared settings declare `org.eclipse.jdt.core.compiler.compliance=1.5`. But we can sort this out separately (Eclipse setup is somewhat broken anyways).

org.jacoco.tests/pom.xml
marchof: Re module names: For consistency we should also adjust Java package names to the new module names.

Godin:
@marchof about import into Eclipse IDE - I was thinking about Oomph

org.jacoco.doc/docroot/doc/build.html
```diff
@@ -80,67 +79,37 @@ <h2>Testing with different JDKs</h2>
...
+
+  &lt;-- you can specify other JDKs in this file --&gt;
+
 &lt;/toolchains&gt;
 </pre>
 
 <p>
-  Now you should be able to execute maven build with specified version of JDK:
+  to execute compilation and tests using JDK 5 or 6:
```
marchof: Shouldn't that be moved after the next section ("Location of toolchains.xml ...")?

Also should start with upper case: "To execute ..."

marchof:
@Godin Oomph might be helpful here. But let's fix this separate from this PR.

org.jacoco.core.test.validation.java5/.project
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.jacoco.core.test.java5</name>
```
marchof: Should now read `org.jacoco.core.test.validation.java5`. Same issue with the other new `org.jacoco.core.test.validation.*` projects.

org.jacoco.doc/docroot/doc/build.html
Godin: Here is how this currently looks/reads to me:

> You can set property jdk.version to use JDK specified in Maven Toolchains with respective value of version tag. For example you can create following ~/.m2/toolchains.xml file
> ...
> to execute compilation and tests using JDK 5 or 6:
> ```
>  mvn clean install -Djdk.version=5
>  mvn clean install -Djdk.version=6
> ```

Notice that above we refer to default location of `toolchains.xml`, after that show that non-default can be specified:

> Location of toolchains.xml can be set via an option:
> ```
>  mvn --toolchains path clean install -Djdk.version=version
> ```

org.jacoco.core.test.validation.java5/.project
Godin: arrrrrr

Godin: Fixed.

org.jacoco.tests/pom.xml
Godin: I added `org.jacoco.core.test.validation` as parent for `org.jacoco.core.test.validation.*`

Not sure that I like that amount of folders in root. Maybe later we can think of grouping like for example

```
validation
  |- pom.xml
  |- java5
  |    \- pom.xml
  |- java7
  |    \- pom.xml
  \- java8
       \- pom.xml
```

org.jacoco.doc/docroot/doc/build.html
marchof: So for me the page currently has the following structure

1) Information about toolchain
2) Information how to run the build for specific JDK versions
3) Information about toolchain
4) Information how to run the build for specific JDK versions

Or do we need a separate `toolchain.xml` for step 4? The this should be noted.

org.jacoco.tests/pom.xml
marchof: Absolutely! The same for the unit tests. We could simplify the structure if we could move to a more Maven standard project layout. But again... separate issue.

Godin: > We could simplify the structure if we could move to a more Maven standard project layout.

In truly standard Maven layout, unit tests are located in the same module as main code, but we can't do this because we need to execute JaCoCo for JaCoCo itself - see https://github.com/jacoco/jacoco/issues/312#issuecomment-282526475

marchof: I know 😉

marchof:
@Godin Regarding package naming: Probably it doesn't make sense any more to distinguish between "filter" and "non filter" as validation tests should verify the overall functionality in respect to different language feature. Therefore I would propose to use two packages only per module:

- `org.jacoco.core.test.validation.xxx`
- `org.jacoco.core.test.validation.xxx.targets`

Side note: Some tests should probably be moved to `org.jacoco.core.test` module (like `ResizeInstructionsTest`). But again: Separate PR.

org.jacoco.build/pom.xml
Godin: I added comments.

PR #714, opened 2018-07-23
Validation tests should use source filename from coverage results

Godin:
As discussed in https://github.com/jacoco/jacoco/pull/709#discussion_r202812090

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/SourceTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/SourceTest.java
@@ -12,7 +12,6 @@
 package org.jacoco.core.test.validation;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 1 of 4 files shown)


Godin:
@marchof while we haven't yet added/merged tests for Kotlin, I did a test and this will work for them. Could you please review?

Godin:
@marchof fixed

org.jacoco.core.test/src/org/jacoco/core/test/validation/SourceTest.java
```diff
@@ -76,12 +76,4 @@ public void testDuplicateTag() throws IOException {
 		new Source(new StringReader(src));
 	}
 
-	@Test
-	public void testGetSourceFor() throws IOException {
-		final Source s = Source.getSourceFor("src", SourceTest.class);
-		// Here we are. $line-testGetSourceFor$
-		final String l = s.getLine(s.getLineNumber("testGetSourceFor"));
-		assertTrue(l, l.contains("Here we are."));
```
marchof: Damn, I've overlooked this: There is another unused import for `assertTrue` now.

Godin: @marchof fixed, also checked that there is no other unused imports in modified files :wink:

marchof: 👍

PR #719, opened 2018-08-03
Add experimental support for Java 11 class files

Godin:


marchof:
@Godin Is there a reason that we use native ASM support for Java 11 but still use the `BytecodeVersion` trick for Java 10?

marchof:
@Godin I found two more places where the new Java version should be added:

* `InstrSupportTest.needFrames_should_return_true_for_versions_greater_than_or_equal_to_1_6()`
* `ClassFileVersionsTest`

Godin:
@marchof updated `InstrSupportTest` and `ClassFileVersionsTest`. IMO we should find a way to reduce number of places that require non functional mechanical update.

Cleanup from `V10` is unrelated to Java 11, so will do it separately.

marchof:
@Godin You're right about reducing numbers of places, where we encode class file versions. Maybe something like

    int[] SUPPORTED_VERSIONS;

in `BytecodeVersion` could be a solution.

Issue #704, opened 2018-06-29
[Kotlin] The last case in when for seal object is never show coverage without already test


thuypt:
I wrote a `when` condition for `seal` object and implement test for all cases, but Jacoco doesn't show coverage for the last case.

ex:
Code
```
        when (index) {
            is Index.First -> instayWifi.clear()
            is Index.Second -> instayWifi.setUpInStayWifiCredential()
        }

```

Test
```
       whenSampleCoverrage.apply(Index.Second)
        verify(instayWifi).setUpInStayWifiCredential()

        whenSampleCoverrage.apply(Index.First)
        verify(instayWifi).clear()
```

In the coverrage result, it doesn't show coverage for last case 
<img width="597" alt="screen shot 2018-06-29 at 11 59 09 am" src="https://user-images.githubusercontent.com/6264092/42074100-88d71c8e-7b94-11e8-847d-97ce86f3e331.png">

JaCoCo version: 0.8.1
Operating system: MacOS
Tool integration: AndroidStudio/Gradle/Java/Kotlin

### Expected behaviour
The last case in when should show coverage after call testing.

### Actual behaviour
The last case in when never show coverage after call testing.

marchof:
@thuypt My guess is that you're missing the case that there is no match at all.

thuypt:
@marchof I had same think with you at the first time, and I run debug for testing then, it already jump to the last case. You can see in my test
```      
 whenSampleCoverrage.apply(Index.Second)
```

marchof:
@thuypt I'm talking about the fall through case, something like:

    whenSampleCoverrage.apply(Index.Third)

This is probably the missed branch.

thuypt:
@marchof Just only two object define in `Index`, 
```
sealed class Index {
    object First : Index()
    object Second : Index()
}
```
btw, I think even we have `Index.Third`, Jacoco should mark the line testing with `Index.Second` are coverage.

marchof:
@thuypt JaCoCo does not know about the semantics of the statement. It only sees a missed branch -- and reports it. The same holds true for Java switches on enums.

We have this issue on our [filtering wish list](https://github.com/jacoco/jacoco/wiki/FilteringOptions) but for now we don't have a solution as we don't have other classes at hand during analysis (e.g. to check whether there are more members in the enum).

Godin:
For the completeness of picture:

[...]

PR #720, opened 2018-08-03
Add module for Kotlin validation tests

Godin:
PR #709 contains more/other tests, however
* it was done before #711 and so doesn't execute them on JDKs and bytecode versions other than 8
* these tests require some rework/cleanup

This PR
* allows to compile Kotlin source files into all bytecode versions supported by `kolinc` (6 and 8) 
* and execute them in all JDKs >= 6
* uses more recent version of Kotlin - most recent as of today
* allows addition of new filters and corresponding tests (#704) without waiting cleanup of tests in #709
* adds basic test that utilizes/validates #714
* adds check of presence of license info in Kotlin source files

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -265,6 +265,7 @@
         <filtering>false</filtering>
         <excludes>
           <exclude>**/*.java</exclude>
+          <exclude>**/*.kt</exclude>
           <exclude>**/*.properties</exclude>
         </excludes>
       </resource>
@@ -554,6 +555,7 @@
             <configuration>
               <target>
                 <fileset dir="${basedir}" includes="**/*.java,**/*.xml,**/*.bsh" excludes="target/**,.idea/**,nb-configuration.xml" id="missinglicense.fileset">
+                  <include name="**/*.kt"/>
                   <not>
                     <and>
                       <contains text="Copyright (c) 2009, 2018 Mountainminds GmbH &amp; Co. KG and Contributors"/>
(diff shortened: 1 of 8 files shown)


marchof:
@Godin I wonder whether we should create a separate module for these tests:

    org.jacoco.core.validation.test.kotlin

As we will probably add tests for more JVM languages (Scala, Groovy, Lombok, ...) in future I would prefer a structure where we have one module per language instead of mixing all the compilers into the Java validation tests.

What do you think about this?

Godin:
@marchof I renamed Kotlin module.
However in case of Lombok really doubt about ROI of addition and hence future maintenance.

marchof:
@Godin 👍 Thanks!

Sure, Lombok was just an example for other languages.

org.jacoco.build/pom.xml
```diff
@@ -265,6 +265,7 @@
         <filtering>false</filtering>
         <excludes>
           <exclude>**/*.java</exclude>
+          <exclude>**/*.kt</exclude>
```
marchof: I wonder whether Kotlin specific configurations should really go into our global parent pom (thus theoretically also affecting also our productive modules). If we keep adding integration tests for other languages it might become confusing here. I personally would prefer to move Kotling specific config to `org.jacoco.core.test.validation.kotlin`. Sure, probably need an new property.

org.jacoco.build/pom.xml
```diff
@@ -554,6 +555,7 @@
             <configuration>
               <target>
                 <fileset dir="${basedir}" includes="**/*.java,**/*.xml,**/*.bsh" excludes="target/**,.idea/**,nb-configuration.xml" id="missinglicense.fileset">
+                  <include name="**/*.kt"/>
```
marchof: Same as above: New property for additional files to check?

Godin: @marchof same as for above: here we define generic behavior for source files extensions,  highly unlikely, I would even say impossible, that same extension will have different behavior. `bsh` also not presented in all modules. IMO another way is far more confusing, more complex than really needed and error prone. Let's keep it simple.

marchof: Ok!

PR #721, opened 2018-08-06
Add filter for Kotlin when-expressions that list all cases of sealed class

### LINKED ISSUE #704: [Kotlin] The last case in when for seal object is never show coverage without already test

Godin:
Fixes #704


org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/KotlinWhenExpressionTest.java
```diff
@@ -0,0 +1,88 @@
...
+	public KotlinWhenExpressionTest() {
+		super(KotlinWhenExpressionTarget.class);
+	}
+
+	/**
+	 * {@link KotlinWhenExpressionTarget#whenSealed(KotlinWhenExpressionTarget.Sealed)}
+	 */
+	@Test
+	public void whenSealed() {
+		assertLine("whenSealed.when", ICounter.FULLY_COVERED, 0, 0);
```
marchof: Probably in conflict with #722.

PR #722, opened 2018-08-06
`assertLine(String, int)` should imply zero number of branches

Godin:
IMO usage of

https://github.com/jacoco/jacoco/blob/bbee0397a1ff4e39bab42c731531d02cc404fc2b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java#L133

is error-prone since it doesn't imply zero number of branches, especially when presence of branch is not obvious like in Kotlin for `when` expressions and `lateinit`. At least I'm constantly stepping on this during implementation of filters, and miss branches that should be filtered out.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
@@ -131,22 +131,28 @@ public abstract class ValidationTestBase {
 	}
 
 	protected void assertLine(final String tag, final int status) {
-		final int nr = source.getLineNumber(tag);
-		final ILine line = sourceCoverage.getLine(nr);
-		final String msg = String.format("Status in line %s: %s",
-				Integer.valueOf(nr), source.getLine(nr));
-		final int insnStatus = line.getInstructionCounter().getStatus();
-		assertEquals(msg, STATUS_NAME[status], STATUS_NAME[insnStatus]);
+		privateAssertLine(tag, status, 0, 0);
 	}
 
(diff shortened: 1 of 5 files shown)


marchof:
@Godin Agreed! I implemented the same strategy for #718:

https://github.com/jacoco/jacoco/blob/experimental-inplace-
asserts/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java#L176

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -131,22 +131,28 @@ protected void assertMethodCount(final int expectedTotal) {
...
+			final int missedBranches, final int coveredBranches) {
 		final int nr = source.getLineNumber(tag);
 		final ILine line = sourceCoverage.getLine(nr);
-		final String msg = String.format("Branches in line %s: %s",
-				Integer.valueOf(nr), source.getLine(nr));
-		assertEquals(msg + " branches",
+		final String lineMsg = String.format("line %s: %s", Integer.valueOf(nr),
+				source.getLine(nr));
+		final int insnStatus = line.getInstructionCounter().getStatus();
+		assertEquals("Status in " + lineMsg, STATUS_NAME[status],
```
marchof: To make assertion message unambiguous I would change "status" to "instructions".

Godin: Changed.

PR #727, opened 2018-08-09
Add source filename to assertion messages

Godin:


org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -120,13 +120,15 @@ public void execute_assertions_in_comments() throws IOException {
 	private void assertCoverage(final Line line, final int insnStatus,
 			final int missedBranches, final int coveredBranches) {
 		final ILine coverage = line.getCoverage();
+		final String location = String.format("(%s:%d)",
```
marchof: What about making `Line.toString()` return `<sourcename>:<nr>`?

marchof: Then all information can be extracted from the `Line` instance passed to the assert methods.

Godin: Done

marchof:
With this nice link do we still need the source code in the message? From my point of view we can now reduce the resulting message to something like:

    Instructions (FinallyTarget.java:38): expected:<[EMPTY]> but was:<[FULLY_COVERED]>

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -121,12 +121,11 @@ private void assertCoverage(final Line line, final int insnStatus,
 			final int missedBranches, final int coveredBranches) {
 		final ILine coverage = line.getCoverage();
 
-		String msg = String.format("Instructions in %s: %s", line,
-				line.getText());
+		String msg = String.format("Instructions in (%s)", line);
```
marchof: Just `Instructions (...)`?

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -121,12 +121,11 @@ private void assertCoverage(final Line line, final int insnStatus,
...
 		final int actualStatus = coverage.getInstructionCounter().getStatus();
 		assertEquals(msg, STATUS_NAME[insnStatus], STATUS_NAME[actualStatus]);
 
-		msg = String.format("Branches in %s: %s", line, line.getText());
+		msg = String.format("Branches in (%s)", line);
```
marchof: Just `Branches (...)` (no "in")?

marchof:
I'll squash/merge as soon as the builds are green.

PR #729, opened 2018-08-13
Add filter for Kotlin when-expressions that list all cases of enum

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
```diff
@@ -355,8 +356,16 @@ private void visitSwitchTarget(final Label label, final int branch) {
...
 		// Wire jumps:
 		for (final Jump j : jumps) {
-			LabelInfo.getInstruction(j.target).setPredecessor(j.source,
-					j.branch);
+
+			final Instruction target = LabelInfo.getInstruction(j.target);
+			final int sourceOpcode = j.source.getNode().getOpcode();
+			if ((sourceOpcode == Opcodes.LOOKUPSWITCH
+					|| sourceOpcode == Opcodes.TABLESWITCH)
+					&& ignored.contains(target.getNode())) {
```
Godin: Better to make this explicit.

marchof: It looks like we don't have unit tests in `MethodAnalyzerTest` which covers this?

Godin: As was discussed by phone: this even should not be here - filter should play role during counting and should not change CFG construction.

Godin: I believe that this can be replaced by API that is being developed in #735

marchof:
@Godin Would prefer to review this after #735 is merged to master.

Godin:
@marchof rebased

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinWhenFilter.java
```diff
@@ -11,18 +11,26 @@
...
 /**
  * Filters bytecode that Kotlin compiler generates for <code>when</code>
- * expressions which list all cases of <code>sealed class</code>, i.e. which
- * don't require explicit <code>else</code>.
+ * expressions which list all cases of <code>enum</code> or
+ * <code>sealed class</code>, i.e. which don't require explicit
+ * <code>else</code>.
  */
-public final class KotlinWhenSealedFilter implements IFilter {
+public final class KotlinWhenFilter implements IFilter {
```
Godin: @marchof we'll have `KotlinWhenStringFilter` and `KotlinWhenFilter`, but I didn't managed to find better name

marchof:
@Godin I rebased this PR. Can you please re-check before merge?

PR #730, opened 2018-08-13
StringSwitchJavacFilter fails to recognize switch that has just a few cases

Godin:
`StringSwitchJavacFilter` fails to recognize `switch` that has just a few cases

```
	private static void example(Object s) {
		switch (String.valueOf(s)) {
		case "a":
			nop("case a");
			break;
		case "b":
			nop("case b");
			break;
		}
	}
```

In this case `javac` generates `lookupswitch`, while filter expects `tableswitch` - https://github.com/jacoco/jacoco/blob/90af0853aed1cf9290acb81ed0cde1699cc00718/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java#L94

`lookupswitch` was noticed by @Forax during [my talk at Devoxx France](https://youtu.be/kADLELCH3Mk?t=28m00s) and seems to be due to following algorithm that estimates space and time costs between `lookupswitch` and `tableswitch` - https://github.com/dmlloyd/openjdk/blob/84ba2b90d46ee4b06664d32594ef0cd2038373a6/src/jdk.compiler/share/classes/com/sun/tools/javac/jvm/Gen.java#L1170


Godin:

Also maybe you can advice how to recording this in changelog - wording? bug fix or new feature?

marchof:
@Godin Ups, sorry, I overlooked your question. For me is a bug fix, because we claimed that we can filter switch in string for javac. But probably no one ever noticed this corner case. Anyway, if we want to fix the latest commit here is my proposal:

> Also filter part of bytecode generated for `switch` statements on `java.lang.String` values when only a few cases are present.

Godin:
@marchof Changelog is also useful for us :wink: Fixed in f2c79fe18060e5cdaf6ad4f921f3ee374f7f0b69

PR #733, opened 2018-08-14
Initial groovy validation setup

aalmiray:
Code developed during Hackergarten Bern August 2018 with @marchof

highres_473757340

### DIFF
--- a/org.jacoco.core.test.validation.groovy/pom.xml
+++ b/org.jacoco.core.test.validation.groovy/pom.xml
@@ -16,7 +16,7 @@
     <parent>
         <groupId>org.jacoco</groupId>
         <artifactId>org.jacoco.core.test.validation</artifactId>
-        <version>0.8.2-SNAPSHOT</version>
+        <version>0.8.3-SNAPSHOT</version>
         <relativePath>../org.jacoco.core.test.validation</relativePath>
     </parent>
 
@@ -26,7 +26,7 @@
 
     <properties>
         <gmaven.version>1.6</gmaven.version>
-        <groovy.version>2.5.2</groovy.version>
+        <groovy.version>2.5.3</groovy.version>
     </properties>
 
     <dependencies>
--- a/org.jacoco.core.test.validation.groovy/pom.xml
+++ b/org.jacoco.core.test.validation.groovy/pom.xml
@@ -9,7 +9,8 @@
    Contributors:
       Andres Almiray - initial API and implementation
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
@@ -52,9 +53,10 @@
                 <inherited>true</inherited>
                 <executions>
                     <execution>
+                        <id>compile</id>
+                        <phase>process-sources</phase>
                         <goals>
                             <goal>compile</goal>
-                            <goal>compileTests</goal>
                         </goals>
                     </execution>
                 </executions>
@@ -65,7 +67,6 @@
                             <directory>${project.basedir}/src</directory>
                             <includes>
                                 <include>**/*.groovy</include>
-                                <include>**/*.java</include>
                             </includes>
                         </source>
                     </sources>
--- a/org.jacoco.core.test.validation.groovy/pom.xml
+++ b/org.jacoco.core.test.validation.groovy/pom.xml
@@ -11,64 +11,64 @@
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-    <modelVersion>4.0.0</modelVersion>
+  <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.jacoco</groupId>
-        <artifactId>org.jacoco.core.test.validation</artifactId>
-        <version>0.8.3-SNAPSHOT</version>
-        <relativePath>../org.jacoco.core.test.validation</relativePath>
-    </parent>
+  <parent>
+    <groupId>org.jacoco</groupId>
+    <artifactId>org.jacoco.core.test.validation</artifactId>
+    <version>0.8.3-SNAPSHOT</version>
+    <relativePath>../org.jacoco.core.test.validation</relativePath>
+  </parent>
 
-    <artifactId>org.jacoco.core.test.validation.groovy</artifactId>
+  <artifactId>org.jacoco.core.test.validation.groovy</artifactId>
 
-    <name>JaCoCo :: Test :: Core :: Validation Groovy 2</name>
+  <name>JaCoCo :: Test :: Core :: Validation Groovy</name>
 
-    <properties>
-        <gmaven.version>1.6.2</gmaven.version>
-        <groovy.version>2.5.3</groovy.version>
-    </properties>
+  <properties>
+    <gmaven.version>1.6.2</gmaven.version>
+    <groovy.version>2.5.3</groovy.version>
+  </properties>
 
-    <dependencies>
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>org.jacoco.core.test</artifactId>
-            <version>${project.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy</artifactId>
-            <version>${groovy.version}</version>
-        </dependency>
-    </dependencies>
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>org.jacoco.core.test</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.codehaus.groovy</groupId>
+      <artifactId>groovy</artifactId>
+      <version>${groovy.version}</version>
+    </dependency>
+  </dependencies>
 
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.codehaus.gmavenplus</groupId>
-                <artifactId>gmavenplus-plugin</artifactId>
-                <version>${gmaven.version}</version>
-                <executions>
-                    <execution>
-                        <id>compile</id>
-                        <phase>process-sources</phase>
-                        <goals>
-                            <goal>compile</goal>
-                        </goals>
-                    </execution>
-                </executions>
-                <configuration>
-                    <targetBytecode>${groovy.targetBytecode}</targetBytecode>
-                    <sources>
-                        <source>
-                            <directory>${project.build.sourceDirectory}</directory>
-                            <includes>
-                                <include>**/*.groovy</include>
-                            </includes>
-                        </source>
-                    </sources>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.gmavenplus</groupId>
+        <artifactId>gmavenplus-plugin</artifactId>
+        <version>${gmaven.version}</version>
+        <executions>
+          <execution>
+            <id>compile</id>
+            <phase>process-sources</phase>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <targetBytecode>${groovy.targetBytecode}</targetBytecode>
+          <sources>
+            <source>
+              <directory>${project.build.sourceDirectory}</directory>
+              <includes>
+                <include>**/*.groovy</include>
+              </includes>
+            </source>
+          </sources>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
--- a/org.jacoco.core.test.validation.groovy/pom.xml
+++ b/org.jacoco.core.test.validation.groovy/pom.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+   All rights reserved. This program and the accompanying materials
+   are made available under the terms of the Eclipse Public License v1.0
+   which accompanies this distribution, and is available at
+   http://www.eclipse.org/legal/epl-v10.html
+
+   Contributors:
+      Andres Almiray
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jacoco</groupId>
+        <artifactId>org.jacoco.core.test.validation</artifactId>
+        <version>0.8.2-SNAPSHOT</version>
+        <relativePath>../org.jacoco.core.test.validation</relativePath>
+    </parent>
+
+    <artifactId>org.jacoco.core.test.validation.groovy</artifactId>
+
+    <name>JaCoCo :: Test :: Core :: Validation Groovy 2</name>
+
+    <properties>
+        <bytecode.version>7</bytecode.version>
+        <gmaven.version>1.6</gmaven.version>
+        <groovy.version>2.5.2</groovy.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>org.jacoco.core.test</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>${groovy.version}</version>
+            <type>pom</type>
+        </dependency>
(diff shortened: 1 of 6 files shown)


org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -61,7 +59,11 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <targetBytecode>${bytecode.version}</targetBytecode>
+                    <!--
+                    Force target bytecode to 1.7 for now
+                    as Groovy does not understand values such as 8 or 9
```
Godin: Same for Kotlin, so we used Kotlin-specific property per bytecode profile - see https://github.com/jacoco/jacoco/blob/811cf32772ce5d36947235d11d3b5af6ee3f1d8b/org.jacoco.core.test.validation/pom.xml#L121

org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/GroovyDataClassTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.jacoco.core.test.validation.ValidationTestBase;
+import org.jacoco.core.test.validation.groovy.target.GroovyDataClassTarget;
+
+/**
+ * Test of <code>data class</code>es.
+ */
+public class GroovyDataClassTest extends ValidationTestBase {
+    public GroovyDataClassTest() {
+        super(GroovyDataClassTarget.class);
```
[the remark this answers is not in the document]
aalmiray: Err nope, we were using the **wrong** package name -> s/target/targets/ 🤦‍♂️

aalmiray:
Can't merge this PR until Groovy adds `@Generated` to base methods. Perhaps in Groovy `2.5.3`.

Godin:
@aalmiray Maybe we can simply test cases where annotation already present - #610 ?
Or have an assertion with current but not expected in future count of methods.
Or use `groovy.lang.GroovySystem.version` to make assertion conditional.

aalmiray:
Follow the discussion on `@Generated` in Groovy.next at https://lists.apache.org/thread.html/49c6ee74767a9cdac419ababf9a0f0f860bf694ad49b0d632bcc3591@%3Cdev.groovy.apache.org%3E

Godin:
BTW @aalmiray if Groovy compiler always produces line numbers for non-generated code, does not produce them for generated, and there is a way to determine that class is a Groovy class, then similarly to Kotlin we can filter out all methods that don't have line number - see #552

Godin:
@aalmiray 

> Would it make sense to protect the workaround with a configuration flag so that it can be turned on and off?

We discourage addition of configuration options for many reasons. They bring significant complexity since require propagation into downstream projects (e.g. Gradle plugin) and especially into ones that embed JaCoCo as a library and perform their own analysis of `exec` file instead of reading of XML report (e.g. SonarQube, old Jenkins plugin, Eclipse EclEmma). And users don't understand this, which negatively impacts time/efforts required from us to answer questions.

> I've got word from Paul King that many additions made by the compiler set the line number to -1.

This is just part of an answer on initial question 😉 Are line numbers and other debug information (such as `SourceFile Attribute`) always produced for non-generated code as like in `kotlinc` or possible to disable as like in `javac` with `-g:none`? This is important, because in presence of later case we won't be able to rely on line numbers.

Anyway @marchof is right about type hierarchies

> JaCoCo instruments every class independent from each other and does trigger the class loader to load additional classes.

just for precision - filtering happens at analysis time, not during instrumentation. However story is pretty much the same - we analyze one file at a time and can't observe type hierarchies, even not guaranteed that we can observe full class path.

> Is there any other unique characteristic?

Wild idea - usage of `SourceFile Attribute`, but most likely fragile, because `.groovy` is not the only one possible file extension?

aalmiray:
The next version of Groovy (`2.5.3`) and by consequence `3.0.0` has added the `@Generated` annotation at more locations. Running this PR with a local version of Groovy and `-Dgroovy.version=3.0.0-SNAPSHOT` generates the following code for the simple `GroovyDataClassTarget` class

    package org.jacoco.core.test.validation.groovy.targets;

    import groovy.lang.GroovyObject;
    import groovy.lang.MetaClass;
    import groovy.transform.EqualsAndHashCode;
    import groovy.transform.Generated;
    import groovy.transform.Internal;
    import groovy.transform.ToString;
    import org.codehaus.groovy.runtime.InvokerHelper;
    import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
    import org.codehaus.groovy.runtime.callsite.CallSite;
    import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
    import org.codehaus.groovy.runtime.typehandling.ShortTypeHandling;
    import org.codehaus.groovy.util.HashCodeHelper;

    @ToString
    @EqualsAndHashCode
    public class GroovyDataClassTarget implements GroovyObject {
        private int valRead;

        @Generated
        public GroovyDataClassTarget(int valRead) {
            CallSite[] var2 = $getCallSiteArray();
            super();
            MetaClass var3 = this.$getStaticMetaClass();
            this.metaClass = var3;
            this.valRead = DefaultTypeTransformation.intUnbox(valRead);
        }

        @Generated
        public GroovyDataClassTarget() {
            CallSite[] var1 = $getCallSiteArray();
            this(Integer.valueOf(0));
        }

        public static void main(String... args) {
            CallSite[] var1 = $getCallSiteArray();
            var1[0].callConstructor(GroovyDataClassTarget.class);
        }

        @Generated
        public String toString() {
            CallSite[] var1 = $getCallSiteArray();
            Object _result = var1[1].callConstructor(StringBuilder.class);
            Object $toStringFirst = Boolean.TRUE;
            var1[2].call(_result, "org.jacoco.core.test.validation.groovy.targets.GroovyDataClassTarget(");
            if (DefaultTypeTransformation.booleanUnbox($toStringFirst)) {
                Boolean var4 = Boolean.FALSE;
            } else {
                var1[3].call(_result, ", ");
            }

            var1[4].call(_result, var1[5].callStatic(InvokerHelper.class, var1[6].callCurrent(this)));
            var1[7].call(_result, ")");
            return (String)ShortTypeHandling.castToString(var1[8].call(_result));
        }

        @Generated
        public int hashCode() {
            CallSite[] var1 = $getCallSiteArray();
            Object _result = var1[9].callStatic(HashCodeHelper.class);
            if (ScriptBytecodeAdapter.compareNotIdentical(var1[10].callCurrent(this), this)) {
                Object var3 = var1[11].callStatic(HashCodeHelper.class, _result, var1[12].callCurrent(this));
                _result = var3;
            }

            return DefaultTypeTransformation.intUnbox(_result);
        }

        @Generated
        public boolean canEqual(Object other) {
            CallSite[] var2 = $getCallSiteArray();
            return other instanceof GroovyDataClassTarget;
        }

        @Generated
        public boolean equals(Object other) {
            CallSite[] var2 = $getCallSiteArray();
            if (ScriptBytecodeAdapter.compareEqual(other, (Object)null)) {
                return false;
            } else if (DefaultTypeTransformation.booleanUnbox(var2[13].callCurrent(this, other))) {
                return true;
            } else if (!(other instanceof GroovyDataClassTarget)) {
                return false;
            } else {
                GroovyDataClassTarget otherTyped = (GroovyDataClassTarget)other;
                if (!DefaultTypeTransformation.booleanUnbox(var2[14].call(otherTyped, this))) {
                    return false;
                } else {
                    return ScriptBytecodeAdapter.compareEqual(var2[15].callCurrent(this), var2[16].call(otherTyped));
                }
            }
        }

        @Generated
        @Internal
        public MetaClass getMetaClass() {
            MetaClass var10000 = this.metaClass;
            if (this.metaClass != null) {
                return var10000;
            } else {
                this.metaClass = this.$getStaticMetaClass();
                return this.metaClass;
            }
        }

        @Generated
        @Internal
        public void setMetaClass(MetaClass var1) {
            this.metaClass = var1;
        }

        @Generated
        @Internal
        public Object invokeMethod(String var1, Object var2) {
            return this.getMetaClass().invokeMethod(this, var1, var2);
        }

        @Generated
        @Internal
        public Object getProperty(String var1) {
            return this.getMetaClass().getProperty(this, var1);
        }

        @Generated
        @Internal
        public void setProperty(String var1, Object var2) {
            this.getMetaClass().setProperty(this, var1, var2);
        }

        @Generated
        public int getValRead() {
            return this.valRead;
        }

        @Generated
        public void setValRead(int var1) {
            this.valRead = var1;
        }
    }

@marchof The test case we wrote now fails with

    [ERROR] test_method_count on test_method_count(org.jacoco.core.test.validation.groovy.GroovyDataClassTest)(org.jacoco.core.test.validation.groovy.GroovyDataClassTest)  Time elapsed: 0.372 s  <<< FAILURE!
    java.lang.AssertionError: expected:<2> but was:<1>
        at org.jacoco.core.test.validation.groovy.GroovyDataClassTest.test_method_count(GroovyDataClassTest.java:32)

It looks to me that the test expects the `main` method and the no-args constructor but as you can see, even the no-args constructor is annotated with `@Generated`. Is this OK for you?

Godin:
@aalmiray 

> Is this OK for you?

IMO this is up to you - what Groovy developers and users wanna see in their coverage reports? :wink: 

It seems that getters and setters are also now marked. FYI we don't filter out getters and setters generated by Kotlin compiler, however getters and setters generated by Lombok are filtered out - see https://github.com/jacoco/jacoco/pull/689#pullrequestreview-126601396 and https://youtrack.jetbrains.net/issue/KT-18383#focus=streamItem-27-2890903-0-0

org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>org.jacoco.core.test</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
```
Godin: `groovy-all` brings quite a lot of dependencies - `mvn -pl org.jacoco.core.test.validation.groovy dependency:tree`:

```
[INFO] \- org.codehaus.groovy:groovy-all:pom:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-ant:jar:2.5.3:compile
[INFO]    |  +- org.apache.ant:ant:jar:1.7.1:compile (version managed from 1.9.13)
[INFO]    |  |  \- org.apache.ant:ant-launcher:jar:1.9.13:compile
[INFO]    |  +- org.apache.ant:ant-junit:jar:1.7.1:runtime (version managed from 1.9.13)
[INFO]    |  \- org.apache.ant:ant-antlr:jar:1.9.13:runtime
[INFO]    +- org.codehaus.groovy:groovy-cli-commons:jar:2.5.3:compile
[INFO]    |  \- commons-cli:commons-cli:jar:1.4:compile
[INFO]    +- org.codehaus.groovy:groovy-cli-picocli:jar:2.5.3:compile
[INFO]    |  \- info.picocli:picocli:jar:3.6.0:compile
[INFO]    +- org.codehaus.groovy:groovy-console:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-datetime:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-docgenerator:jar:2.5.3:compile
[INFO]    |  \- com.thoughtworks.qdox:qdox:jar:1.12.1:compile
[INFO]    +- org.codehaus.groovy:groovy-groovydoc:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-groovysh:jar:2.5.3:compile
[INFO]    |  \- jline:jline:jar:2.14.6:compile
[INFO]    +- org.codehaus.groovy:groovy-jmx:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-json:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-jsr223:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-macro:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-nio:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-servlet:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-sql:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-swing:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-templates:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-test:jar:2.5.3:compile
[INFO]    +- org.codehaus.groovy:groovy-test-junit5:jar:2.5.3:compile
[INFO]    |  +- org.junit.platform:junit-platform-launcher:jar:1.3.1:compile
[INFO]    |  |  +- org.apiguardian:apiguardian-api:jar:1.0.0:compile
[INFO]    |  |  \- org.junit.platform:junit-platform-engine:jar:1.3.1:compile
[INFO]    |  |     +- org.junit.platform:junit-platform-commons:jar:1.3.1:compile
[INFO]    |  |     \- org.opentest4j:opentest4j:jar:1.1.1:compile
[INFO]    |  \- org.junit.jupiter:junit-jupiter-engine:jar:5.3.1:runtime
[INFO]    |     \- org.junit.jupiter:junit-jupiter-api:jar:5.3.1:runtime
[INFO]    +- org.codehaus.groovy:groovy-testng:jar:2.5.3:compile
[INFO]    |  \- org.testng:testng:jar:6.13.1:runtime
[INFO]    |     \- com.beust:jcommander:jar:1.72:runtime
[INFO]    \- org.codehaus.groovy:groovy-xml:jar:2.5.3:compile
```

`maven-surefire-plugin` sees `org.testng:testng` and uses it for execution - stacktrace of failed test:
```
at org.testng.TestNG.run(TestNG.java:996)
at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:135)
at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:112) 
```

so most likely `maven-surefire-plugin` version `2.9` declared in our parent pom doesn't work because of https://issues.apache.org/jira/browse/SUREFIRE-1019

I think that in any case just `groovy` should be used instead of `groovy-all`.

org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.0</version>
+                <configuration>
+                    <testSourceDirectory>${project.build.sourceDirectory}</testSourceDirectory>
+                    <testClassesDirectory>${project.build.outputDirectory}</testClassesDirectory>
```
Godin: `testSourceDirectory` and `testClassesDirectory` already specified in parent pom, so should be removed.

org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+                </executions>
+                <configuration>
+                    <!--
+                    Force target bytecode to 1.7 for now
+                    as Groovy does not understand values such as 8 or 9
+                    -->
+                    <targetBytecode>1.7</targetBytecode>
+                    <sources>
+                        <source>
+                            <directory>${project.basedir}/src</directory>
```
Godin: Replace `${project.basedir}/src` by `${project.build.sourceDirectory}`.

org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+                            <includes>
+                                <include>**/*.groovy</include>
+                            </includes>
+                        </source>
+                    </sources>
+                    <sourceEncoding>${project.build.sourceEncoding}</sourceEncoding>
```
Godin: Remove - `${project.build.sourceEncoding}` is default in `gmavenplus-plugin`.

org.jacoco.core.test.validation.groovy/pom.xml
aalmiray: Reason to having this block is to apply version `2.22.0` of the surefire plugin, however I've removed this block as the `groovy-all` dependency was changed to just `groovy`.

org.jacoco.core.test.validation.groovy/pom.xml
aalmiray: OK

org.jacoco.core.test.validation.groovy/pom.xml
aalmiray: OK

aalmiray:
Changes applied. Test still fails with `Instructions (GroovyDataClassTarget.groovy:25) expected:<[NOT_COVERED]> but was:<[EMPTY]>`

org.jacoco.core.test.validation.groovy/pom.xml
Godin: I believe that Maven takes configuration from parent even in case of version change.

aalmiray: Does it? I think that the parent can define `<inherited>true</inherited>` in the plugin block to allow children to override values, but I may be wrong. Anyway, old version and smaller groovy dep are now in place.

Godin:
@aalmiray Thanks a lot that you made changes! This is really appreciated! ❤️ Quite often contributors leave unfinished work and disappear.

Coming back to https://github.com/jacoco/jacoco/pull/733#issuecomment-432732309 : I don't see reasons to not change assertions in test as following

groovydataclasstarget_groovy

in order to merge this. Or I miss something?

org.jacoco.core.test.validation.groovy/pom.xml
Godin: can define `<inherited>false</inherited>` which is not the case for our parent that uses [default `<inherited>true</inherited>`](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Using_the_inherited_Tag_In_Build_Plugins) 😉 anyway, thanks!

aalmiray:
Do you mean changing the comment at the class level to `assertEmpty()`? Then yes, I can do that :smile:

Godin:
@aalmiray also line `int valRead` since Groovy 2.5.3 mark everything as generated ... and done :wink:

Godin:
@aalmiray also seems that second `doCall` method that has no parameters and line numbers and delegates to the above `doCall(java.lang.Object)` lacks `@Generated` annotation

aalmiray:
@Godin I've reached to the Groovy dev list to figure out what may be the reason for this behavior

PR #734, opened 2018-08-14
Simplify unit tests of filters

Godin:


marchof:
@Godin Not directly related to this change: I just noted that unlike the unit tests in all other packaged we don't have class comments here naming the SUT or describing the functionality of a base class. For consistency and a clean code base I would also add them here. Maybe this change is a good opportunity?

Godin:
@marchof added

marchof:
@Godin Just a side note about the verification strategy when we need to assert multiple calls to the `IFilterOutput` interface like in `FinallyTest`: In `FrameSnapshotTest` we implemented the strategy that the test case might do multiple `expectX()` calls. In the `@After` method the actual filtered items can be compared with the expected filtered items. Also I would then use sets instead of lists as the call sequence is not relevant.

Godin:
Here my main goal was to reduce number of implementations of `IFilterOutput` to simplify experiments with addition of methods into it. Unfortunately it is not class having all methods `abstract` and we can't use `default` methods :(

But also was thinking about introduction of `expectX` methods - we can do this later.

While sequence is not important, was wondering whether number of invocations should be considered or not?

marchof:
@Godin Sure, we can check for duplicate calls triggered by the filter and fail the test in this case.

PR #735, opened 2018-08-15
Add filter for bytecode that ECJ generates for String in switch

Godin:


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/AnnotationGeneratedFilterTest.java
```diff
@@ -134,4 +136,8 @@ public void merge(final AbstractInsnNode i1, final AbstractInsnNode i2) {
 		fail();
 	}
 
+	public void replace(AbstractInsnNode i, List<AbstractInsnNode> branches) {
```
Godin: To be removed by/after #734

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
```diff
@@ -371,6 +378,22 @@ public void visitEnd() {
 				nodeToInstruction.get(r).merge(i);
 			}
 		}
+
+		// Replace:
+		for (final Map.Entry<AbstractInsnNode, List<AbstractInsnNode>> r : replacements
```
Godin: Actually this can be done during counting without need to modify instructions.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -41,4 +43,6 @@
 	 */
 	void merge(AbstractInsnNode i1, AbstractInsnNode i2);
 
+	void replace(AbstractInsnNode i, List<AbstractInsnNode> branches);
```
marchof: I wander what exactly this method does... can you please add some JavaDoc?

Godin: @marchof I was about to ask your advice about how to document it based on what its implementation does :wink:

Something like

```
	/**
	 * Marks instruction that should be considered as having branches into the
	 * given instructions during computation of coverage.
	 * 
	 * @param i
	 *            instruction to be replaced
	 * @param branches
	 *            targets of branches
	 */
```

?

marchof: Ok, so it does actually not replace the instruction but its branches. In this case I would rename the method and parameters to

    replaceBranches(origin, targets)

Also maybe a second sentence can clarify the behavior: "The original target branches of the `origin` instruction are ignored."

Godin: As like before this change term "branch" is fragile - we consider that each instruction has at least one branch and will be counted as covered when at least one branch is covered:

https://github.com/jacoco/jacoco/blob/d0e461a331989925776199978d22d8db82ceba03/org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java#L381-L388

So in some sense this replaces instruction and not just its branches:

https://github.com/jacoco/jacoco/blob/cc432204049c41720c8d8b2b64610fccc43d0380/org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java#L389-L411

and while we don't have use-case and don't need this, even

```
output.replace(original, Collections.singletonList(replacement))
```

works.

Thus not sure about renaming of method.
Renaming of second parameter into `targets` sounds good.
And for first one feel that `original` sounds better than `origin`.

----

I like the idea to add second statement :thumbsup: with slight rewording

> The original target branches of the **original** instruction **will be** ignored.

marchof: Well, at least in the Analyzer I used the term "branch" as a synonym for a edge in the control flow graph. But with the JavaDoc we can leave the method as is.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/StringSwitchEcjFilterTest.java
```diff
@@ -0,0 +1,134 @@
...
+		// if equal "\0a", then goto its case
+		m.visitJumpInsn(Opcodes.IFNE, case2);
+
+		// goto default case
+		m.visitJumpInsn(Opcodes.GOTO, caseDefault);
+
+		m.visitLabel(h2);
+
+		m.visitVarInsn(Opcodes.ALOAD, 2);
+		m.visitLdcInsn("\0a");
```
Godin: Copy-paste mistake - here and below should be `"b"` instead of `"\0a"`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
Godin: @marchof I added
* Javadoc to `IFilterOutput`
* tests to `MethodAnalyzerTest`
* changelog entry

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -41,4 +43,16 @@
...
+	 * Marks instruction that should be considered as having branches into the
+	 * given instructions during computation of coverage. The original target
+	 * branches of the original instruction will be ignored.
+	 *
+	 * @param original
+	 *            instruction to be replaced
+	 * @param targets
+	 *            new targets of branches
+	 */
+	void replace(AbstractInsnNode original, List<AbstractInsnNode> targets);
```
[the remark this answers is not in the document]
Godin: > The targets have no order and must be unique. Should be a Set.

Actually there is order, which is IMO well defined and very important property of/in `MethodAnalyzer`, otherwise how `merge` works? 😉 In particular `default` of `switch` has number `0`:
https://github.com/jacoco/jacoco/blob/63d55acf9071446e9ea6706d6a65c94d87faca17/org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java#L334-L335

And so maybe `List` will be less confusing and more consistent even if doesn't play role in this API ?

----

> The instruction is not replaced, only the CFG edges to its successors is replaced by new edges.

As we discussed earlier and as in implementation - we don't modify CFG, this affects only afterwards computation.

> I know I introduced the term "branch" again.

No problem with term "branch" - it is already in Javadoc.

Right now I realized that with or without renaming of method on `replaceBranches` from this interface alone not clear that replacement of branches changes coverage of instruction and not just branch coverage.

----

> ```
>     /**
>      * Replaces the outgoing branches of a given instruction with new target
> ```

Other methods use term `mark` (consistency?) and state that they affect stage of computations, i.e. not the original CFG and not the propagation of probes, what again as was discussed earlier is quite important in this case. So not sure that this is cleaner/better than current

> ```
> 	/**
> 	 * Marks instruction that should be considered as having branches into the
> 	 * given instructions during computation of coverage.
> ```

Can imagine combination of both:

```java
 	/**
	 * Marks instruction whose outgoing branches should be replaced during
	 * computation of coverage.
 	 * 
 	 * @param source
	 *            instruction which branches should be replaced
 	 * @param newTargets
	 *            new targets of branches
 	 */
	void replaceBranches(AbstractInsnNode source,
```

----

In either case to me seems that important things are missing in this documentation:

* `merge` is performed before `ignore` and `replace`
* `ignore` doesn't affect `merge` and targets in `replace`, but affects source in `replace`
* targets in `replace` are affected by `merge`s
* `replace` overrides `merge`s of `source`

PR #738, opened 2018-08-15
Add experimental support for Java 12 class files

Godin:
As per https://github.com/jacoco/jacoco/pull/725#issue-207161199 hopefully this will give us quite a good amount of time without complaints about usage with JDK 12 **Early Access** Builds.


org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -24,6 +24,8 @@ <h3>New Features</h3>
 <ul>
   <li>Experimental support for Java 11 class files
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/719">#719</a>).</li>
+  <li>Experimental support for Java 12 class files
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/???">#???</a>).</li>
```
marchof: PR number missing.

Godin: Oops, sorry. Fixed.

PR #740, opened 2018-08-17
Use ASM also for Java 10

marchof:
While we have support for Java 11 and 12 in the meantime, we still [downgrade ](https://github.com/jacoco/jacoco/blob/master/org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java) Java 10 class files.

From my point of view we should now be able to process Java 10 class files directly with ASM.


org.jacoco.core.test/src/org/jacoco/core/instr/InstrumenterTest.java
```diff
@@ -71,51 +66,20 @@ public String toString() {
...
-		final byte[] bytes = new byte[originalBytes.length];
-		System.arraycopy(originalBytes, 0, bytes, 0, originalBytes.length);
-		final long expectedClassId = CRC64.classId(bytes);
-
-		final byte[] instrumentedBytes = instrumenter.instrument(bytes, "");
-
-		assertArrayEquals(originalBytes, bytes);
-		assertEquals(BytecodeVersion.V10,
-				BytecodeVersion.get(instrumentedBytes));
-		assertEquals(expectedClassId, accessorGenerator.classId);
```
Godin: This removes test that was preventing disappearance of

https://github.com/jacoco/jacoco/blob/68ab19f3463bf5076485185fe2abe5ec8a726679/org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java#L105

i.e. test that was preventing mistake that we did with Java 9 - https://github.com/jacoco/jacoco/pull/636

org.jacoco.core.test/src/org/jacoco/core/analysis/AnalyzerTest.java
```diff
@@ -89,23 +89,6 @@ public void should_ignore_synthetic_classes() throws Exception {
...
 	}
 
-	@Test
-	public void should_analyze_java10_class() throws Exception {
-		final byte[] bytes = createClass(BytecodeVersion.V10);
-		final long expectedClassId = CRC64.classId(bytes);
-
-		analyzer.analyzeClass(bytes, "");
-
-		assertEquals(expectedClassId, classes.get("Foo").getId());
```
Godin: This removes test that was checking that we use original version for `classId`

https://github.com/jacoco/jacoco/blob/68ab19f3463bf5076485185fe2abe5ec8a726679/org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java#L115

org.jacoco.core.test/src/org/jacoco/core/instr/InstrumenterTest.java
Godin: @marchof and this reminded me about https://gitlab.ow2.org/asm/asm/merge_requests/143#note_34290 :

> ```java
> public class MyClassReader extends ClassReader {
> 
>   public MyClassReader(byte[] classFile, int version) {
>     super(setVersion(classFile, version));
>     // Restore the original version.
>     for (int i = 0; i < 4; ++i) classFile[i + 4] = classFile[i];
>     // Optional: restore the 0xCAFEBABE magic header.
>   }
> 
>   private static byte[] setVersion(byte[] classFile, int version) {
>     // Copy the original version over the 0xCAFEBABE magic header.
>     for (int i = 0; i < 4; ++i) classFile[i] = classFile[i + 4];
>     // Change the version to the given one.
>     classFile[4] = (byte) (version >>> 24);
>     classFile[5] = (byte) (version >>> 16);
>     classFile[6] = (byte) (version >>> 8);
>     classFile[7] = (byte) version;
>     return classFile;
>   }
> }
> ```

what might be safer/easier to use than our `BytecodeVersion.downgradeIfNeeded` followed by `BytecodeVersion.set`.

marchof:
@Godin You're right about the tests. As I removed the downgrade logic for `V10` these tests can  be applied any more. I pushed another proposal: There is `BytecodeVersion.MAX_VERSION` which points to the latest version supported by ASM (currently `V12`).

- 👍 Tests are now generic (`MAX_VERSION + 1) and we only need to adjust `MAX_VERSION` in future.
- 👎 We automatically downgrade any future version (might fail later in ASM or when classes are loaded). This might also be actually be seen as feature because in many cases it would have worked in the past.

WDYT?

marchof:
@Godin Rebased. WDYT?

org.jacoco.core/src/org/jacoco/core/internal/BytecodeVersion.java
```diff
@@ -56,25 +56,25 @@ public static void set(final byte[] b, final int version) {
...
 			final byte[] source) {
-		if (V10 != version) {
+		// consider major version only (due to 1.1 anomaly)
+		if ((version & 0xff) <= MAX_VERSION) {
 			return source;
 		}
 		final byte[] b = new byte[source.length];
 		System.arraycopy(source, 0, b, 0, source.length);
-		set(b, Opcodes.V9);
+		set(b, MAX_VERSION);
```
Godin: This trick not needed anymore given that ASM now supports and will continue supporting release+1.

org.jacoco.core.test/src/org/jacoco/core/instr/InstrumenterTest.java
```diff
@@ -97,17 +97,18 @@ public void setup() throws Exception {
 	}
 
 	@Test
-	public void should_instrument_java10_class() throws Exception {
-		final byte[] originalBytes = createClass(BytecodeVersion.V10);
+	public void should_instrument_class_greater_than_max_version()
```
Godin: While "downgrade" trick can be removed, this test must be preserved under another name to prevent regressions, because it verifies that `Instrumenter` computes `classId` using original class bytes.

org.jacoco.core.test/src/org/jacoco/core/analysis/AnalyzerTest.java
```diff
@@ -90,8 +91,10 @@ public void should_ignore_synthetic_classes() throws Exception {
 	}
 
 	@Test
-	public void should_analyze_java10_class() throws Exception {
-		final byte[] bytes = createClass(BytecodeVersion.V10);
+	public void should_analyze_class_greater_than_max_version()
```
Godin: While "downgrade" trick can be removed, this test must be preserved under another name to prevent regressions, because it verifies that `Analyzer` computes `classId` using original class bytes.

org.jacoco.core.test/src/org/jacoco/core/runtime/ModifiedSystemClassRuntimeTest.java
```diff
@@ -82,8 +83,9 @@ public void should_instrument_java10_class() {
 	 * "java.lang.reflect.Module" introduced in JDK 9.
 	 */
 	private Instrumentation newInstrumentationMock() {
-		return (Instrumentation) Proxy.newProxyInstance(getClass()
-				.getClassLoader(), new Class[] { Instrumentation.class },
+		return (Instrumentation) Proxy.newProxyInstance(
```
Godin: Minor remark: it took me some time during review to divide this and similar non-functional style changes from real ones. Wondering if we can improve this somehow later. As a note - usually I try to not reformat otherwise non-changed existing lines.

marchof:
Yes, if you have the time, please move forward! Thx

PR #741, opened 2018-08-18
StringSwitchEcjFilter fails to recognize switch where default is first 

Godin:
```
	private static void example(Object s) {
		switch (String.valueOf(s)) { // assertFullyCovered(0, 2)
			default:
				nop("default");
				break;
			case "a":
				nop("case a");
				break;
		}
	}
```

in this case `ECJ` omits last `goto` expected by filter - https://github.com/jacoco/jacoco/blob/68ab19f3463bf5076485185fe2abe5ec8a726679/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchEcjFilter.java#L83-L86


Godin:
@marchof found this during check that #735 covers all cases from #496 , which mentions this case in comments :wink: don't remember how/why comment was originally added into #496 , however can admit that in C/C++ `default` as first case with early-return seems quite handy to me in some cases, especially in big switches

PR #744, opened 2018-08-20
Refactor Analysis Package

marchof:
### General Objectives

* Separate bytecode CGF analysis from filtering
* Create smaller testable units
* Potentially simplify data structures and memory throughput

### Open Tasks

- [x] Improve API, unit tests and JavaDoc of `Instruction`
- [x] Check whether merge can be implemented to not modify `Instruction`
- [x] Unit tests for `InstructionsBuilder`
- [x] Unit tests for `MethodCoverageCalculator`
- [x] Move `Instruction` to analysis package (remove dependency from `LabelInfo`)

Maybe separate PR:

- Rework `MethodCoverageCalculator` in a way that `IFIlterOutput` methods directly work on `Instruction` map to avoid intermediate lists (optional).


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -884,14 +884,22 @@ private void runMethodAnalzer() {
 
 	private void runMethodAnalzer(IFilter filter) {
 		LabelFlowAnalyzer.markLabels(method);
-		final MethodAnalyzer analyzer = new MethodAnalyzer("doit", "()V", null,
-				probes, filter, new FilterContextMock());
+		InstructionsBuilder icc = new InstructionsBuilder(
```
marchof: Rename `icc` to `builder`

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodCoverageCalculatorTest.java
```diff
@@ -0,0 +1,207 @@
...
+	public void should_ignore_single_instruction() {
+		addInsn(1, true);
+		InsnNode i1 = addInsn(1, false);
+		addInsn(2, true);
+
+		MethodCoverageCalculator c = new MethodCoverageCalculator(instructions);
+		c.ignore(i1, i1);
+		c.calculate(coverage);
+
+		assertLine(1, 0, 1, 0, 0); // only one instruction nit filtered
```
marchof: nit --> not

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -80,21 +84,39 @@ public MethodProbesVisitor visitMethod(final int access, final String name,
 
 		InstrSupport.assertNotInstrumented(name, coverage.getName());
 
-		return new MethodAnalyzer(stringPool.get(name), stringPool.get(desc),
-				stringPool.get(signature), probes, Filters.ALL, this) {
+		final InstructionsBuilder icc = new InstructionsBuilder(
```
marchof: Rename `icc` --> `builder`

org.jacoco.core/src/org/jacoco/core/internal/analysis/Instruction.java
```diff
@@ -0,0 +1,207 @@
...
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis;
+
+import java.util.BitSet;
+import java.util.Collection;
+
+import org.jacoco.core.analysis.ICounter;
+
+/**
+ * Execution status of a single byte code instruction internally used for
```
marchof: "byte code" --> "bytecode"

org.jacoco.core/src/org/jacoco/core/internal/analysis/Instruction.java
```diff
@@ -0,0 +1,207 @@
...
+ * outgoing branch. Each instructions has at least one branch, for example in
+ * case of a simple sequence of instructions (by convention branch 0). Instances
+ * of this class are used in two steps:
+ * 
+ * <h3>Step 1: Building the CFG</h3>
+ * 
+ * For each bytecode instruction of a method a {@link Instruction} instance is
+ * created. In correspondence with the CFG these instances are linked with each
+ * other with the <code>addBranch()</code> methods. The executions status is
+ * either be directly derived from a probe which has been inserted in the
```
marchof: Remove "be".

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+
+import org.jacoco.core.analysis.ISourceNode;
+import org.jacoco.core.internal.flow.LabelInfo;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.tree.AbstractInsnNode;
+
+/**
+ * Stateful builder for the {@link Instruction}s of a method. All instructions
+ * of a method must be added in their original sequence along with additional
+ * information like line numbers. Afterwards the the instructions can be
```
marchof: Remove "the".

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+	private final List<Label> currentLabel;
+
+	/**
+	 * List of all jumps within the control flow. We need to store jumps
+	 * temporarily as the target {@link Instruction} may not been known yet.
+	 */
+	private final List<Jump> jumps;
+
+	/**
+	 * Creates a new build instance which can be used to analyze a single
```
marchof: Typo: "build" --> "builder"

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+	 * assigned to this line. If no line is set (e.g. for classes compiled
+	 * without debug information) {@link ISourceNode#UNKNOWN_LINE} is assigned
+	 * to the instructions.
+	 */
+	void setCurrentLine(final int line) {
+		currentLine = line;
+	}
+
+	/**
+	 * Adds a label which applies to the subsequent added instruction. Due to
```
marchof: "subsequent" --> "subsequently"

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+	 */
+	void addLabel(final Label label) {
+		currentLabel.add(label);
+		if (!LabelInfo.isSuccessor(label)) {
+			noSuccessor();
+		}
+	}
+
+	/**
+	 * Add a new instruction. Instructions are by default linked with the
```
marchof: "Add" --> "Adds"

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+		}
+		if (currentInsn != null) {
+			currentInsn.addBranch(insn, 0);
+		}
+		currentInsn = insn;
+		instructions.put(node, insn);
+	}
+
+	/**
+	 * Declares that the next instruction will not be an successor of the
```
marchof: Typo: "an" --> "a"

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+		currentInsn = null;
+	}
+
+	/**
+	 * Adds a jump from the last added instruction.
+	 * 
+	 * @param target
+	 *            jump target
+	 * @param branch
+	 *            unique branch number for the last instruction
```
marchof: Remove "last instruction", already mentioned in the general description

org.jacoco.core/src/org/jacoco/core/internal/analysis/InstructionsBuilder.java
```diff
@@ -0,0 +1,186 @@
...
+	void addProbe(final int probeId, final int branch) {
+		final boolean executed = probes != null && probes[probeId];
+		currentInsn.addBranch(executed, branch);
+	}
+
+	/**
+	 * Returns the status for all instructions of this method. This method must
+	 * be called exactly once after the instructions have been added.
+	 * 
+	 * @return map of ASM instruction node to corresponding {@link Instruction}
```
marchof: Type: "node" --> "nodes"

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -0,0 +1,159 @@
...
+				.entrySet()) {
+			if (ignored.contains(i.getKey())) {
+				continue;
+			}
+
+			Instruction insn = i.getValue();
+
+			final Set<AbstractInsnNode> r = replacements.get(i.getKey());
+			if (r != null) {
+				final List<Instruction> newBranches = new ArrayList<Instruction>();
```
marchof: Pre-size with `r.size()`

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -0,0 +1,159 @@
...
+
+		for (final Map.Entry<AbstractInsnNode, Instruction> entry : instructions
+				.entrySet()) {
+
+			final Instruction i = entry.getValue();
+			final AbstractInsnNode n = entry.getKey();
+			final AbstractInsnNode r = findRepresentative(n);
+			if (r != n) {
+				ignored.add(n);
+				instructions.put(r, instructions.get(r).merge(i));
```
Godin: To me seems that something is wrong here.

Before this change we were performing merges in two steps - see #747
1. from all instructions into their representative instruction
2. from all representative instructions back

and seems that this is gone.

Consider following case for `MethodCoverageCalculatorTest`, whose assertion currently will fail

```java
	@Test
	public void test() {
		InsnNode i1 = addInsn(1, false);
		InsnNode i2 = addInsn(2, false);
		InsnNode i3 = addInsn(2, true);

		MethodCoverageCalculator c = new MethodCoverageCalculator(instructions);
		c.merge(i3, i2);
		c.replaceBranches(i1, Collections.<AbstractInsnNode> singleton(i2));
		c.calculate(coverage);

		assertLine(1, 0, 1, 0, 0);
	}
```

where

1. initially `i1` and `i2` was not covered, `i3` was covered
2. merge `i3` and `i2` - both should be considered as covered since `i3` was covered
3. replace `i1` by `i2` - `i1` should be considered as covered because of previous merge

This test fails even if `c.merge(i3, i2)` replaced by

```
c.merge(i3, i2);
c.merge(i2, i3);
```

and passes when replaced by

```
c.merge(i2, i3);
```

or by

```
c.merge(i2, i3);
c.merge(i3, i2);
```

marchof: @Godin Excellent catch! I added your test case and applied a slightly different fix: Instead of iterating over all instructions every time I think we can pick the representative instruction when replacing branches.

Godin: @marchof 

> pick the representative instruction when replacing branches

IMO this might become source of errors in future. What do you think about replacement of

```
instructions.put(r, instructions.get(r).merge(i));
```

by

```
instruction = instructions.get(representativeNode).merge(instruction);
instructions.put(r, instruction);
instructions.put(n, instruction);
```

instead?

marchof: Looks good to me at the first glance and all tests are green (if I remove the lookup during handling replacements).

But: Not sure whether this is correct in case we merge more than two instructions. If we merge pairwise with the representative will the total merge result be propagated to all instructions of the group? Trying to create a test case now...

Godin: @marchof representative is the same

Godin: @marchof this will work if we mutate instructions, but now they are immutable so probably you right and this won't work :(

marchof: Here is a demonstration of the issue:

	@Test
	public void should_replace_branches_with_merged_instructions() {
		InsnNode i1 = addInsn(1, false, false, false);
		InsnNode i2 = addInsn(2, true);
		InsnNode i3 = addInsn(2, false);
		InsnNode i4 = addInsn(2, false);

		MethodCoverageCalculator c = new MethodCoverageCalculator(instructions);
		c.merge(i4, i3);
		c.merge(i3, i2);
		c.replaceBranches(i1,
				new HashSet<AbstractInsnNode>(Arrays.asList(i2, i3, i4)));
		c.calculate(coverage);

		assertLine(1, 0, 1, 0, 3);
	}

Works with current version, but breaks without representative lookup during replacement.

PR #745, opened 2018-08-20
Update and rename RT.java to RTT.java

bm-sushma:


bm-sushma:
Sorry, I did it by mistake

PR #760, opened 2018-10-01
Upgrade ASM to 7.0 for official support of Java 11

Godin:


Godin:

PR #761, opened 2018-10-03
Add filter for Kotlin "unsafe" cast operator

### LINKED ISSUE #754: Weird coverage behavior with Kotlin .trim()
### Steps to reproduce

JaCoCo version: 0.8.2
Operating system: macOS High Sierra 10.13.6
Tool integration: Gradle

Run JaCoCo using gradle on this Kotlin code:

```kotlin
    fun returnString(): String {
        return "string"
    }

    fun coverage1() {
        returnString().trim()
    }

    fun coverage2() {
        returnString()
    }
```

Here's the code for Kotlin's `trim()` method:
```kotlin
/**
 * Returns a string having leading and trailing whitespace removed.
 */
@kotlin.internal.InlineOnly
public inline fun String.trim(): String = (this as CharSequence).trim().toString()

```

### Expected behaviour
I expect the line to be fully covered when using Kotlin `.trim()`

### Actual behaviour

screen shot 2018-09-07 at 2 34 52 pm

Godin:
Fixes #754

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilter.java
@@ -1,0 +1,71 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import org.objectweb.asm.Opcodes;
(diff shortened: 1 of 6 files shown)


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilter.java
```diff
@@ -0,0 +1,74 @@
...
+			cursor = start;
+
+			nextIsNew(KOTLIN_TYPE_CAST_EXCEPTION);
+			nextIs(Opcodes.DUP);
+			nextIs(Opcodes.LDC);
+			if (cursor == null) {
+				return;
+			}
+			final LdcInsnNode ldc = (LdcInsnNode) cursor;
+			if (ldc.cst.getClass() != String.class) {
```
marchof: Why not simply using `instanceof` here?

    if (!(ldc.cst instanceof String)) {

Godin: @marchof just because negation of `instanceof` requires parentheses and so looks a bit ugly 😄

marchof: 👍 Good idea to combine both conditions.

PR #764, opened 2018-10-09
Add filter for instructions inlined by Kotlin compiler

### LINKED ISSUE #763: Add filter for instructions inlined by Kotlin compiler
Consider following files

`a.kt`
```
inline fun a() {
  nop()
}
```

`b.kt`
```
inline fun b() {
  nop()
}
```

and `callsite.kt` with 8 lines
```
fun main(args : Array<String>) {
  a() // line 2
  b() // line 3
} // line 4

fun nop() {
} // line 7
```

Kotlin compiler produces `CallsiteKt.class` whose bytecode

```
Constant pool:
...
   #7 = Utf8               Lorg/jetbrains/annotations/NotNull;
...
  #24 = Utf8               Lkotlin/Metadata;
...
{
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
         0: aload_0
         1: ldc           #9                  // String args
         3: invokestatic  #15                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
         6: nop
         7: invokestatic  #19                 // Method nop:()V
        10: nop
        11: nop
        12: invokestatic  #19                 // Method nop:()V
        15: nop
        16: return
      LocalVariableTable:
            7       4     1 $i$f$a   I
           12       4     1 $i$f$b   I
            0      17     0  args   [Ljava/lang/String;
      LineNumberTable:
        line 2: 6
        line 9: 7
        line 10: 10
        line 3: 11
        line 11: 12
        line 12: 15
        line 4: 16
    RuntimeInvisibleParameterAnnotations:
      0:
        0: #7()

  public static final void nop();
    descriptor: ()V
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=0, locals=0, args_size=0
         0: return
      LineNumberTable:
        line 7: 0
}
SourceFile: "callsite.kt"
SourceDebugExtension:
  SMAP
  callsite.kt
  Kotlin
  *S Kotlin
  *F
  + 1 callsite.kt
  CallsiteKt
  + 2 a.kt
  AKt
  + 3 b.kt
  BKt
  *L
  1#1,8:1
  2#2,2:9
  2#3,2:11
  *E
  *S KotlinDebug
  *F
  + 1 callsite.kt
  CallsiteKt
  *L
  2#1,2:9
  3#1,2:11
  *E
RuntimeVisibleAnnotations:
  0: #24(#25=[I#26,I#26,I#27],#28=[I#26,I#29,I#30],#31=I#30,#32=[s#33],#34=[s#5,s#35,s#8,s#35,s#35,s#6,s#16])
```

Contains inlined instructions with line numbers that are greater than actually present in source file:

```
$ java -jar jacoco-0.8.2/lib/jacococli.jar classinfo CallsiteKt.class
    11      0      8      2      2   class 0xcc96f697bc87baa7 CallsiteKt
    10      0      7      1      1   +- method main([Ljava/lang/String;)V
     1      0                        |  +- line 2
     1      0                        |  +- line 3
     1      0                        |  +- line 4
     1      0                        |  +- line 9
     1      0                        |  +- line 10
     1      0                        |  +- line 11
     1      0                        |  +- line 12
     1      0      1      1      1   +- method nop()V
     1      0                           +- line 7
```

Since we use source line numbers for generation of page with highlighted source in our HTML report, additional lines (9-12) do not appear on it and do not cause troubles. However they add to values of counters as shown by `classinfo` above and what is also visible in HTML report:

methods

sources

Not demonstrated by above example, but inlining can also affect "branches" and "complexity" counters.

Moreover these additional lines appear in XML report:

```
        <sourcefile name="callsite.kt">
            <line nr="2" mi="2" ci="0" mb="0" cb="0"/>
            <line nr="3" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="4" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="7" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="9" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="10" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="12" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="13" mi="1" ci="0" mb="0" cb="0"/>
            <line nr="14" mi="1" ci="0" mb="0" cb="0"/>
            <counter type="INSTRUCTION" missed="13" covered="0"/>
            <counter type="LINE" missed="9" covered="0"/>
            <counter type="COMPLEXITY" missed="2" covered="0"/>
            <counter type="METHOD" missed="2" covered="0"/>
            <counter type="CLASS" missed="1" covered="0"/>
        </sourcefile>
```

Which is not good for tools that read our XML report and try to enforce its consistency with source code by checking that line numbers are not greater than in source code.

While ideally we need to somehow map these additional instructions back to their original location in order to show coverage of original (#654).

I propose to do a baby step and at least ignore them at call sites. Seems that for that is enough to obtain minimal value of `OutputStartLine` in `LineSection` that comes after `KotlinDebug` `StratumSection` of `SMAP` recorded in class file - minimum of 9 and 11 in the above example:

```
  *S KotlinDebug
  *F
  + 1 callsite.kt
  CallsiteKt
  *L
  2#1,2:9
  3#1,2:11
  *E
```

Godin:
Fixes #763 

After this change for the same example as in #763 :
```
$ java -jar jacococli.jar classinfo --verbose CallsiteKt.class
     7      0      4      2      2   class 0xcc96f697bc87baa7 CallsiteKt
     6      0      3      1      1   +- method main([Ljava/lang/String;)V
     1      0                        |  +- line 2
     1      0                        |  +- line 3
     1      0                        |  +- line 4
     1      0      1      1      1   +- method nop()V
     1      0                           +- line 7
```


Godin:
@marchof could you please do first review pass?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,116 @@
...
+			final IFilterContext context, final IFilterOutput output) {
+		if (context.getSourceDebugExtension() == null) {
+			return;
+		}
+		if (!context.getClassAnnotations()
+				.contains(KotlinGeneratedFilter.KOTLIN_METADATA_DESC)) {
+			return;
+		}
+
+		final int max = parseSMAP(context.getSourceDebugExtension());
```
Godin: Currently filters are stateless and do processing of one method after another, so the same SMAP will be parsed multiple times - need to think how to avoid this without introduction of coupling with Kotlin's SMAP outside of filter. Either allow filter to somehow store some information outside? Or simply creation of instances of filters for each analyzed class, so that they can keep per-class state?

marchof: Another option would be to return an `SourceDebugExtension` object from `IFilterContext.getSourceDebugExtension()'. Such an instance would do the String parsing  lazily and only once.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,116 @@
...
+			readLine(br, "SMAP");
+			// OutputFileName
+			br.readLine();
+			// DefaultStratumId
+			readLine(br, "Kotlin");
+			// StratumSection
+			readLine(br, "*S Kotlin");
+			// FileSection
+			readLine(br, "*F");
+			String line;
```
marchof: This variable is not used and gives a warning.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,116 @@
...
+			if (AbstractInsnNode.LINE == i.getType()) {
+				line = ((LineNumberNode) i).line;
+			}
+			if (line >= max) {
+				output.ignore(i, i);
+			}
+		}
+	}
+
+	private static int parseSMAP(final String smap) {
```
marchof: Took me some time to understand the exclusion logic. Maybe renaming this method to `getFirstIncludeLineNumber()` or so would help to understand how the references to external files are implemented. Also a link to Kotlin documentation or a short description of the logic in the filter's JavaDoc might be helpful for future reference.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: This is what I called 

> coupling with Kotlin's SMAP outside of filter

unless the parsing is generalized from Kotlin's SMAP to just SMAP, but then not even necessarily that `SourceDebugExtension` attribute contains data in JSR-45 format. JVM Spec §4.7.11 says

> The debug_extension array holds extended debugging information which has no semantic effect on the Java Virtual Machine. The information is represented using a modified UTF-8 string with no terminating zero byte.

And IMO that's why in ASM it is just a `String` and ASM doesn't parse it.

But maybe something like `tryParseAsSMAP` won't damage API clearness.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: So far I haven't found any spec that at least states that Kotlin uses SMAP, not even talking about additional `KotlinDebug` strata. And maybe there is none at all. Understanding of how to parse comes from JSR-45 and how to interpret from `kotlinc`. But anyway you right - will work on addition of description.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: Leftover of refactorings... fixed!

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: Renamed into `getFirstGeneratedLineNumber`

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: As IMO a good first solution made filters stateful.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodAnalyzer.java
```diff
@@ -388,6 +381,16 @@ public void visitEnd() {
 			if (r != m) {
 				i.merge(nodeToInstruction.get(r));
 			}
+
+			if (!ignored.contains(m)) {
+				final int line = i.getLine();
+				if (firstLine > line || lastLine == ISourceNode.UNKNOWN_LINE) {
+					firstLine = line;
```
Godin: @marchof IMO this is good change on its own - can extract into separate PR if desired.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -104,6 +105,11 @@ private void analyze(final Analyzer analyzer, final ExecutionData data)
 	 */
 	@Test
 	public void execute_assertions_in_comments() throws IOException {
+		assertTrue(String.format(
```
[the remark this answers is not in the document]
Godin: done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,115 @@
...
+			// LineSection
+			int min = Integer.MAX_VALUE;
+			String line;
+			while (!"*E".equals(line = br.readLine())) {
+				final Matcher m = LINE_INFO_PATTERN.matcher(line);
+				if (!m.matches()) {
+					throw new AssertionError();
+				}
+				final int outputStartLine = Integer
+						.parseInt(m.group(4).substring(1));
```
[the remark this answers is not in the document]
Godin: done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,115 @@
...
+				min = Math.min(outputStartLine, min);
+			}
+			return min;
+		} catch (IOException e) {
+			throw new AssertionError(e);
+		}
+	}
+
+	private static void readLine(final BufferedReader br, final String expected)
```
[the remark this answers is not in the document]
Godin: renamed

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,115 @@
...
+				if (!m.matches()) {
+					throw new AssertionError();
+				}
+				final int outputStartLine = Integer
+						.parseInt(m.group(4).substring(1));
+				min = Math.min(outputStartLine, min);
+			}
+			return min;
+		} catch (IOException e) {
+			throw new AssertionError(e);
```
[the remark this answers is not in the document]
Godin: done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+
+	private static void readLine(final BufferedReader br, final String expected)
+			throws IOException {
+		if (!expected.equals(br.readLine())) {
+			throw new AssertionError();
```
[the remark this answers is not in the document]
Godin: From the stacktrace we know which line is expected, so replaced by

```
throw new IllegalStateException("Unexpected line: " + line);
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: Now method `Filters.all` creates new instances of all filters, however there is another option - keep all existing filters except new one as `static`. Something like

```
/**
 * Filter that combines other filters.
 */
public final class Filters implements IFilter {

	/**
	 * Filter that does nothing.
	 */
	public static final IFilter NONE = new Filters();

	/**
	 * Stateless filters.
	 */
	private static final IFilter STATELESS = new Filters(new EnumFilter(),
			new SyntheticFilter(), new SynchronizedFilter(),
			new TryWithResourcesJavac11Filter(),
			new TryWithResourcesJavacFilter(), new TryWithResourcesEcjFilter(),
			new FinallyFilter(), new PrivateEmptyNoArgConstructorFilter(),
			new StringSwitchJavacFilter(), new StringSwitchEcjFilter(),
			new EnumEmptyConstructorFilter(), new AnnotationGeneratedFilter(),
			new KotlinGeneratedFilter(), new KotlinLateinitFilter(),
			new KotlinWhenFilter(), new KotlinWhenStringFilter(),
			new KotlinUnsafeCastOperatorFilter(),
			new KotlinDefaultArgumentsFilter());

	private final IFilter[] filters;

	/**
	 * Creates filter that combines all other filters.
	 * 
	 * @return filter that combines all other filters
	 */
	public static IFilter all() {
		return new Filters(STATELESS, new KotlinInlineFilter());
	}
```

@marchof WDYT?

Godin:
@marchof I addressed all your previous comments and added entry to changelog, so not counting little idea to keep all existing filters except new one as static (see https://github.com/jacoco/jacoco/pull/764#discussion_r237972705), this PR is IMO ready to merged.

Don't forget to squash and clean description if you decide to merge 😉

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
```diff
@@ -0,0 +1,122 @@
...
+				final Matcher m = LINE_INFO_PATTERN.matcher(line);
+				if (!m.matches()) {
+					throw new IllegalStateException(
+							"Unexpected SMAP line: " + line);
+				}
+				final int outputStartLine = Integer.parseInt(m.group(4));
+				min = Math.min(outputStartLine, min);
+			}
+			return min;
+		} catch (IOException e) {
```
marchof: Missing `final` (gets inserted on save).

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
marchof: @Godin Creating all filters per class isn't a big deal. Let's keep it simple for now.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
marchof: Just fixes.

jwalter:
When I try this out locally I get this error:

`Caused by: java.lang.IllegalStateException: Unexpected SMAP line: null
        at org.jacoco.core.internal.analysis.filter.KotlinInlineFilter.expectLine(KotlinInlineFilter.java:114)
        at org.jacoco.core.internal.analysis.filter.KotlinInlineFilter.getFirstGeneratedLineNumber(KotlinInlineFilter.java:75)
        at org.jacoco.core.internal.analysis.filter.KotlinInlineFilter.filter(KotlinInlineFilter.java:42)
        at org.jacoco.core.internal.analysis.filter.Filters.filter(Filters.java:55)
        at org.jacoco.core.internal.analysis.ClassAnalyzer.addMethodCoverage(ClassAnalyzer.java:110)
        at org.jacoco.core.internal.analysis.ClassAnalyzer.access$100(ClassAnalyzer.java:31)
        at org.jacoco.core.internal.analysis.ClassAnalyzer$1.accept(ClassAnalyzer.java:99)
        at org.jacoco.core.internal.flow.ClassProbesAdapter$2.visitEnd(ClassProbesAdapter.java:89)
        at org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1279)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:679)
        at org.objectweb.asm.ClassReader.accept(ClassReader.java:391)
        at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:121)
        at org.jacoco.core.analysis.Analyzer.analyzeClass(Analyzer.java:137)
`

I _think_ this is how the SMAP looks when this happens for me:

`SMAP
GetFormTemplateBodyBuilder.kt
Kotlin
*S Kotlin
*F
+ 1 GetFormTemplateBodyBuilder.kt
se/cambio/openservices/mediatorendpoints/forms/GetFormTemplateBodyBuilder
*L
1#1,20:1
*E`

Godin:
@jwalter thanks a lot for testing and this valuable feedback just in time to keep our master free from bugs :thumbsup: our small team really appreciates this ❤️ 

Don't know yet how to reproduce this, but apparently `KotlinDebug` section is not always present in `SMAP`. Will try to do a deeper look and fix later today.

jwalter:
Not sure it helps, but the class triggering looks like this (lines 8-20. 1-7 are package declaration and imports):

```
internal class GetFormTemplateBodyBuilder: BodyBuilder<GetFormTemplateType> {
  override fun buildElement(mediatorRequest: MediatorClientRequest): JAXBElement<GetFormTemplateType> {
    val getFormsRequest = ObjectFactory().createGetFormTemplateType().apply {
      subjectOfCare = mediatorRequest.subjectOfCare()
      assert(subjectOfCare != null) { "No patient id provided" }
      sessionID = mediatorRequest.correlationId
      assert(sessionID != null) { "No correlation id provided" }
      templateID = mediatorRequest.getParameter("id")
      assert(templateID != null) { "No template id provided" }
    }
    return ObjectFactory().createGetFormTemplate(getFormsRequest)
  }
}
```

Godin:
@jwalter thank you for the code - it gave an idea about reproducer, which is actually quite simple:

compilation of

```
    assert(args.size < 0) {
        println("Ouch")
    }
} // line 5
```

results in

```
...
        17: getstatic     #21                 // Field kotlin/_Assertions.ENABLED:Z
        20: ifeq          53
...
      LineNumberTable:
        line 2: 6
        line 3: 27
        line 4: 37
        line 2: 40
        line 5: 53
...
SourceDebugExtension:
  SMAP
  Example.kt
  Kotlin
  *S Kotlin
  *F
  + 1 Example.kt
  ExampleKt
  *L
  1#1,6:1
  *E
```

Will work on a fix in context of this ticket.

However this is also quite interesting in context of a https://github.com/jacoco/jacoco/issues/654#issuecomment-373667831 (CC @goodwinnk @yole): content of `assert` function clearly becomes inlined (see above bytecode at offsets 17 and 20), however there doesn't seem to be a way to determine this from `SMAP` as it doesn't mention `AssertionsJVM.kt`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineFilter.java
Godin: 👍

Godin:
And here is another example:

`F1.kt`
```
package example

inline fun call(crossinline init: () -> Unit) {
    return {
        init()
    }()
} // line 7
```

`F2.kt`
```
package org.jacoco.core.test.validation.kotlin.targets

fun test(): String {
    var res = "Fail"
    call {
        res = "OK"
    }
    return res
}
```

that produces `F2Kt$test$$inlined$call$1.class` without `KotlinDebug` strata, however with line numbers that go beyond line numbers in source file `F1.kt`

```
...
  public final void invoke();
    descriptor: ()V
    flags: ACC_PUBLIC, ACC_FINAL
    Code:
         0: nop
         1: aload_0
         2: getfield      #18                 // Field $res$inlined:Lkotlin/jvm/internal/Ref$ObjectRef;
         5: ldc           #34                 // String OK
         7: putfield      #40                 // Field kotlin/jvm/internal/Ref$ObjectRef.element:Ljava/lang/Object;
        10: nop
        11: return
      LocalVariableTable:
            1       9     1 $i$a$1$call   I
            0      12     0  this   Lorg/jacoco/core/test/validation/kotlin/targets/F2Kt$test$$inlined$call$1;
      LineNumberTable:
        line 5: 0
        line 9: 1
        line 10: 10
        line 6: 11
...
SourceFile: "F1.kt"
SourceDebugExtension:
  SMAP
  F1.kt
  Kotlin
  *S Kotlin
  *F
  + 1 F1.kt
  org/jacoco/core/test/validation/kotlin/targets/F1Kt$call$1
  + 2 F2.kt
  org/jacoco/core/test/validation/kotlin/targets/F2Kt
  *L
  1#1,8:1
  6#2,2:9
  *E
```

From which to me seems that we can change algorithm on following:
in `LineSection` of main `Kotlin` strata
find minimal `OutputStartLine`
excluding `LineInfo`s whose `LineFileID` corresponds to `SourceFile` attribute.

* for the above example this means: ignore `1#1,8:1` and from `6#2,2:9` take `9`
* this obviously also works for the previous example (https://github.com/jacoco/jacoco/pull/764#issuecomment-444874645): ignore the only existing `1#1,6:1`
* for current validation test: ignore `1#1,33:1` and from `22#1,2:34` take `34`
* for initial example (https://github.com/jacoco/jacoco/issues/763#issue-368413598): ignore `1#1,8:1`, from `2#2,2:9` and `2#3,2:11` take `9`

Any feedback/thoughts about this new algorithm are welcome.

Godin:
I was wrong about current validation test 😞 

```
object KotlinInlineTarget { // line 19

    inline fun inlined() {
        nop()
    }

    @JvmStatic

        inlined()

    }

} // line 32
```

```
SMAP
KotlinInlineTarget.kt
Kotlin
*S Kotlin
*F
+ 1 KotlinInlineTarget.kt
org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget
*L
1#1,33:1
22#1,2:34
*E
*S KotlinDebug
*F
+ 1 KotlinInlineTarget.kt
org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget
*L
28#1,2:34
*E
```

Godin:
```diff
 excluding `LineInfo`s whose `LineFileID` corresponds to `SourceFile` attribute
+and `InputStartLine` equal to `OutputStartLine`
```

Godin:
@jwalter could you please retest?

jwalter:
@Godin Works great! (My coverage went from 40% to 90%)

Godin:
@jwalter thanks a lot for testing! 👍 ❤️ 

Please excuse me for pedantism, but as a remark - when talking about metrics it is IMO very important to always specify metric, i.e. percentage of what? "instructions"? 😉

Nevertheless, while #763 initially is about mismatch between numbers of lines in XML report and source files, another aspect - is indeed a fix of metrics (instructions, lines, branches, complexity), which before this change were including inlined code that potentially comes from other libraries and therefore doesn't require tests on a side of consumers of library.

Thanks again for spotting bug and testing! ❤️ Hope there won't be other obstacles and we'll publish this soon.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -27,6 +27,9 @@ <h3>New Features</h3>
   <li>Branch added by the Kotlin compiler for "unsafe" cast operator is filtered
       out during generation of report
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/761">#761</a>).</li>
+  <li>Instructions inlined by Kotlin compiler from other class files are filtered
```
Godin: Strictly speaking not only "**other** class files" - can be the same.

org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget.kt
```diff
@@ -0,0 +1,36 @@
...
+        nop() // assertNotCovered()
+    }
+
+    @JvmStatic
+    fun main(args: Array<String>) {
+
+        inlined() // assertFullyCovered()
+
+        /* Following inlined method for some reasons doesn't appear in SMAP: */
+        assert(t()) // assertPartlyCovered(2, 2)
```
marchof: Looks like intrinsics or whatever kick-in when using stdlib. When inlining from another source file it seems to work:

        KotlinInlineFun.inlined() // assertFullyCovered()

Or what is the point here?

marchof: Thanks for clarification!

PR #766, opened 2018-10-11
Add default with Assertion Error to switch statement

renuka-fernando:
- Add default with Assertion Error to switch statement
- Replace switch statement which using only two enum types (out of four) to an If statement for readability.


marchof:
@renuka-fernando I'm not sure about the value of this changes. This is primarly about personal flavour. Would like to reject such changes.

renuka-fernando:
hi @marchof
I like 👍  but I feel it improves readability and future issues :)

All the constants of the enum are not used in the case statements and default statement is also not used. It is not recommonded.

marchof:
Please understand that we don't want to invest time about coding style. You probably find 100s of other places where the code could be written in a different style.

Issue #767, opened 2018-10-11
Probes inserted at locals that will be overridden


mkroghj:
When injecting probes into functions jacoco assumes that the arguments are not overridden as per the following JVM bytecode for class Test:

```
public final class Test
  minor version: 0
  major version: 52
  flags: ACC_PUBLIC, ACC_FINAL, ACC_SUPER
Constant pool:
   #1 = Utf8               Test
   #2 = Class              #1             // Test
   #3 = Utf8               java/lang/Object
   #4 = Class              #3             // java/lang/Object
   #5 = Utf8               Test.java
   #6 = Utf8               <init>
   #7 = Utf8               ()V
   #8 = NameAndType        #6:#7          // "<init>":()V
   #9 = Methodref          #4.#8          // java/lang/Object."<init>":()V
  #10 = Utf8               this
  #11 = Utf8               LTest;
  #12 = Utf8               foo
  #13 = Utf8               (I)I
  #14 = Utf8               main
  #15 = Utf8               ([Ljava/lang/String;)V
  #16 = Utf8               java/lang/System
  #17 = Class              #16            // java/lang/System
  #18 = Utf8               out
  #19 = Utf8               Ljava/io/PrintStream;
  #20 = NameAndType        #18:#19        // out:Ljava/io/PrintStream;
  #21 = Fieldref           #17.#20        // java/lang/System.out:Ljava/io/PrintStream;
  #22 = NameAndType        #12:#13        // foo:(I)I
  #23 = Methodref          #2.#22         // Test.foo:(I)I
  #24 = Utf8               java/io/PrintStream
  #25 = Class              #24            // java/io/PrintStream
  #26 = Utf8               println
  #27 = Utf8               (I)V
  #28 = NameAndType        #26:#27        // println:(I)V
  #29 = Methodref          #25.#28        // java/io/PrintStream.println:(I)V
  #30 = Utf8               Code
  #31 = Utf8               LineNumberTable
  #32 = Utf8               LocalVariableTable
  #33 = Utf8               SourceFile
{
  static int foo(int);
    descriptor: (I)I
    flags: ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: iload_0
         1: i2l
         2: lstore_0
         3: bipush        15
         5: ireturn

  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: ACC_PUBLIC, ACC_STATIC
    Code:
[...]

PR #774, opened 2018-10-24
Report code coverage correctly for Kotlin methods with default arguments

### LINKED ISSUE #773: Kotlin default arguments should not be filtered out
Because they can be arbitrary expressions and so can contain branches:

```
object KotlinDefaultArgumentsTarget {

    private fun f(a: Int, b: Int = if (a > 0) 1 else 2) {
    }

    @JvmStatic
    }

}
```

which are currently not shown

```
$ kotlinc KotlinDefaultArgumentsTarget.kt

$ java -javaagent:jacoco-0.8.2/lib/jacocoagent.jar \
    -cp .:kotlin-runtime.jar \
    KotlinDefaultArgumentsTarget

$ java -jar jacoco-0.8.2/lib/jacococli.jar \
    report jacoco.exec \
    --classfiles KotlinDefaultArgumentsTarget.class \
    --sourcefiles . \
    --html report
```

report

because this line associated with synthetic method:

```
  static void f$default(KotlinDefaultArgumentsTarget, int, int, int, java.lang.Object);
    descriptor: (LKotlinDefaultArgumentsTarget;IIILjava/lang/Object;)V
    flags: ACC_STATIC, ACC_BRIDGE, ACC_SYNTHETIC
    Code:
      stack=3, locals=5, args_size=5
         0: iload_3
         1: iconst_2
         2: iand
         3: ifeq          16
         6: iload_1
         7: ifle          14
        10: iconst_1
        11: goto          15
        14: iconst_2
        15: istore_2
        16: aload_0
        17: iload_1
        18: iload_2
        19: invokespecial #15                 // Method f:(II)V
        22: return
      LineNumberTable:
        line 3: 6
```

Godin:
As far as I understood

* Kotlin generates synthetic method with suffix `$default` that executes code to create default arguments and pass them to non-synthetic method that contains code of a method
* penultimate argument of synthetic method - is an `int` that specifies which default arguments are used (bit mask), therefore introduces two branches per default argument
* line numbers in this methods - are line numbers of default arguments
* first line number associated with instruction straight after first bit mask check
* synthetic method is not used for invocations that do not use default arguments

See https://github.com/JetBrains/kotlin/blob/v1.2.71/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/DefaultArgumentStubGenerator.kt

Since
* default arguments are human-written expressions
* and IMO makes sense to test both cases - usage of default and custom argument, i.e. generated branches

I propose to simply stop filtering out such methods. And later we might think about first two branches that are not associated with any line number.

Fixes #773

After this change for the same example as in #773 :

report


org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -22,6 +22,9 @@ <h2>Snapshot Build @qualified.bundle.version@ (@build.date@)</h2>
 
 <h3>New Features</h3>
 <ul>
+  <li>Do not filter out synthetic methods generated by Kotlin compiler that
+      contain values of default arguments
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/774">#774</a>).</li>
```
Godin: Maybe this should be `Fixed Bugs` @marchof WDYT?

marchof: Strictly speaking this is a bug. Even if nobody noticed it ;)

Maybe it is more readable in a positive way: "Report code coverage correctly for Kotlin methods with default arguments"

Godin: Updated.

But maybe this description forces us to really fix/workaround  line number weirdness.

Godin:
Actually there is one branch that is impossible to cover, because synthetic method is not used for invocation without default arguments, i.e. when bit mask is zero:

```
object KotlinDefaultArgumentsTarget {

    @JvmStatic
        one()
        /* next invocation doesn't use synthetic method: */
        one("a")
    }

    private fun one(a: String = "a") {
    }

}
```

```
    flags: ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=4, locals=1, args_size=1
         0: aload_0
         1: ldc           #10                 // String args
         3: invokestatic  #16                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
         6: getstatic     #20                 // Field INSTANCE:LKotlinDefaultArgumentsTarget;
         9: aconst_null
        10: iconst_1
        11: aconst_null
        12: invokestatic  #24                 // Method one$default:(LKotlinDefaultArgumentsTarget;Ljava/lang/String;ILjava/lang/Object;)V
        15: getstatic     #20                 // Field INSTANCE:LKotlinDefaultArgumentsTarget;
        18: ldc           #26                 // String a
        20: invokespecial #30                 // Method one:(Ljava/lang/String;)V
        23: return
      LocalVariableTable:
            0      24     0  args   [Ljava/lang/String;
      LineNumberTable:
        line 5: 6
        line 7: 15
        line 8: 23

  private final void one(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: ACC_PRIVATE, ACC_FINAL
    Code:
      stack=0, locals=2, args_size=2
         0: return
      LocalVariableTable:
            0       1     0  this   LKotlinDefaultArgumentsTarget;
            0       1     1     a   Ljava/lang/String;
      LineNumberTable:
        line 11: 0

  static void one$default(KotlinDefaultArgumentsTarget, java.lang.String, int, java.lang.Object);
    descriptor: (LKotlinDefaultArgumentsTarget;Ljava/lang/String;ILjava/lang/Object;)V
    flags: ACC_STATIC, ACC_BRIDGE, ACC_SYNTHETIC
    Code:
      stack=2, locals=4, args_size=4
         0: iload_2
         1: iconst_1
         2: iand
         3: ifeq          9
         6: ldc           #26                 // String a
         8: astore_1
         9: aload_0
        10: aload_1
        11: invokespecial #30                 // Method one:(Ljava/lang/String;)V
        14: return
      LineNumberTable:
        line 10: 6
```

kotlindefaultargumentstarget

The only idea that comes to my mind - is to exclude all branches. In this case instruction coverage still can be used to determine if all cases were covered or not.

Godin:
Don't forget to squash and clean description if you decide to merge :wink:

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java
```diff
@@ -21,11 +21,22 @@
...
-					methodNode.instructions.getLast());
+		if ((methodNode.access & Opcodes.ACC_SYNTHETIC) == 0) {
+			return;
 		}
+
+		if (methodNode.name.startsWith("lambda$")) {
+			return;
+		}
+
+		if (methodNode.name.endsWith(KotlinDefaultArgumentsFilter.SUFFIX)
```
marchof: Instead of duplicating the logic here I would prefer to encapsulate it in the respective filters, i.e.

* `KotlinGeneratedFilter.isKotlinClass(IFilterContext cts)` and
* `KotlinDefaultArgumentsFilter.isDefaultArgumentMethod(String name)`

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -0,0 +1,75 @@
...
+ * ILOAD maskVar
+ * ICONST_x, BIPUSH, SIPUSH, LDC or LDC_W
+ * IAND
+ * IFEQ label
+ * default argument
+ * label:
+ * </pre>
+ * 
+ * Where <code>maskVar</code> is penultimate argument of synthetic function with
+ * suffix {@link #SUFFIX "$default"}. And it can't be zero - invocation with all
```
marchof: In Eclipse I get a warning here, as this reference is not visible in JavaDoc. The member can become private if the check logic is encapsulated in a public methods (see comment on `SyntheticFilter`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -0,0 +1,75 @@
...
+		for (AbstractInsnNode i = methodNode.instructions
+				.getFirst(); i != null; i = i.getNext()) {
+			if (AbstractInsnNode.VAR_INSN != i.getType()) {
+				continue;
+			}
+			if (maskVar != ((VarInsnNode) i).var) {
+				continue;
+			}
+
+			final AbstractInsnNode jump = i.getNext().getNext().getNext();
```
marchof: So we assume that this is the only use case for the variable where maskVar was stored inistially? Wouldn't it be more safe to implement a small matcher here to verify that it the actual byte code sequence described in the class comment. We recently learned that argument variables can be re-used (#767) and Kotlin is used for Android development.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
Godin: when replacing field by method, IntelliJ suggests to make it package-local, which results in the same warning, so I'll just remove link from description

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinGeneratedFilter.java
```diff
@@ -23,6 +23,10 @@
 
 	static final String KOTLIN_METADATA_DESC = "Lkotlin/Metadata;";
 
+	static boolean isKotlinClass(final IFilterContext context) {
```
marchof: As this method is used by other filters too, I wonder whether the `GeneratedFilter` is the right place. Maybe a separate `KotlinFilterUtils` class would be a better place?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -0,0 +1,108 @@
...
+					break;
+				}
+				next();
+				nextIs(Opcodes.IAND);
+				nextIs(Opcodes.IFEQ);
+				if (cursor == null) {
+					return;
+				}
+				ignore.add(cursor);
+				cursor = instructionAfterLabel(((JumpInsnNode) cursor).label);
```
marchof: If we make `skipNonOpcodes()` of the base class visible we can use this method here:

    cursor = ((JumpInsnNode) cursor).label;
    skipNonOpcodes();

org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinDefaultArgumentsTarget.kt
```diff
@@ -0,0 +1,36 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation.kotlin.targets
+
+/**
+ * This test target is function with default arguments.
```
marchof: Wording sounds strange to me. Proposal: "Test target for functions with default arguments."

Godin: applied your proposal

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinGeneratedFilter.java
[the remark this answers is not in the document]
Godin: done

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilterTest.java
```diff
@@ -0,0 +1,98 @@
...
+		final MethodNode m = new MethodNode(InstrSupport.ASM_API_VERSION,
+				access, name, descriptor, null, null);
+
+		m.visitVarInsn(Opcodes.ILOAD, 2);
+		m.visitInsn(Opcodes.ICONST_1);
+		m.visitInsn(Opcodes.IAND);
+		final Label label = new Label();
+		m.visitJumpInsn(Opcodes.IFEQ, label);
+		// default argument
+		m.visitLdcInsn(42);
```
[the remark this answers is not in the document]
Godin: done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
Godin: done

Godin:
@marchof 

> As this method is used by other filters too, I wonder whether the GeneratedFilter is the right place. Maybe a separate KotlinFilterUtils class would be a better place?

If we want to use it in all Kotlin filters (your other question), then maybe better to introduce `KotlinFilter` that will have this method and call other filters?

> I wonder whether KotlinGeneratedFilter.isKotlinClass() should also be used in the existing Kotlin filters as a shortcut before iterating through all instructions of a method. WDYT?

This also raises question - shouldn't we exclude Kotlin classes from non-Kotlin filters such as one for `try-with-resources` ? In any case IMO this is unrelated to this PR.

And hence I would prefer to not do any changes regarding these two remaining points in this PR.

marchof:
@Godin  Sure, usage and re-location of `isKotlinClass()` should be handeled in a separate PR.

After the builds are green I'll merge this. Thanks for addressing my remarks!

PR #781, opened 2018-10-31
Configurable logging of exceptions

### LINKED ISSUE #32: Jacoco maven plugin clogs up logs with Exceptions
When I run Jacoco using the Jacoco maven plugin on a project, I get an enourmous amount of these exceptions in the log:

```
java.lang.instrument.IllegalClassFormatException: Error while instrumenting class org/apache/maven/surefire/report/ConsoleOutputCapture$ForwardingPrintStream.
    at org.jacoco.agent.rt_kqcpih.CoverageTransformer.transform(CoverageTransformer.java:91)
    at sun.instrument.TransformerManager.transform(TransformerManager.java:188)
    at sun.instrument.InstrumentationImpl.transform(InstrumentationImpl.java:424)
    at java.lang.ClassLoader.defineClass1(Native Method)
    at java.lang.ClassLoader.defineClass(ClassLoader.java:791)
    at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
    at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
    at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
    at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
    at java.security.AccessController.doPrivileged(Native Method)
    at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:423)
    at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
    at java.lang.ClassLoader.loadClass(ClassLoader.java:356)
    at org.apache.maven.surefire.junit.JUnit3Provider.invoke(JUnit3Provider.java:84)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:601)
    at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray(ReflectionUtils.java:189)
    at org.apache.maven.surefire.booter.ProviderFactory$ProviderProxy.invoke(ProviderFactory.java:165)
    at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:85)
    at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:103)
    at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:74)
Caused by: java.lang.IllegalStateException: Class org/apache/maven/surefire/report/ConsoleOutputCapture$ForwardingPrintStream is already instrumented.
    at org.jacoco.agent.rt_kqcpih.core.internal.instr.ClassInstrumenter.assertNotInstrumented(ClassInstrumenter.java:133)
    at org.jacoco.agent.rt_kqcpih.core.internal.instr.ClassInstrumenter.visitField(ClassInstrumenter.java:81)
    at org.jacoco.agent.rt_kqcpih.asm.ClassVisitor.visitField(Unknown Source)
    at org.jacoco.agent.rt_kqcpih.asm.ClassReader.accept(Unknown Source)
    at org.jacoco.agent.rt_kqcpih.asm.ClassReader.accept(Unknown Source)
    at org.jacoco.agent.rt_kqcpih.core.instr.Instrumenter.instrument(Instrumenter.java:69)
    at org.jacoco.agent.rt_kqcpih.core.instr.Instrumenter.instrument(Instrumenter.java:82)
    at org.jacoco.agent.rt_kqcpih.CoverageTransformer.transform(CoverageTransformer.java:89)
    ... 24 more
```

I get these exceptions for a lot of classes that come from external libraries, like Spring, Hibernate, JPA, Maven, etc., yet Jacoco seems to work fine and generates the reports correctly. I think that the issue is caused by the following line of code:

JacocoAgent.java, line 158:

  final JacocoAgent agent = new JacocoAgent(options,
                new IExceptionLogger() {
                    public void logExeption(final Exception ex) {
                        ex.printStackTrace();
                    }
                });

Maybe it's a good idea not to have these exceptions in the log when they don't affect the correct behaviour of Jacoco? Or at least make it configurable, so that people with lots of exceptions in the log can turn it off? Also, printing them to the STDOUT is perhaps not be best way of logging things. Maven provides a org.apache.maven.plugin.logging.Log object, that should be used to log things that occur during a Maven Build.

papegaaij:
This pull request introduces a new option `log`, which can be used to limit or suppress exception logging. This fixes #32.

In some cases it may not be possible (or feasible) to prevent duplicate instrumentation. For example, in our case we use Arquillian to run tests inside our application server. However, part of the code also runs inside the test runner. Therefore, the agent needs to be loaded at both java processes, causing duplicate instrumentation. At the moment this produces an enormous amount of logging (several GBs per run), but the tests run fine and the instrumentation is correct.

### DIFF
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/Agent.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/Agent.java
@@ -46,7 +46,7 @@ public class Agent implements IAgent {
 	 */
 	public static synchronized Agent getInstance(final AgentOptions options) {
 		if (singleton == null) {
-			final Agent agent = new Agent(options, IExceptionLogger.SYSTEM_ERR);
+			final Agent agent = new Agent(options, getLogger(options));
 			agent.startup();
 			Runtime.getRuntime().addShutdownHook(new Thread() {
 				@Override
@@ -67,11 +67,32 @@ public class Agent implements IAgent {
 	 * @throws IllegalStateException
 	 *             if no Agent has been started yet
 	 */
-	public static synchronized Agent getInstance() throws IllegalStateException {
+	public static synchronized Agent getInstance()
+			throws IllegalStateException {
 		if (singleton == null) {
 			throw new IllegalStateException("JaCoCo agent not started.");
 		}
 		return singleton;
+	}
+
+	/**
+	 * Returns the {@link IExceptionLogger} instance configured by the given
+	 * options.
+	 * 
+	 * @param options
+	 *            options to configure the agent
+	 * @return the resolved {@link IExceptionLogger}
+	 */
+	public static IExceptionLogger getLogger(final AgentOptions options) {
+		switch (options.getLog()) {
+		default:
+		case full:
+			return IExceptionLogger.SYSTEM_ERR;
+		case normal:
+			return IExceptionLogger.NORMAL;
+		case silent:
+			return IExceptionLogger.SILENT;
+		}
 	}
 
 	private final AgentOptions options;
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/IExceptionLogger.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/IExceptionLogger.java
@@ -27,6 +27,23 @@ public interface IExceptionLogger {
 	};
 
 	/**
+	 * Implementation which dumps the exception message to System.err.
+	 */
+	IExceptionLogger NORMAL = new IExceptionLogger() {
+		public void logExeption(final Exception ex) {
+			System.err.println(ex.getMessage());
+		}
+	};
+
+	/**
+	 * Implementation which is silent
+	 */
+	IExceptionLogger SILENT = new IExceptionLogger() {
+		public void logExeption(final Exception ex) {
+		}
+	};
+
+	/**
 	 * Logs the given exception.
 	 * 
 	 * @param ex
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -47,12 +47,13 @@ public final class PreMain {
 		final IRuntime runtime = createRuntime(inst);
 		runtime.startup(agent.getData());
 		inst.addTransformer(new CoverageTransformer(runtime, agentOptions,
-				IExceptionLogger.SYSTEM_ERR));
+				Agent.getLogger(agentOptions)));
 	}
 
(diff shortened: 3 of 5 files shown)


Godin:
@papegaaij Could you please elaborate more why in your case instrumentation happens twice?

> the agent needs to be loaded at both java processes

How usage of two java agents in **two different java processes** can result in repeated instrumentation?

> In some cases it may not be possible (or feasible) to prevent duplicate instrumentation.

And why existing options `excludes` and `exclclassloader` are not sufficient?

----

So far I'm inclined to reject this, because these exceptions for example reveal cases when different JaCoCo versions are used for same classes, and usage of different versions within single toolchain is bad. And so IMO  we should not encourage ability to blindly ignore configuration problems, which can lead to other problems that then will be hard to diagnose. @marchof WDYT?

papegaaij:
@Godin I decided to dive deeper into these duplicate instrumentation issues to find the actual cause of the duplicate instrumentation. I was under the assumption that the classes were being instrumented by the surefire vm instances before being deployed on wildfly, but I was wrong. It turned out that the agent was added to multiple times to the same vm on every next run of a testclass. This caused the agent to be loaded up to 3 times, naturally causing duplicate instrumentation.

Still I believe the excessive flooding of exceptions is too much for this problem. Maybe the logging should only log the first 10 exceptions and then revert to just logging the error message. Another improvement could be to check the version of Jacoco and not log an error if it matches the version used to instrument the class with.

You are free to do what you like with this PR. Maybe you can keep the option to not log the stacktraces, but only the message, maybe you can use it to provide other means of logging.

marchof:
@papegaaij I agree with @Godin: The only valid measure here is to fix the root cause. Any suppression or reduction of debug output will make it harder to help people finding the cause.

Godin:
> check the version of Jacoco and not log an error if it matches the version used to instrument the class with

Currently there is no markers in instrumented classes that allow to determine version, and what is more important

> agent to be loaded up to 3 times

consider the case when multiple agents configured to instrument same classes and save results into different locations. Even if agents have same version, suppression of error messages in this case IMO is completely misleading and error-prone - your setup is broken.

Furthermore IMO usage of two agents makes no sense at all. And so detection of this early prior to instrumentation will be a better contribution :wink:

papegaaij:
I agree. An accurate message pointing me to the fact that the agent was loaded multiple times would have saved me a lot of time :smile:

PR #789, opened 2018-11-15
Add test: computation of lines range should respect ignored instructions

Godin:
Without following condition all unit tests still pass

https://github.com/jacoco/jacoco/blob/4d4d02e6dbfb17f454bf4579ee13209e9f035154/org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java#L128

however this condition is important for #763 , so I propose to add test.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodCoverageCalculatorTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodCoverageCalculatorTest.java
@@ -108,6 +108,21 @@ public class MethodCoverageCalculatorTest {
 	}
 
 	@Test
+	public void should_exclude_ignored_instructions_from_computation_of_first_and_last_lines() {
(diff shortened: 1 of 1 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodCoverageCalculatorTest.java
```diff
@@ -107,6 +107,21 @@ public void should_ignore_instruction_range() {
 		assertLine(3, 0, 1, 0, 0);
 	}
 
+	@Test
+	public void should_exclude_ignored_instructions_from_computation_of_first_and_last_lines() {
+		InsnNode i1 = addInsn(1, false);
+		addInsn(2, false);
+		InsnNode i2 = addInsn(3, false);
```
marchof: Rename this to `i3` for consistency?

PR #796, opened 2018-12-10
Add unit tests for AbstractMatcher

Godin:


marchof:
@Godin Thanks for adding the test! I would prefer to remove the `test_` prefix to comply with the new nomenclature. I understand that there are multiple tests per test method, which is ok for me. This results in missing test descriptions, which are not too obvious (at least for) me. I would suggest to add inline comments per test. For example:

	@Test
	public void test_skipNonOpcodes() {
		m.visitFrame(Opcodes.F_FULL, 0, null, 0, null);
		final Label label = new Label();
		m.visitLabel(label);
		m.visitLineNumber(42, label);
		m.visitInsn(Opcodes.NOP);

		// should skip all non opcodes
		matcher.cursor = m.instructions.getFirst();
		matcher.skipNonOpcodes();
		assertSame(m.instructions.getLast(), matcher.cursor);

		// should not skip if current node is opcode
		matcher.skipNonOpcodes();
		assertSame(m.instructions.getLast(), matcher.cursor);

		// should ignore null cursor
		matcher.cursor = null;
		matcher.skipNonOpcodes();
	}

Godin:
@marchof I added comments and renamed methods.

Issue #787, opened 2018-11-10
Jacoco is not show koltin sources in coverage report


javadev:
This is a issue tracker. Please use our mailing list for general questions: 
https://groups.google.com/forum/?fromgroups=#!forum/jacoco

Also check FAQ before opening an issue: http://www.jacoco.org/jacoco/trunk/doc/faq.html

### Steps to reproduce

JaCoCo version: 0.8.2
Operating system: Win10
Tool integration: Maven/Ant/API/Other

Maven

Project sources: https://github.com/javadev/classic-cherries-kotlin

### Expected behaviour

Source code should be present in coverage report

### Actual behaviour

Only coverage percent information is present

![jacoco-kotlin](https://user-images.githubusercontent.com/348496/48300350-18875980-e4e4-11e8-8da0-917c1dd1fc0a.png)

javadev:
It happened because of wrong package name. After fixing package name the problem was solved.

Issue #751, opened 2018-08-20
Invalid missing line coverage in report


kevcodez:
### Steps to reproduce

Clone https://github.com/kevcodez/pubg-api-kotlin and run `./gradlew test`. It generates this [Jacoco report](https://gist.github.com/kevcodez/eec43fe9f26868b296d6b391859aa6c2).

JaCoCo version: 0.8.1
Operating system: Mac
Tool integration: Gradle

### Expected behaviour

Missed line coverage in source file is the sum of missed source files in class' methods. See PlayerLogin class in the report.

### Actual behaviour

Sourcefile missed line coverage says zero, even though it should be > 0.

Am I missing something here?

marchof:
Line coverage is the sum of all code with the line. Depending on the syntax and the compiler multiple methods can account to a single line. Due to our definition a line is not covered, if no instruction in this line is executed.

Is your example `getAccountId` and `setAccountId` (both are not executed) overlap with the constructor, which is fully covered. Therefore there lines are counted as covered. In the `line` element you can see how many instructions are covered and missed on every line.

marchof:
BTW, next release of JaCoCo will filter many of the methods which are not present in source but only generated by the Kotlin compiler.

kevcodez:
Thanks for clearing that up. Is there an ETA for the next version of Jacoco @marchof ?

marchof:
Probably these days.

Godin:
Let me clarify/add a bit:

Explanations above are about the fact that multiple methods can share the same line numbers and therefore it is incorrect to compare sum of lines in methods with lines in class or source file.

This is unrelated to

> filter many of the methods which are not present in source but only generated by the Kotlin compiler

because such methods anyway do not have line numbers and that's exactly how they are filtered out - see #689 , so this filter affects counters of instructions and methods, but not lines.

There is another filter - see #707 , it will affect number of instructions and branches in `PlayerLogin  .getAccountId`, but not lines, since again - it has nothing to do with the fact that methods share the same line numbers.

And as visual explanation - here is HTML reports for `PlayerLogin`

with version `0.8.1`

![playerlogin](https://user-images.githubusercontent.com/138671/44368446-232ce680-a4d3-11e8-80c2-26aa6bc8b7a8.png)

and with latest snapshot of version `0.8.2`

![playerlogin](https://user-images.githubusercontent.com/138671/44368459-32139900-a4d3-11e8-9080-5dedafa80b3a.png)

[...]

PR #801, opened 2018-12-19
Show message in HTML report when source file can't be found

Godin:
One of common problems already described in [our FAQ](https://www.jacoco.org/jacoco/trunk/doc/faq.html) and still frequently faced by users - incorrect location of source files, which leads to absence of pages with sources in HTML report:

* https://stackoverflow.com/a/53833558/244993
* https://groups.google.com/d/msg/jacoco/wmwYxLKKn3Q/5AH5qSGBDgAJ
* https://github.com/jacoco/jacoco/issues/787#issuecomment-437578208
* https://github.com/jacoco/jacoco/issues/751#issuecomment-414480178

So I propose to add following message to a page with class methods

```
Source file `org/example/Example.java` was not found during generation of report.
```

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -30,6 +30,8 @@
   <li>Instructions inlined by Kotlin compiler are filtered out during generation
       of report
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/764">#764</a>).</li>
+  <li>HTML report shows message when source file can't be found
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/801">#801</a>).</li>
 </ul>
 
 <h3>Fixed Bugs</h3>
--- a/org.jacoco.report.test/src/org/jacoco/report/internal/html/page/ClassPageTest.java
+++ b/org.jacoco.report.test/src/org/jacoco/report/internal/html/page/ClassPageTest.java
@@ -47,6 +47,7 @@ public class ClassPageTest extends PageTestBase {
 		page.render();
 
 		final Document doc = support.parse(output.getFile("Foo.html"));
+		assertEquals("", support.findStr(doc, "doc/body/p[1]"));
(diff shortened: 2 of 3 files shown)


Godin:
> But what if users do not provide source file at all (e.g. because they don't have it). In this case the same message would still be included in the report.

@marchof I don't think that this little hint will bother such users.

marchof:
@Godin In this case there are more common problems where we can add a hint to report:

* Only class with different class id found
* Class files without debug information

I would prefer to put the checks and message construction to a seperate, testable class `DiagnosticsInfo`.

Godin:
> In this case there are more common problems where we can add a hint to report:
> * Only class with different class id found
> * Class files without debug information

@marchof baby steps, I was planning to add these via follow-up PRs 😉 

> I would prefer to put the checks and message construction to a seperate, testable class `DiagnosticsInfo`.

Did I get it right - you just wanna pull entire `if` into separate class? i.e. without removal of dependency on `HTMLElement body` and `ILinkable sourcePage` ?

marchof:
@Godin I haven't thought about the design in detail. My idea is to encapsulate the checks and the creation of warnings (as html elements) in a separate class. So for example the hint for missing line information can be shown on different page types.

Godin:
@marchof I agree with idea :thumbsup: Other than simple `if` extraction, design looks foggy to me for the time being because this message targets only single page. This message alone IMO is already quite valuable, so I'd like to propose merge as is and I'll work on addition of other messages and thus on the re-design in a next PR. Missing unit tests were added.

Godin:
BTW should be noted that there are actually three cases of

> Class files without debug information

* `javac -g:source` produces only `SourceFile` attribute - line coverage can not be computed, however can link with source file without highlighting 😆 
* `javac -g:lines` produces only `LineNumberTable` attribute - line coverage can be computed, but can not link with source file
* `javac -g:none` - no debug information at all

PR #803, opened 2018-12-22
Update filter for Kotlin coroutines that restore state

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinCoroutineFilter.java
```diff
@@ -72,32 +73,43 @@ private void match(final MethodNode methodNode,
...
 
-			for (AbstractInsnNode i = methodNode.instructions
-					.getFirst(); i != null; i = i.getNext()) {
+			int suspensionPoint = 1;
+			for (AbstractInsnNode i = cursor; i != null
+					&& suspensionPoint < s.labels.size(); i = i.getNext()) {
 				cursor = i;
 				nextIsVar(Opcodes.ALOAD, "COROUTINE_SUSPENDED");
 				nextIs(Opcodes.IF_ACMPNE);
+				AbstractInsnNode continuationAfterLoadedResult = cursor;
```
marchof: At this point we know that cursor is `JumpInsnNode`. Cast happens later, which confused me. Also `continuationAfterLoadedResult` first points to jump insn and then later gets replaced with jump target. Both can be fixed if we replace this line with:


	if (cursor == null) {
		continue;
	}
	final AbstractInsnNode continuationAfterLoadedResult = skipNonOpcodes(((JumpInsnNode) cursor).label);

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
```diff
@@ -158,11 +158,21 @@ final void next() {
...
 	final void skipNonOpcodes() {
+		cursor = skipNonOpcodes(cursor);
+	}
+
+	/**
+	 * Returns first instruction from given and following it that is not
+	 * {@link AbstractInsnNode#FRAME}, {@link AbstractInsnNode#LABEL},
+	 * {@link AbstractInsnNode#LINE}.
+	 */
+	final AbstractInsnNode skipNonOpcodes(AbstractInsnNode cursor) {
```
marchof: As this method does not change state (unlike the others) I would prefer to declare it `static`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinCoroutineFilter.java
Godin: Believe me - this is exactly how first implementation looked, then I decided to make flow linear 😆

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java
Godin: Indeed. Fixed!

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinCoroutineFilter.java
Godin: Fixed.

PR #805, opened 2018-12-23
Show additional diagnostic information in HTML report

marchof:
In a follow-up to #801 more diagnostic information should be added to the HTML report:

* Only class with different class id found (`IClassCoverage.isNoMatch()`): ClassPage
* Class file without debug source file information (`IClassCoverage.getSourceFileName() == null`): ClassPage
* Class files without debug source file information (`IPackageCoverage.getSourceFiles().isEmpty()` for non empty packages): PackagePage
* Class files without debug line information (`ICoverageCounter.getLineCounter().getTotal() == 0`): All pages


Godin:
@marchof IMO one more case that will be useful to diagnose - is empty report due to no classes being analyzed. However seems that this will require resolution of #806 to avoid false-negatives when all analyzed classes were filtered out.

Godin:
And IMO one more case - absence of execution data.

PR #811, opened 2018-12-28
Add example that shows difference between two JaCoCo XML reports

Godin:
I used this little utility to check changes in filters on a real projects. And would like to add it as an example.


marchof:
@Godin I'm not sure whether this more like a tool than a educational example about our API usage. Have you considered adding this to the CLI instead?

Godin:
@marchof I thought about this option - as it is now, it fits well into examples, because
* we don't have example that works with XML, whereas I believe that this is the better path for third-party integrations than processing of exec-file
* inclusion into CLI will require more hardening

org.jacoco.examples/src/org/jacoco/examples/XMLDiff.java
```diff
@@ -0,0 +1,231 @@
...
+				final int cb = Integer.parseInt(attributes.getValue("cb"));
+				s.put(lineNumber, new LineInfo(mi, ci, mb, cb));
+			}
+		}
+
+		@Override
+		public InputSource resolveEntity(final String publicId,
+				final String systemId) {
+			if (publicId.startsWith("-//JACOCO//DTD Report")) {
+				return new InputSource(new StringReader(""));
```
marchof: This is an example and should follow best practices:

- check for exact DTD version to ensure compatibility
- Supply actual DTD otherwise this will fail when we merge #813

marchof: Sorry, my bad: I overlooked that `parseAsOptionalInt()` already checks for `null` values.

PR #813, opened 2019-01-02
Do not print default zero values of attributes in XML report

Godin:
On example of our own XML report majority of lines do not have branches:

```
$ cat jacoco.xml | tr '>' '\n' | grep "<line" | wc -l
6403
$ cat jacoco.xml | tr '>' '\n' | grep "mb=\"0\" cb=\"0\"" | wc -l
5522
```

By not printing `mb="0" cb="0"` we can save 77308 bytes, which is about 7% out of 915390.

And by not printing zeros at all for [`IMPLIED` attributes `mi`, `ci`, `mb` and `cb`](https://github.com/jacoco/jacoco/blob/9c45dab3e9d9ed888481005095b30d3e5b55b338/org.jacoco.report/src/org/jacoco/report/xml/report.dtd#L68-L75) (as implemented in this PR) we can save 127498 bytes, which is about 14%.

----

Furthermore (not implemented in this PR) by [changing DTD to mark `missed` and `covered` as not `REQUIRED`](https://github.com/jacoco/jacoco/blob/9c45dab3e9d9ed888481005095b30d3e5b55b338/org.jacoco.report/src/org/jacoco/report/xml/report.dtd#L81-L84) and by not printing these too, we can save additional 93098 bytes, which is about 10%.

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -52,6 +52,10 @@
 <ul>
   <li>JaCoCo now depends on ASM 7.0
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/760">#760</a>).</li>
+  <li>Default zero values are not printed in XML report for <code>mi</code>,
+      <code>ci</code>, <code>mb</code> and <code>cb</code> attributes of
+      <code>line</code> element
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/813">#813</a>).</li>
 </ul>
 
 <h2>Release 0.8.2 (2018/08/21)</h2>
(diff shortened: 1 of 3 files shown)


marchof:
This looks like a major braking change as probably probably nobody supplies our DTD.

@Godin Do you really think this optimization is worth it?

Godin:
@marchof 

> probably nobody supplies our DTD

what about embedment of DTD into XML:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<!DOCTYPE report [
  <!ELEMENT report ... >
  ...
]>
<report>
...
```

or maybe even XSD ?

---

> looks like a major braking change

The second part of description **not implemented here** about change for attributes marked as required is indeed a breaking change.

The first implemented part about change for attributes **already marked as optional** [since a very long time](https://github.com/jacoco/jacoco/blame/b075623319d8b01b02b75a530dfa6d3b79392d1e/org.jacoco.report/src/org/jacoco/report/xml/report.dtd#L66-L73) did not looked as breaking to me, exactly because consumers already must not assume their presence.

Unfortunately you're probably right - no one reads DTD as like any documentation. [SonarQube for example assumes presence these of attributes](https://github.com/SonarSource/sonar-jacoco/blob/1.0.1.143/src/main/java/org/sonar/plugins/jacoco/XmlReportParser.java#L108-L115) 😞 

---

> worth it?

**14%** just for our small project and just for first implemented part, and *27%* in total with second part...

marchof:
I like the idea of inlining the DTD! Also we can improve our DTD to define default values. Then existing parsers shouldn't be affected.

PR #815, opened 2019-01-05
Add filter for Kotlin not-null assertion operator

Godin:
https://groups.google.com/d/msg/jacoco/V4z41kgUnK8/JbqDkAZ5FgAJ


Godin:
BTW IMO there is no need to generate explicit `null`-check in cases such as

```
fun example(x: String?) {
   val y:  = x!!.length
   // ...
}
```

because `invokevirtual` anyway will throw NPE

```
         0: aload_1
         1: dup
         2: ifnonnull     8
         5: invokestatic  #12                 // Method kotlin/jvm/internal/Intrinsics.throwNpe:()V
         8: invokevirtual #18                 // Method java/lang/String.length:()I
        11: istore_2
```

while definitely useless in cases such as

```
fun example(x: String?) {
   val y = x!!.length
   // ...
}
```

@goodwinnk maybe this is something what can be improved in Kotlin compiler?

goodwinnk:
@Godin There's a plan to unify different null checks in Kotlin in favor to simply throwing Java NPE: https://youtrack.jetbrains.com/issue/KT-22275

Godin:
@goodwinnk good to know, thanks!

PR #817, opened 2019-01-06
Preserve empty class and sourcefile nodes in XML report

### LINKED ISSUE #806: Preserve empty IClassCoverage nodes in CoverageBuilder and in XML report
Hey all,

Integrations that consume XML reports are unable to distinguish between two reasons of classes and sources absence in the report

* does not contain executable lines (e.g. Java interfaces without default methods) or was filtered out (e.g. marked with `@Generated` annotation)
* was not analyzed at all

and therefore cannot inform user about last one.

This happens because of

https://github.com/jacoco/jacoco/blob/v0.8.2/org.jacoco.core/src/org/jacoco/core/analysis/CoverageBuilder.java#L101-L102

which was added a long time ago in version 0.3.2 by

https://github.com/jacoco/jacoco/commit/e47d22d5205ec826f82b3072289f56564a3f74d6

----

This is exactly what happens with both SonarQube integrations (one [that reads XML](https://github.com/SonarSource/sonar-jacoco) and another [that reads exec-file](https://github.com/SonarSource/sonar-java/tree/5.9.2.16552/java-jacoco)) - https://github.com/jacoco/jacoco/issues/797 : in absence of information for source file (in XML report in first case, `CoverageBuilder` in second), SonarQube assumes that it was not executed/analyzed and tries to estimate from syntax tree of source file its missed coverage, and this estimation is not consistent with filtering of `@Generated` by JaCoCo.

----

There are some workarounds for integrations that read exec-file - e.g. custom `ICoverageVisitor` that will track `IClassCoverage.getSourceFileName()` prior delegation to `CoverageBuilder`.

However there are doesn't seem to be any option for integrations that read only XML reports.

----

Could you please (re)consider preservation of empty `IClassCoverage` nodes in `CoverageBuilder` and in XML report, and skip them only when generating HTML ?

Cheers,
Michael

Godin:
Fixes #806

Our own XML report before this change is 915700 bytes, after is 920811 - increase is just 0.6%.

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
@@ -222,9 +222,9 @@ final class ReportSupport {
 						c.getName()));
 			}
 		}
-		if (bundle.getClassCounter().getTotalCount() > 0
-				&& bundle.getLineCounter().getTotalCount() == 0) {
-			log.warn("To enable source code annotation class files have to be compiled with debug information.");
+		if (!bundle.isEmpty() && bundle.getLineCounter().getTotalCount() == 0) {
+			log.warn(
+					"To enable source code annotation class files have to be compiled with debug information.");
 		}
 	}
 
--- a/org.jacoco.ant/src/org/jacoco/ant/ReportTask.java
+++ b/org.jacoco.ant/src/org/jacoco/ant/ReportTask.java
@@ -593,8 +593,7 @@ public class ReportTask extends Task {
 	}
 
 	private void checkForMissingDebugInformation(final ICoverageNode node) {
-		if (node.getClassCounter().getTotalCount() > 0
-				&& node.getLineCounter().getTotalCount() == 0) {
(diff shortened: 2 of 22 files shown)


marchof:
As we have this snippet now all over the place:

    coverage.getInstructionCounter().getTotalCount() == 0

I would like to propose to add a new method to `ICoverageNode`:

	/**
	 * Checks whether this is an empty node.
	 * 
	 * @return <code>true</code> is this node does not contain any instructions
	 */
	boolean isEmpty();

Godin:
@marchof added

marchof:
@Godin Thx for adding `isEmpty()`. I found three more candidates for using it:

* `org.jacoco.core.internal.analysis.ClassAnalyzer.addMethodCoverage(String, String, String, InstructionsBuilder, MethodNode)`
* `org.jacoco.ant.ReportTask.checkForMissingDebugInformation(ICoverageNode)`
* `org.jacoco.maven.ReportSupport.logBundleInfo(IBundleCoverage, Collection<IClassCoverage>)`

Godin:
@marchof thanks for spotting this - made changes, hopefully correctly 😉

PR #820, opened 2019-01-06
Upgrade maven-plugin-plugin and maven-shade-plugin

slachiewicz:
Fixes problem with compilation for bytecode 9+

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -364,7 +364,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-plugin-plugin</artifactId>
-          <version>3.5</version>
+          <version>3.6.0</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


slachiewicz:
Yes, i'll work on that.

org.jacoco.build/pom.xml
```diff
@@ -734,14 +734,6 @@
       </build>
     </profile>
 
-    <!--
-    Following profiles enable compilation into bytecode version 9
-    when requested "bytecode.version" greater than 9,
-    because this is maximum that can be processed by
-    maven-shade-plugin (https://issues.apache.org/jira/browse/MSHADE-289)
```
Godin: [First builds of JDK 13 EA already available for testing](http://jdk.java.net/13/) and I think same trick will be needed to test new bytecode version, but ok we can restore comment and add profile when will be needed.

marchof:
@Godin Do the changed profiles have an impact on your bytecode version matrix?

    org.jacoco.doc/docroot/doc/build.html

Godin:
@marchof this matrix shows defaults, so no impact 😉

Issue #314, opened 2015-05-22
Drop support for Java 5 runtime?


marchof:
I wonder whether we could drop support for Java 5 runtime with the following benefits:

* Simplify test/build setup: With the advent of Java 9 we're now testing on 5 different JDK versions.
* Allow dynamic attachment of JaCoCo agent
* Simplify runtime data access
* allow coverage on bootstrap classes
* support re-transformation (no schema changes any more)

The last three become possible by adding companion classes to the bootstrap classpath with new method java.lang.instrument.Instrumentation.appendToBootstrapClassLoaderSearch().

kibertoad:
So much yes.

Godin:
@marchof I'm not against drop of support.

However to prevent misunderstanding for others:
* we already allow coverage on bootstrap classes
  * in online instrumentation mode partially due to inability to retransform already loaded classes
  * and I suppose that completely in offline mode?
* ability to use "appendToBootstrapClassLoaderSearch" for companion classes still requires practical confirmation
* might be other ways for bootstrap classes - subject needs to be studied more deeply?
* I believe that dynamic attachment of agent might be added without removal of support of Java 5

But indeed - removal of support simplifies things.

silviu-burcea:
Oracle is planning to support only the last 3(current + previous 2)
versions of Java. You may want to follow this policy too.

On Fri, May 22, 2015 at 11:48 PM, Evgeny Mandrikov <notifications@github.com
> wrote:

> @marchof <https://github.com/marchof> I'm not against drop of support.
>
> However to prevent misunderstanding for others:
>
>    - we already allow coverage on bootstrap classes
>       - in online instrumentation mode partially due to inability to
>       retransform already loaded classes
>       - and I suppose that completely in offline mode?
>     - ability to use "appendToBootstrapClassLoaderSearch" for companion
>    classes still requires practical confirmation
>    - might be other ways for bootstrap classes - subject needs to be
>    studied more deeply?
>    - I believe that dynamic attachment of agent might be added without
>    removal of support of Java 5
>
> But indeed - removal of support simplifies things.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jacoco/jacoco/issues/314#issuecomment-104763758>.
>



-- 
- OCPJP7 (90%)
- OCAJP7 (93%)
- Java and Big Data Enthusiast

marchof:
[...]

PR #821, opened 2019-01-07
Require at least Maven 3.0 for jacoco-maven-plugin

### LINKED ISSUE #194: Revoke support for Maven 2.2.*

slachiewicz:
Fixes #194

Tests with Java 5 failed due to artifact doxia-site-renderer-1.7.4 compiled for Java 6 (from maven-reporting-impl:jar:3.0.0) - blocked by #314

### DIFF
--- a/jacoco-maven-plugin.test/it/it-check-fails-halt/pom.xml
+++ b/jacoco-maven-plugin.test/it/it-check-fails-halt/pom.xml
@@ -40,15 +40,13 @@
             </goals>
             <configuration>
               <rules>
-                <!-- implementation is needed only for Maven 2 -->
-                <rule implementation="org.jacoco.maven.RuleConfiguration">
+                <rule>
                   <element>CLASS</element>
                   <includes>
                     <include>Example</include>
                   </includes>
                   <limits>
-                    <!-- implementation is needed only for Maven 2 -->
-                    <limit implementation="org.jacoco.report.check.Limit">
+                    <limit>
                       <counter>METHOD</counter>
                       <value>MISSEDCOUNT</value>
                       <maximum>0</maximum>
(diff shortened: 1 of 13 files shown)


Godin:
@slachiewicz thank you for your contribution - it is good to know about connection between #194 and #314 👍 However #314 was not in our short-term plans and #194 is definitely not enough to justify #314 - Maven is not the only way of JaCoCo usage. Is there a chance that you can work on a way to resolve this without #314 ? If not, then we would prefer to close this incomplete PR.

slachiewicz:
Unfortunately, I don't see any reasonable solution, because maven-reporting-impl is required for reporting functionality. 
@khmarbaise maybe You have an idea?

Godin:
In fact sounds strange that #194 requires #314 , because according to page https://maven.apache.org/docs/history.html Maven 3.0 was still on Java 5 😉 @slachiewicz are you sure that update of `maven-reporting-impl` is needed?

Godin:
@marchof are you ok if in upcoming release we drop support of Maven 2.2.1 (last release in November 2009) ? If so, then I'll take care of further testing, review of documentation, etc to finalize this PR.

Godin:
@marchof I updated documentation and IMO this is ready to be merged.

PR #829, opened 2019-01-11
Agent can inject class into Java 9+ bootstrap class loader

Godin:
Starting from JDK 9 there is [`java.lang.invoke.MethodHandles.Lookup.defineClass`](https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/MethodHandles.Lookup.html#defineClass-byte:A-)
that can define class even in bootstrap classloader

```
MethodHandles
  .privateLookupIn(Object.class, MethodHandles.lookup())
  .defineClass(classBytes);
```

when `java.base` module opens `java.lang` package to module of agent,
what can be done by agent itself using [`java.lang.instrument.Instrumentation.redefineModule`](https://docs.oracle.com/javase/9/docs/api/java/lang/instrument/Instrumentation.html#redefineModule-java.lang.Module-java.util.Set-java.util.Map-java.util.Map-java.util.Set-java.util.Map-)

```
instrumentation.redefineModule(
  Object.class.getModule(),
  Collections.emptySet(), // extraReads
  Collections.emptyMap(), // extraExports
  Collections.singletonMap(
    "java.lang",
    Collections.singleton(Agent.class.getModule())
  ), // extraOpens
  Collections.emptySet(), // extraUses
  Collections.emptyMap() // extraProvides
);
```

This way agent can be used with JDKs compiled into bytecode version whose instrumentation is not yet supported, e.g. [JDK 13 EA (now each new JDK version is compiled into a new bytecode version)](https://bugs.openjdk.java.net/browse/JDK-8205626). And I believe that majority of people do not compile their code into new bytecode version when testing new JDK EA builds, so this will be already useful for them.

For this use-case such approach is IMO
* definitely safer than automation of [replacement of class](https://github.com/jacoco/jacoco/pull/629#issuecomment-351872653)
* and even than usage of ASM with experimental support of new bytecode version, which is as fragile as our previous tricks with downgrade-upgrade.

Furthermore this approach allows to get rid of [conflicts with other agents](https://github.com/jacoco/jacoco/pull/555) on JDK 9+.

I wrapped reflective calls of these methods into few helper classes to keep compatibility with JDK 5.

### DIFF
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -64,6 +64,15 @@ public final class PreMain {
 		return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
 	}
 
+	/**
+	 * Opens {@code java.base} module for {@link InjectedClassRuntime} when
+	 * executed on Java 9 JREs or higher.
+	 *
+	 * @return <code>true</code> when running on Java 9 or higher,
+	 *         <code>false</code> otherwise
+	 * @throws Exception
+	 *             if unable to open
+	 */
 	private static boolean redefineJavaBaseModule(
 			final Instrumentation instrumentation) throws Exception {
 		try {
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -12,6 +12,9 @@
 package org.jacoco.agent.rt.internal;
 
 import java.lang.instrument.Instrumentation;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
 
 import org.jacoco.core.runtime.AgentOptions;
 import org.jacoco.core.runtime.ClassInjectionRuntime;
@@ -54,12 +57,48 @@ public final class PreMain {
 	private static IRuntime createRuntime(final Instrumentation inst)
 			throws Exception {
 
-		final IRuntime runtime = ClassInjectionRuntime.create(inst);
-		if (runtime != null) {
-			return runtime;
+		if (redefineJavaBaseModule(inst)) {
+			return new ClassInjectionRuntime(Object.class, "$JaCoCo");
 		}
 
 		return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
+	}
+
+	private static boolean redefineJavaBaseModule(
+			final Instrumentation instrumentation) throws Exception {
+		try {
+			Class.forName("java.lang.Module");
+		} catch (final ClassNotFoundException e) {
+			return false;
+		}
+
+		Instrumentation.class.getMethod("redefineModule", //
+				Class.forName("java.lang.Module"), //
+				Set.class, //
+				Map.class, //
+				Map.class, //
+				Set.class, //
+				Map.class //
+		).invoke(instrumentation, // instance
+				getModule(Object.class), // module
+				Collections.emptySet(), // extraReads
+				Collections.emptyMap(), // extraExports
+				Collections.singletonMap("java.lang",
+						Collections.singleton(
+								getModule(ClassInjectionRuntime.class))), // extraOpens
+				Collections.emptySet(), // extraUses
+				Collections.emptyMap() // extraProvides
+		);
+		return true;
+	}
+
+	/**
+	 * @return {@code cls.getModule()}
+	 */
+	private static Object getModule(final Class<?> cls) throws Exception {
+		return Class.class //
+				.getMethod("getModule") //
+				.invoke(cls);
 	}
 
 }
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -13,6 +13,7 @@ package org.jacoco.agent.rt.internal;
 
 import java.lang.instrument.Instrumentation;
 
+import org.jacoco.core.internal.runtime.ClassInjectionRuntime;
 import org.jacoco.core.runtime.AgentOptions;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.ModifiedSystemClassRuntime;
@@ -52,6 +53,12 @@ public final class PreMain {
 
 	private static IRuntime createRuntime(final Instrumentation inst)
 			throws Exception {
+
+		final IRuntime runtime = ClassInjectionRuntime.create(inst);
+		if (runtime != null) {
+			return runtime;
+		}
+
 		return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
 	}
 
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -13,8 +13,8 @@ package org.jacoco.agent.rt.internal;
 
 import java.lang.instrument.Instrumentation;
 
-import org.jacoco.core.internal.runtime.ClassInjectionRuntime;
 import org.jacoco.core.runtime.AgentOptions;
+import org.jacoco.core.runtime.ClassInjectionRuntime;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.ModifiedSystemClassRuntime;
 
--- a/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
+++ b/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
@@ -17,8 +17,8 @@ import java.util.Map;
 import java.util.Set;
 
 import org.jacoco.core.runtime.AgentOptions;
-import org.jacoco.core.runtime.ClassInjectionRuntime;
 import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.InjectedClassRuntime;
 import org.jacoco.core.runtime.ModifiedSystemClassRuntime;
 
 /**
@@ -58,7 +58,7 @@ public final class PreMain {
 			throws Exception {
 
 		if (redefineJavaBaseModule(inst)) {
-			return new ClassInjectionRuntime(Object.class, "$JaCoCo");
+			return new InjectedClassRuntime(Object.class, "$JaCoCo");
 		}
 
 		return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
@@ -85,7 +85,7 @@ public final class PreMain {
 				Collections.emptyMap(), // extraExports
 				Collections.singletonMap("java.lang",
 						Collections.singleton(
-								getModule(ClassInjectionRuntime.class))), // extraOpens
+								getModule(InjectedClassRuntime.class))), // extraOpens
 				Collections.emptySet(), // extraUses
 				Collections.emptyMap() // extraProvides
 		);
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/runtime/ClassInjectionRuntimeTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/runtime/ClassInjectionRuntimeTest.java
@@ -1,0 +1,83 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2018 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.runtime;
+
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.lang.reflect.InvocationTargetException;
+
+import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.RuntimeTestBase;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.internal.AssumptionViolatedException;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Opcodes;
+
+/**
+ * Unit test for {@link ClassInjectionRuntime}.
+ */
+public class ClassInjectionRuntimeTest extends RuntimeTestBase {
+
+	private static IRuntime runtime;
+
+	@BeforeClass
+	public static void setupClass() throws Exception {
+		try {
+			defineJava9Class();
+		} catch (final UnsupportedClassVersionError e) {
+			throw new AssumptionViolatedException(
+					"this test requires at least Java 9");
+		}
+
+		runtime = new ClassInjectionRuntime(ClassInjectionRuntimeTest.class);
+	}
+
+	@Override
+	public IRuntime createRuntime() {
+		return runtime;
+	}
+
+	@Test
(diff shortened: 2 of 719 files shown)


marchof:
@Godin I would prefer to implement a separate `IRuntime` for this approach and also move it to `org.jacoco.core`. Looks cleaner to me. This would require some new API like `isApplicable()`.

What I don't like about re-using `ModifiedSystemClassRuntime` is that there is no actual modification, it just works as the newly injected class looks like a modified class, i.e. has a `$jacocoAccess` field.

Godin:
@marchof implemented `ClassInjectionRuntime`

marchof:
@Godin As a first step I directly fixed Eclipse warnings, hope this is ok for you. Will have a closer look now.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+		 *
+		 * @param bytes
+		 *            the class bytes
+		 * @return class
+		 */
+		Class<?> defineClass(final byte[] bytes) throws Exception {
+			return (Class<?>) Class //
+					.forName("java.lang.invoke.MethodHandles$Lookup")
+					.getMethod("defineClass", byte[].class)
+					.invoke(this.instance, (Object) bytes);
```
marchof: @Godin When editing this file Eclipse removed the `(Object)` cast automatically. Oops.

Not sure whether the primitive array will be autoboxed here or if will be a single parameter even without the cast. In the first case a local variable of type Object should be safe.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.runtime;
```
marchof: Why is this implementation in a private package while the other runtime implementations are in `org.jacoco.core.runtime`? Would move this to the same package.

org.jacoco.core.test/src/org/jacoco/core/runtime/RuntimeTestBase.java
```diff
@@ -38,7 +38,7 @@
 
 	private TestStorage storage;
 
-	abstract IRuntime createRuntime();
+	public abstract IRuntime createRuntime();
```
marchof: No need to make this public when new runtime is in same package (see my other comment).

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+	public static IRuntime create(final Instrumentation instrumentation) {
+		try {
+			redefineJavaBaseModule(instrumentation);
+			return new ClassInjectionRuntime(Object.class);
+		} catch (final Exception e) {
+			return null;
+		}
+	}
+
+	ClassInjectionRuntime(final Class<?> cls) throws Exception {
```
marchof: At a first look I was a bit confused about the purpose of the `cls` parameter. Can we give it a more meaningful name (e.g. `lookupContext`?) and add some Javadoc?

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+	/**
+	 * Creates new instance of runtime.
+	 *
+	 * @param instrumentation
+	 *            instrumentation interface
+	 * @return new runtime instance or <code>null</code> if unable to create
+	 */
+	public static IRuntime create(final Instrumentation instrumentation) {
+		try {
+			redefineJavaBaseModule(instrumentation);
```
marchof: What is the best error handling strategy here?

A) Just check whether we're on Java 9 and then fail if this runtime can't be initialized?
B) Ignore any error while initialization and simply fall back to `ModifiedSystemRuntime`?

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+			return new ClassInjectionRuntime(Object.class);
+		} catch (final Exception e) {
+			return null;
+		}
+	}
+
+	ClassInjectionRuntime(final Class<?> cls) throws Exception {
+		this.injectedClassName = cls.getName().replace('.', '/') + "$JaCoCo";
+		final Class<?> injectedClass = Lookup //
```
marchof: Depending on the error handling strategy this might be moved to the `startup()` method.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+		try {
+			redefineJavaBaseModule(instrumentation);
+			return new ClassInjectionRuntime(Object.class);
+		} catch (final Exception e) {
+			return null;
+		}
+	}
+
+	ClassInjectionRuntime(final Class<?> cls) throws Exception {
+		this.injectedClassName = cls.getName().replace('.', '/') + "$JaCoCo";
```
marchof: I wonder whether a random class name would be more robust e.g. in case of multiple JaCoCo agents?

org.jacoco.core.test/src/org/jacoco/core/internal/runtime/ClassInjectionRuntimeTest.java
```diff
@@ -0,0 +1,83 @@
...
+public class ClassInjectionRuntimeTest extends RuntimeTestBase {
+
+	private static IRuntime runtime;
+
+	@BeforeClass
+	public static void setupClass() throws Exception {
+		try {
+			defineJava9Class();
+		} catch (final UnsupportedClassVersionError e) {
+			throw new AssumptionViolatedException(
```
marchof: Nice, I wasn't aware of this exception up to now 😸

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: @marchof to be honest I don't like that existing implementations of `IRuntime` are in public package - IMO they are implementation details of agent. But ok - can move it.
BTW `org.jacoco.agent.rt.internal.Offline` is in internal package 😉

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: > case of multiple JaCoCo agents

@marchof in #781 we agreed that case of multiple agents is completely misleading and error-prone, so IMO not an argument for random class name. Other things being equal, I prefer predictable behavior over random.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: > add some Javadoc?

@marchof and so maybe you already have draft of it in mind? :wink:

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
marchof: Another option would be to move all runtimes to internal packages of the agent. On the other hand the current public API allows to use runtimes outside the agent (like in the `CoreTutorial` example).

For now I would move it next to the other implementations and we can discuss this in a separate PR, ok?

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
marchof: Ok!

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
marchof: /**
	 * Creates a new runtime which will inject a class in the module and package
	 * of the given class.
	 * 
	 * @param lookupContext
	 *            class to identify the target module and package
	 */

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,297 @@
...
+
+import org.jacoco.core.runtime.AbstractRuntime;
+import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.RuntimeData;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+/**
+ * {@link IRuntime} that uses introduced in Java 9
```
marchof: Would prefer to describe what the class does: "`IRuntime` which injects a new class into the `java.base` module for JaCoCo runtime access using Java 9 APIs."

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: > Not sure whether the primitive array will be autoboxed here or if will be a single parameter even without the cast.

behavior is well defined - `byte[] bytes` is passed as `new Object[] { bytes }`, however confusing as some might wrongly think that each element will be boxed 😉 

> Eclipse removed the (Object) cast automatically

and exactly because of this confusion IntelliJ proposes to do opposite to Eclipse and add explicit cast to `Object` 😆

marchof: Damn. Will a local variable be a fix point for both IDEs? 🤔 

			final Object arg = bytes;
			return (Class<?>) Class //
					.forName("java.lang.invoke.MethodHandles$Lookup")
					.getMethod("defineClass", byte[].class)
					.invoke(this.instance, arg);

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: @marchof argh, then tests will be throwing

```
java.lang.reflect.InvocationTargetException
Caused by: java.lang.LinkageError: loader (instance of jdk/internal/loader/ClassLoaders$AppClassLoader): attempted duplicate class definition for name: "org/jacoco/core/runtime/ClassInjectionRuntimeTest$JaCoCo"
```

😞 

Currently they don't because `createRuntime` returns `static` instance.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: > for JaCoCo runtime access

is IMO redundant, because this is what all `IRuntime` implementations do.
Kept the rest.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: @marchof nothing about the modules in [java.lang.invoke.MethodHandles.Lookup.defineClass](https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/MethodHandles.Lookup.html#defineClass-byte:A-) :

> Defines a class to the same class loader and in the same runtime package and protection domain as this lookup's lookup class.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: Changed implementation from second to first.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: Done.

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: Fix point is

```
.invoke(this.instance, new Object[] { bytes });
```

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
Godin: @marchof solved - tests use name of method as name of class

org.jacoco.core/src/org/jacoco/core/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,308 @@
...
+	 *   Collections.emptyMap())
+	 * }
+	 * </pre>
+	 */
+	private static void redefineJavaBaseModule(
+			final Instrumentation instrumentation) throws Exception {
+		Lookup.lookup() //
+				.bind( //
+						instrumentation, //
+						"redefineModule", //
```
Godin: @marchof doh, not only reflection, but even this doesn't work in case of `--illegal-access=deny` 😢

Godin: In fact

```
instrumentation.getClass().getMethod(...).invoke(...)
```

and

```
Lookup.lookup().bind(instrumentation, ...).invokeWithArguments(...)
```

are illegal, whereas

```
java.lang.Instrumentation.class.getMethod(...).invoke(...)
```

and

```
Lookup.lookup().findVirtual(java.lang.Instrumentation.class, ...).invokeWithArguments(...)
```

are legal, because `instrumentation.getClass()` is `sun.instrument.InstrumentationImpl` and

```
module java.instrument {
    exports java.lang.instrument;

    // allow java launcher to load agents in executable JAR files
    exports sun.instrument to java.base;
}
```

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
marchof: Ok!

org.jacoco.core/src/org/jacoco/core/internal/runtime/ClassInjectionRuntime.java
marchof: Ok.

org.jacoco.core/src/org/jacoco/core/runtime/ClassInjectionRuntime.java
```diff
@@ -0,0 +1,256 @@
...
+	 */
+	public static IRuntime create(final Instrumentation instrumentation) {
+		try {
+			Class.forName("java.lang.Module");
+		} catch (final ClassNotFoundException e) {
+			return null;
+		}
+		try {
+			redefineJavaBaseModule(instrumentation);
+		} catch (final Exception e) {
```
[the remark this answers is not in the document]
Godin: Done

org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
```diff
@@ -52,7 +56,49 @@ public static void premain(final String options, final Instrumentation inst)
...
 			throws Exception {
+
+		if (redefineJavaBaseModule(inst)) {
+			return new InjectedClassRuntime(Object.class, "$JaCoCo");
+		}
+
 		return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
 	}
 
+	private static boolean redefineJavaBaseModule(
```
marchof: Maybe add JavaDoc here about Java 9 dependency:

    /**
     * Opens the java.base module for the InjectedClassRuntime when executed on Java 9 JREs or higher.
     *
     * @return <code>true</true> when we're running on Java 9 or higher, otherwise <code>false</code>
     */

org.jacoco.core/src/org/jacoco/core/runtime/InjectedClassRuntime.java
```diff
@@ -0,0 +1,140 @@
...
+				.set(null, data);
+	}
+
+	public void shutdown() {
+		// nothing to do
+	}
+
+	public int generateDataAccessor(final long classid, final String classname,
+			final int probecount, final MethodVisitor mv) {
+		mv.visitFieldInsn(Opcodes.GETSTATIC, className, FIELD_NAME, FIELD_TYPE);
```
marchof: `classname` vs. `className` is tricky. I would prefer a more distinct naming, e.g. rename member `className` to `injectedClassName`.

org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
Godin: @marchof done

org.jacoco.core/src/org/jacoco/core/runtime/InjectedClassRuntime.java
Godin: @marchof done

PR #833, opened 2019-01-19
Replace empty table in HTML report by messages

Godin:
Using Ant Task and therefore Gradle Plugin, that is based on it, possible to generate report without classes.

For example in case of following `build.xml`

```
<project xmlns:jacoco="antlib:org.jacoco.ant" default="build">
    <classpath path="jacoco/lib/jacocoant.jar"/>
  </taskdef>

  <target name="build">
    <jacoco:report>
      <structure name="example"/>
      <html destdir="report"/>
    </jacoco:report>
  </target>
</project>
```

and `build.gradle`

```
apply plugin: 'base'
apply plugin: 'jacoco'

repositories {
  mavenCentral()
  mavenLocal()
}

task report(type:JacocoReport) {
  executionData.from(files(file('jacoco.exec')))
}
```

execution of

```
$ ant
Buildfile: /private/tmp/example/build.xml

build:
[jacoco:report] Writing bundle 'example' with 0 classes

BUILD SUCCESSFUL
Total time: 0 seconds

$ ant -silent
Buildfile: /private/tmp/example/build.xml

$ gradle-5.0 clean report

BUILD SUCCESSFUL in 1s
2 actionable tasks: 1 executed, 1 up-to-date
```

will produce

empty

Note the absence of messages in the second and third cases.

Also unfortunately many users don't provide screenshots and the way they describe such report is ambiguous: "empty report" and "zero(s)" - in both cases some indeed refer to such report, some other refer to report with classes where covered counters are zero.

Could also be noted that produced `index.html` claims to be XHTML 1.0 Strict whereas this is not the case because of `<tbody/>`.

So I propose to replace empty table by messages for two cases thanks to #817 

```java
if (bundle.getPackages().isEmpty()) {
  ... // No class files specified.
} else if (bundle.isEmpty()) {
  ... // None of the analyzed classes contain code relevant for code coverage.
} else {
  ...
}
```

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -51,6 +51,9 @@
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/818">#818</a>).</li>
   <li>HTML report shows message when analyzed class does not match executed
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/819">#819</a>).</li>
+  <li>HTML report shows message when no class files specified and when all
+      analyzed classes are abstract or generated code
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/833">#833</a>).</li>
   <li>Empty class and sourcefile nodes are preserved and available in XML report
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/817">#817</a>).</li>
 </ul>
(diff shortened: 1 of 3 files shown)


marchof:
@Godin I'm not so sure about this message:

> No classes to display - all analyzed classes are abstract or generated code.

Strictly speaking abstract classes will at least have a constructor (which might only be filtered if it is empty and private) also we have other filters also for non generated code. So I'm not sure whether this is correct right now or might become wrong when adding more filters in future. Therefore I would provide a less specific message, like

> None of the analyzed classes contain code relevant for code coverage.

Godin:
@marchof I was looking for neutral message and your proposal looks great!

marchof:
@Godin Just to make sure: same applies for change log entry.

PR #835, opened 2019-01-22
Add experimental support for Java 13 class files

Godin:


org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -227,4 +228,24 @@ public static void push(final MethodVisitor mv, final int value) {
 		}
 	}
 
+	/**
+	 * Creates {@link ClassReader}.
```
marchof: We should explain the purpose here:

    Creates a {@link ClassReader} instance even when the given class definition is not yet supported by ASM.

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -227,4 +228,24 @@ public static void push(final MethodVisitor mv, final int value) {
...
+	 *
+	 * @param b
+	 *            bytes of class
+	 * @return {@link ClassReader}
+	 */
+	public static ClassReader classReaderFor(final byte[] b) {
+		final byte[] originalVersion = new byte[] { b[4], b[5], b[6], b[7] };
+		if (getVersionMajor(b) == Opcodes.V12 + 1) {
```
marchof: Add constants for this and use `NEXT_VERSION` also in other places instead of `Opcodes.V12 + 1`?

    public static final int MAX_ASM_VERSION = Opcodes.V12;
    public static final int NEXT_VERSION = MAX_ASM_VERSION + 1;

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: Done

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: @marchof not counting `ContentTypeDetector` all other occurrences of `V12 + 1` are in tests, where I would prefer to keep explicit `V12 + 1`.

marchof: Sure, for tests explicit `V12 + 1` seems better. About production code this was just an idea that came in my mind to avoid "magic numbers".

Godin:
@marchof for the record here is end-to-end test:

```
$ mkdir src

$ cat <<END > src/Example.java
class Example {
    System.out.println("Hello, Java 13!");
  }
}
END

$ java -version
openjdk version "13-ea" 2019-09-17
OpenJDK Runtime Environment (build 13-ea+4)
OpenJDK 64-Bit Server VM (build 13-ea+4, mixed mode, sharing)

$ javac --release 13 src/Example.java -d classes

$ javap -v -p classes/Example.class | grep major
  major version: 57

$ java -javaagent:jacoco/lib/jacocoagent.jar -cp classes Example
Hello, Java 13!

$ java -jar jacoco/lib/jacococli.jar classinfo classes/Example.class
     7      0      3      2      2   class 0x6ed8641f71bae73b Example

$ java -jar jacoco/lib/jacococli.jar instrument classes --dest instrumented
[INFO] 1 classes instrumented to /tmp/example/instrumented.

$ javap -v -p instrumented/Example.class | grep major
  major version: 57

$ java -cp instrumented:jacoco/lib/jacocoagent.jar Example
Hello, Java 13!

$ java -jar jacoco/lib/jacococli.jar execinfo jacoco.exec
[INFO] Loading exec file jacoco.exec.
CLASS ID         HITS/PROBES   CLASS NAME
Session "godin-laptop2-7ab9aa52": Wed Jan 23 01:47:30 CET 2019 - Wed Jan 23 01:47:31 CET 2019
6ed8641f71bae73b    1 of   2   Example
Session "godin-laptop2-26fcc118": Wed Jan 23 01:48:43 CET 2019 - Wed Jan 23 01:48:43 CET 2019
6ed8641f71bae73b    1 of   2   Example

$ java -jar jacoco/lib/jacococli.jar report jacoco.exec --classfiles classes 
[INFO] Loading execution data file /tmp/example/jacoco.exec.
[INFO] Analyzing 1 classes.
```

PR #836, opened 2019-01-22
Use default values for zeros

marchof:
Alternative approach to #813 reducing the XML report size. The DTD has been modified to define default value `0` for all counter values. Also XML reports now embed the DTD. This means XML parsers will still get `0` for now omitted counter attributes.

Total reduction of file size is about 24%:

* XML report size `master`: 930.772 bytes
* XML report size `xml_default_values`: 710.345 bytes

### DIFF
--- a/org.jacoco.report.test/src/org/jacoco/report/internal/xml/ReportElementTest.java
+++ b/org.jacoco.report.test/src/org/jacoco/report/internal/xml/ReportElementTest.java
@@ -1,0 +1,73 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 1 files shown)


Godin:
@marchof  Unfortunately integrations such as SonarQube that [ignore DTD](https://github.com/SonarSource/sonar-jacoco/blob/1.0.1.143/src/main/java/org/sonar/plugins/jacoco/XmlReportParser.java#L44) and [do not expect absence of attributes](https://github.com/SonarSource/sonar-jacoco/blob/1.0.1.143/src/main/java/org/sonar/plugins/jacoco/XmlReportParser.java#L108-L115) anyway should be updated 😞

Godin:
For the record I created https://jira.sonarsource.com/browse/JACOCO-5

Godin:
> So we're loosing a version identifier

@marchof good point

> add a `reportversion` and `jacocoversion` attribute to the root element. WDYT?

what do you think about redundancy here: on one side `jacocoversion` has benefit of showing exact version and `reportversion` can be inferred from it, on the other less strightforward to infer as format supposed to change less frequently ?

marchof:
@Godin I think at least the report version is mandatory: So integrators can test compatibility with it. Otherwise integrators would need to update their mapping (JaCoCo version -> report version) with every JaCoCo release.

Issue #623, opened 2017-11-20
FR: Gitlab coverage report


enexusde:
As described in https://medium.com/@kaiwinter/javafx-and-code-coverage-on-gitlab-ci-29c690e03fd6
The percentage of coverage can only be reported to gitlab by print the report to the console.

### Steps to reproduce

JaCoCo version: Latest
Operating system: All
Tool integration: Maven, Gitlab, Github

### Expected behaviour

Parse build-out 

### Actual behaviour

Print report to build-out and parse build-out.

Godin:
Duplicate of #488

enexusde:
I disagree. The coverage is the result of a test-report! All test results in maven are outputed to the console.

Godin:
@enexusde as of today to our taste and compared with other tasks, addition of printing of coverage percentage in console raises far more questions (one for example in case of GitLab is multi-module projects) with all the consequences than benefits. Especially that such information already can be easily extracted from reports, which are btw far more actionable than plain total value, and also since the use case (as it is described) is quite specific to GitLab.

Thank you for your understanding.

PR #840, opened 2019-02-06
This commit is for showing instruction coverage report on maven log

kgcorner:
It introduces two properties
1. showCoverageOnLog (Boolean) which indecates whether jacoco will print coverage on maven log or not, default is false
2. coverageText (String) Text along side which total code coverage will be shown, default value is "Total Code Coverage"

If showCoverageOnLog is true then jacoco will also print below values for Instruction coverage
- Covered count
- Missed Count
- Missed Covered Ratio
- Total count

To utilize this feature, user has to define property `showCoverageOnLog` set to `true` in order to enable JaCoCo to print coverage information on maven log and `coverageText` set to some informative text to show total coverage.


Godin:
@kgcorner sorry, but each new feature requires detailed explanation of its usefulness (what is absent here), from past discussions (#488, #623) we don't see value in such feature

Issue #834, opened 2019-01-19
ArrayIndexOutOfBoundsException


chanjarster:
### Steps to reproduce

JaCoCo version: 0.8.2
Operating system: "mac os x", version: "10.14.2", arch: "x86_64", family: "mac"
Tool integration: Maven 3.5.4
Java version: java version "1.8.0_181"

1. Have docker installed.
2. Use the reproducible code: https://github.com/chanjarster/jacoco-array-index-out-of-bounds-bug
3. Maven command: `mvn clean install`

### Expected behaviour

Everything goes fine.

### Actual behaviour

Maven result:

```
[ERROR] Failures:
[ERROR] me.chanjar.jacocobug.system.repo.DictionaryTypeRepositoryIT.testDictionaryTypeSort(me.chanjar.jacocobug.system.repo.DictionaryTypeRepositoryIT)
[INFO]   Run 1: PASS
[ERROR]   Run 2: DictionaryTypeRepositoryIT>AbstractTestNGSpringContextTests.run:180->testDictionaryTypeSort:31 » ArrayIndexOutOfBounds
```

Failsafe report:

```txt
testDictionaryTypeSort (me.chanjar.jacocobug.system.domain.DictionaryType@4057fa5f)
java.lang.ArrayIndexOutOfBoundsException: 23
	at me.chanjar.jacocobug.system.domain.DictionaryType.setIsDataCenter(DictionaryType.java:100)
	at me.chanjar.jacocobug.system.repo.DictionaryTypeRepository.createDictionaryType(DictionaryTypeRepository.java:42)
	at me.chanjar.jacocobug.system.repo.DictionaryTypeRepository$$FastClassBySpringCGLIB$$9d92e7db.invoke(<generated>)
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204)
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:746)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:139)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
[...]

Issue #816, opened 2019-01-05
Maven test fails to run moco server when using jacoco maven plugin 0.7.8 or newer


chenhengjie123:
This is a issue tracker. Please use our mailing list for general questions: 
https://groups.google.com/forum/?fromgroups=#!forum/jacoco

Also check FAQ before opening an issue: http://www.jacoco.org/jacoco/trunk/doc/faq.html

### Steps to reproduce

test case code:
```
import com.github.dreamhead.moco.Moco;
import com.github.dreamhead.moco.junit.MocoJunitRunner;
import org.apache.http.HttpResponse;
import org.apache.http.client.fluent.Request;
import org.junit.Rule;
import org.junit.Test;
import org.junit.runner.RunWith;

import java.io.IOException;

public class BzdWithdrawHisServiceImplTest {
    @Rule
    public MocoJunitRunner runner = MocoJunitRunner.jsonHttpRunner(12306, Moco.pathResource("foo.json"));

    @Test
    public void  mockTest() throws IOException {
        HttpResponse httpResponse = Request.Get("http://localhost:12306").execute().returnResponse();
        System.out.println(httpResponse);
    }
}
```
configuration in pom.xml: 

```
...
        <!-- https://mvnrepository.com/artifact/com.github.dreamhead/moco-junit -->
        <dependency>
            <groupId>com.github.dreamhead</groupId>
            <artifactId>moco-junit</artifactId>
            <version>0.12.0</version>
        </dependency>

        <!-- https://mvnrepository.com/artifact/com.github.dreamhead/moco-runner -->
        <dependency>
            <groupId>com.github.dreamhead</groupId>
            <artifactId>moco-runner</artifactId>
            <version>0.12.0</version>
        </dependency>
...
            <plugin>
                <groupId>org.jacoco</groupId>
                <artifactId>jacoco-maven-plugin</artifactId>
                <version>${jacoco.maven.plugin.version}</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>prepare-agent</goal>
                        </goals>
                    </execution>
                    <execution>
                        <id>report</id>
                        <phase>prepare-package</phase>
                        <goals>
                            <goal>report</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

error occurs when running `mvn test` to run tests with jacoco.

JaCoCo version: multiple, mentioned in Actual behaviour
Operating system: occurs on mac os 10.14 and windows 7
[...]

Issue #799, opened 2018-12-16
ArrayIndexOutOfBoundsException by runnig unit tests with jacoco plugin


OnkelBums:
### Steps to reproduce

JaCoCo version: 0.8.2
Operating system: Win10
Tool integration: Maven/Ant

### Expected behaviour
The report generation should be a success

### Actual behaviour
BUILD FAILED (the jacoco report generation fails)

A had some DTOs used in my test classes. This DTOs inherit from a BasisDTO class.

When I execute my tests I get the following error log:

> [ERROR] Tests run: 4, Failures: 0, Errors: 4, Skipped: 0, Time elapsed: 1.878 s <<< FAILURE! - in de.dbh.acfs.auftragsverwaltung.mapper.AuftragsContainerEditDTOMapperTest
> [ERROR] mapFromEditToEntityDTOCaseLadungstraegerChanged  Time elapsed: 1.772 s  <<< ERROR!
> java.lang.ArrayIndexOutOfBoundsException: 412
> 	at de.dbh.acfs.auftragsverwaltung.mapper.AuftragsContainerEditDTOMapperTest.mapFromEditToEntityDTOCaseLadungstraegerChanged(AuftragsContainerEditDTOMapperTest.java:41)
> 
> [ERROR] mapEntityToEditDTOCaseSammelladungsIsNull  Time elapsed: 0.017 s  <<< ERROR!
> uk.co.jemos.podam.exceptions.PodamMockeryException
> 	at de.dbh.acfs.auftragsverwaltung.mapper.AuftragsContainerEditDTOMapperTest.mapEntityToEditDTOCaseSammelladungsIsNull(AuftragsContainerEditDTOMapperTest.java:93)
> Caused by: java.lang.reflect.InvocationTargetException
> 	at de.dbh.acfs.auftragsverwaltung.mapper.AuftragsContainerEditDTOMapperTest.mapEntityToEditDTOCaseSammelladungsIsNull(AuftragsContainerEditDTOMapperTest.java:93)
> Caused by: java.lang.ArrayIndexOutOfBoundsException: 789
> 	at de.dbh.acfs.auftragsverwaltung.mapper.AuftragsContainerEditDTOMapperTest.mapEntityToEditDTOCaseSammelladungsIsNull(AuftragsContainerEditDTOMapperTest.java:93)
> 
> [ERROR] mapFromEditToEntityDTOCaseLadungstraegerNotChanged  Time elapsed: 0.034 s  <<< ERROR!
[...]

PR #845, opened 2019-02-15
Use condy for probes array in Java 11+ class files

Godin:
Users often forget to ignore **synthetic** fields when using reflection and continue asking questions even if this is already described in [our FAQ](https://github.com/jacoco/jacoco/blob/v0.8.3/org.jacoco.doc/docroot/doc/faq.html#L130-L141) - here is some recent ones:

* #834
* #816
* #799
* https://groups.google.com/d/msg/jacoco/3wT1WmrRPBA/A9MPMa6tHAAJ

We can get rid of this field in Java 11+ class files by using condy ([JEP 309](https://openjdk.java.net/jeps/309)).

Also I think that condy might be better in terms of performance than `ClassFieldProbeArrayStrategy` in following cases

* cold non JITed code and I suppose that this is quite common case when running unit tests
* when JIT fails to inline `$jacocoInit`

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
@@ -12,6 +12,7 @@
 package org.jacoco.core.instr;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.objectweb.asm.Opcodes.ACC_PUBLIC;
 import static org.objectweb.asm.Opcodes.ACC_SUPER;
 import static org.objectweb.asm.Opcodes.ALOAD;
@@ -150,8 +151,9 @@ public class ClassFileVersionsTest {
 							public void visitEnd() {
 								if (CondyProbeArrayStrategy.B_DESC
 										.equals(desc)) {
-									assertEquals(Boolean.FALSE,
-											Boolean.valueOf(frames));
+									assertFalse(
+											"CondyProbeArrayStrategy does not need frames",
+											frames);
 								} else {
 									assertEquals(Boolean.valueOf(expected),
 											Boolean.valueOf(frames));
--- a/org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
@@ -37,6 +37,7 @@ import static org.objectweb.asm.Opcodes.V9;
 
 import java.io.IOException;
 
+import org.jacoco.core.internal.instr.CondyProbeArrayStrategy;
 import org.jacoco.core.internal.instr.InstrSupport;
 import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.SystemPropertiesRuntime;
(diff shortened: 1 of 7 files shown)


org.jacoco.core/src/org/jacoco/core/internal/instr/CondyProbeArrayStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+		this.accessorGenerator = accessorGenerator;
+	}
+
+	public int storeInstance(final MethodVisitor mv, final boolean clinit,
+			final int variable) {
+		final Handle bootstrapMethod = new Handle(Opcodes.H_INVOKESTATIC,
+				className, InstrSupport.INITMETHOD_NAME, B_DESC, isInterface);
+		mv.visitLdcInsn(new ConstantDynamic(InstrSupport.DATAFIELD_NAME,
+				"Ljava/lang/Object;", bootstrapMethod));
+		mv.visitTypeInsn(Opcodes.CHECKCAST, "[Z");
```
marchof: Why is CHECKCAST required here? I think I know, but maybe we should also let the reader know ;)

Godin: @marchof I moved comment about JDK-8216970 from test into main code

org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
```diff
@@ -147,8 +148,14 @@ public void visitFrame(int type, int nLocal,
 
 							@Override
 							public void visitEnd() {
-								assertEquals(Boolean.valueOf(expected),
-										Boolean.valueOf(frames));
+								if (CondyProbeArrayStrategy.B_DESC
+										.equals(desc)) {
+									assertEquals(Boolean.FALSE,
```
marchof: Use `assertFalse()`

marchof: Add comment:

    // CondyProbeArrayStrategy does not need frames:

org.jacoco.core.test/src/org/jacoco/core/internal/instr/CondyProbeArrayStrategyTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.ClassNode;
+import org.objectweb.asm.tree.LdcInsnNode;
+import org.objectweb.asm.tree.MethodNode;
+import org.objectweb.asm.tree.TypeInsnNode;
+import org.objectweb.asm.tree.VarInsnNode;
+
+public class CondyProbeArrayStrategyTest {
+
+	private final CondyProbeArrayStrategy strategy = new CondyProbeArrayStrategy(
```
marchof: Move initialization to `setup()` method for consistency with other tests.

Godin: done

org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
Godin: > Use `assertFalse()`

done

> Add comment

added as assertion message

PR #846, opened 2019-02-17
Ability to use service loader (spi) to load a list of additional filters

szhem:
This pull requests proposes to add an ability to load additional filters by means of java's SPI mechanism aka ServiceLoaders to easily extend a list of available filters by just dropping a jar into the plugin's classpath.

It could help to increase the amount of contributors to the jacoco's filters in case of (and even if) jacoco maintainers are not experts in the particular technology or library. Later on after the careful review the corresponding external filters can be included into the main distribution.

Candidates of externally available filters:
- [scala ones](https://groups.google.com/forum/#!topic/jacoco/npl9SUL8_WA)
- [aspectj ones ](https://github.com/jacoco/jacoco/pull/710)

Even jacoco itself can be considered to be a subject of using SPI mechanism to load its own filters (e.g. to improve its own modularity).

To include additional external filters into a build classpath it will be necessary to declare a jacoco dependency line the following:

**Maven**
```xml
<plugin>
  <groupId>org.jacoco</groupId>
  <version>${project.version}</version>
  <dependencies>
    <dependency>
      <groupId>foo.bar.filters</groupId>
      <artifactId>external-filters</artifactId>
      <version>${external-filters.version}</version>
    </dependency>
  </dependencies>
</plugin>
```
**Gradle**
```groovy
dependencies {
  jacocoAnt [
    "org.jacoco:org.jacoco.core:${versions.jacoco}",
    "org.jacoco:org.jacoco.ant:${versions.jacoco}",
    "org.jacoco:org.jacoco.report:${versions.jacoco}",
    "foo.bar.filters:external-filters:${external-filters.version}"
  ]
}
```

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -141,7 +141,7 @@
     <asm.version>7.0</asm.version>
     <ant.version>1.7.1</ant.version>
     <args4j.version>2.0.28</args4j.version>
-    <junit.version>4.8.2</junit.version>
+    <junit.version>4.12</junit.version>
 
     <!-- ================== -->
     <!-- For SonarQube analysis -->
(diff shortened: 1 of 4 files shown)


marchof:
Hi @szhem, there is a couple of issues why we don't want to merge this PR:

1) It is not compatible with Java 5, which is our current base line.
2) We don't have a public SPI, the current internal API can be change at any time
3) We don't think we can handle support for people using 3rd party filters if something gongs wrong.

PR #847, opened 2019-02-17
Ability to use service loader (spi) to load a list of additional filters using java5-way

szhem:
This pull request proposes the same thing as #846 does, but using java5-way of loading additional filters just in case its (java5) support is still necessary.

--- 
#### Description below is copied from the #846
This pull requests proposes to add an ability to load additional filters by means of java's SPI mechanism aka ServiceLoaders to easily extend a list of available filters by just dropping a jar into the plugin's classpath.

It could help to increase the amount of contributors to the jacoco's filters in case of (and even if) jacoco maintainers are not experts in the particular technology or library. Later on after the careful review the corresponding external filters can be included into the main distribution.

Candidates of externally available filters:
- [scala ones](https://groups.google.com/forum/#!topic/jacoco/npl9SUL8_WA)
- [aspectj ones ](https://github.com/jacoco/jacoco/pull/710)

Even jacoco itself can be considered to be a subject of using SPI mechanism to load its own filters (e.g. to improve its own modularity).

To include additional external filters into a build classpath it will be necessary to declare a jacoco dependency line the following:

**Maven**
```xml
<plugin>
  <groupId>org.jacoco</groupId>
  <version>${project.version}</version>
  <dependencies>
    <dependency>
      <groupId>foo.bar.filters</groupId>
      <artifactId>external-filters</artifactId>
      <version>${external-filters.version}</version>
    </dependency>
  </dependencies>
</plugin>
```
**Gradle**
```groovy
dependencies {
  jacocoAnt [
    "org.jacoco:org.jacoco.core:${versions.jacoco}",
    "org.jacoco:org.jacoco.ant:${versions.jacoco}",
    "org.jacoco:org.jacoco.report:${versions.jacoco}",
    "foo.bar.filters:external-filters:${external-filters.version}"
  ]
}
```

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/ServiceLoaderFilterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/ServiceLoaderFilterTest.java
@@ -1,0 +1,259 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Sergey Zhemzhitsky - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.analysis.filter;
+
+import java.io.File;
+import java.io.FileOutputStream;
+
+import org.jacoco.core.internal.instr.InstrSupport;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.tree.MethodNode;
+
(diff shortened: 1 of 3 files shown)


marchof:
Same problem as with PR #846:

* We don't have a public SPI, the current internal API can be change at any time
* We don't think we can handle support for people using 3rd party filters if something gongs wrong.

szhem:
@marchof  This feature could be declared as experimental one and a subject to change at any time not to support people who uses it at their own risk. 
Don't you think that it can decrease a number of support requests instead? People can use it to implement their own filters without necessity to open support tickets and feature-requests which are difficult to handle by you team due to lack of time or any other reason.

marchof:
@szhem There are two kind of users: One that actually can work with our code base and can develop extension (we know maybe max 3). The other group (probably >> 100'000) which only uses JaCoCo as is. If the first group publishes a extension which is used by the second group who do you think will get the all the support requests when we actually change the API? We see this almost daily with questions coming from people who actually use 3rd partly integration.

The first group still can fork JaCoCo for their own purposes and use our integration test framework to develop well tested extensions e.g. for Scala. If they actually work and come with proper tests they can be merged into upstream.

In theory I agree with you that libraries should be extensible from an architecture point of view. The practice of maintaining coverage tools for more than a decade teached me that this adds enourmous complexity and load to our support channels.

PR #851, opened 2019-03-04
Upgrade ASM to 7.1

### LINKED ISSUE #839: Failure after instrumentation of code which has stackmap frames and array of more than 7 dimensions
I am in the process of migrating a SW from Java 8 to Java 11. We are using JaCoCo for code coverage and I have found a problem when running the unit tests with JaCoCo tool (using ant). 

JaCoCo fails when it instantiates a class that contains a primitive array with more than 8 dimensions. Please do not ask about the motivation of having an 8 dimensional array .... legacy code I have to maintain.

I have reduced the code to the min that reproduces the problem. 

- The execution goes fine when I do not use JaCoCo/EclEmma or I use Java 8
- I have tested with two different JaCoCo versions , 0.8.2 and 0.8.3

### Steps to reproduce

**JaCoCo version**: tested with 0.8.2 and 0.8.3
**Operating system**: MacOS HighSierra 10.13.6 and CentOS Linux release 7.5.1804 
**Tool integration**: Ant 1.10.5 and Eclipse EclEmma 3.1.1.201809121651
**Java Versions**: OpenJDK 11.0.1 and OpenJDK 11.0.2

Use the following piece of code

```
import org.junit.Test;
public class PrimitiveArrayJacocoTest {

	/**
	 * Test with a boolean array
	 */
	@Test
	public void testBooleanArray() {
		// Just instantiate the class
		new BooleanArrayJacoco();		
	}
	
	/**
	 * Test with an integer array
	 */
	@Test
	public void testIntegerArray() {	
		// Just instantiate the class
		new IntegerArrayJacoco();	
	}
	
	public class BooleanArrayJacoco {
		
		public BooleanArrayJacoco() {
		    
			// Declare a 8 dimension boolean array , if the amount of dimensions is reduced test does not fail. 
		    boolean[][][][][][][][] array = new boolean[12][][][][][][][];
		    int idxDepth0 = 0; 
		    if(array[idxDepth0] == null) {
		    	array[0] = new boolean[4][][][][][][]; // Create new node with 4 array pointers
		    }
		}
		
	}
	
	public class IntegerArrayJacoco {
		
		public IntegerArrayJacoco() {
		    
			// Declare a 8 dimension integer array , if the amount of dimensions is reduced test does not fail. 
		    int[][][][][][][][] array = new int[12][][][][][][][];
		    int idxDepth0 = 0; 
		    if(array[idxDepth0] == null) {
		    	array[0] = new int[4][][][][][][]; // Create new node with 4 array pointers
		    }
		}
		
	}

}
```

### Expected behaviour

JaCoCo is able to run the tests with no failure.

### Actual behaviour

Tests fail with the following stack trace

```
java.lang.NoClassDefFoundError: I
 at gaia.cu1.tools.util.lookuptree.PrimitiveArrayJacocoTest.testIntegerArray(PrimitiveArrayJacocoTest.java:44)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
 at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
 at java.base/java.lang.reflect.Method.invoke(Method.java:566)
 at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
 at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
 at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
 at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
 at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
 at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
 at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
 at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
 at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
 at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
 at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
 at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
 at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:538)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:760)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:460)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:206)
Caused by: java.lang.ClassNotFoundException: I
 at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
 at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
 at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
 ... 24 more
```

and running the testBooleanArray .... the same but with different ClassNotFoundException:

```
java.lang.NoClassDefFoundError: Z
 at gaia.cu1.tools.util.lookuptree.PrimitiveArrayJacocoTest.testBooleanArray(PrimitiveArrayJacocoTest.java:33)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
 at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
 at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
 at java.base/java.lang.reflect.Method.invoke(Method.java:566)
 at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
 at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
 at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
 at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
 at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
 at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
 at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
 at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
 at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
 at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
 at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
 at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
 at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
 at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:538)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:760)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:460)
 at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:206)
Caused by: java.lang.ClassNotFoundException: Z
 at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
 at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
 at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
 ... 24 more
```

Godin:
Fixes #839

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -169,7 +169,7 @@ public final class InstrSupport {
 	 * @see #setMajorVersion(int, byte[])
 	 * @see #getMajorVersion(ClassReader)
 	 */
-	static int getMajorVersion(final byte[] b) {
+	public static int getMajorVersion(final byte[] b) {
 		return ((b[6] & 0xFF) << 8) | (b[7] & 0xFF);
 	}
 
@@ -183,13 +183,13 @@ public final class InstrSupport {
 	 *            bytes of class
 	 * @see #getMajorVersion(byte[])
 	 */
-	static void setMajorVersion(final int majorVersion, final byte[] b) {
+	public static void setMajorVersion(final int majorVersion, final byte[] b) {
 		b[6] = (byte) (majorVersion >>> 8);
 		b[7] = (byte) majorVersion;
 	}
 
 	/**
-	 * Gets major of bytecode version number from given {@link ClassReader}.
+	 * Gets major version number from given {@link ClassReader}.
 	 *
 	 * @param reader
 	 *            reader to get information about the class
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -158,18 +158,16 @@ public final class InstrSupport {
 	 */
 	static final int CLINIT_ACC = Opcodes.ACC_SYNTHETIC | Opcodes.ACC_STATIC;
 
-	private static final int MAJOR_VERSION_INDEX = 6;
-
 	/**
-	 * Gets major of bytecode version number from given bytes of class.
+	 * Gets major of bytecode version number from given {@link ClassReader}.
 	 *
-	 * @param b
-	 *            bytes of class
-	 * @return version of bytecode
+	 * @param reader
+	 *            reader to get information about the class
+	 * @return major version of bytecode
 	 */
-	public static int getVersionMajor(final byte[] b) {
-		return (short) (((b[MAJOR_VERSION_INDEX] & 0xFF) << 8)
-				| (b[MAJOR_VERSION_INDEX + 1] & 0xFF));
+	public static int getVersionMajor(final ClassReader reader) {
+		final int firstConstantPoolEntryOffset = reader.getItem(1) - 1;
+		return reader.readShort(firstConstantPoolEntryOffset - 4);
 	}
 
 	/**
@@ -238,7 +236,7 @@ public final class InstrSupport {
 	 */
 	public static ClassReader classReaderFor(final byte[] b) {
 		final byte[] originalVersion = new byte[] { b[4], b[5], b[6], b[7] };
-		if (getVersionMajor(b) == Opcodes.V12 + 1) {
+		if (b[6] == 0x00 && b[7] == 0x39) {
 			b[4] = (byte) (Opcodes.V12 >>> 24);
 			b[5] = (byte) (Opcodes.V12 >>> 16);
 			b[6] = (byte) (Opcodes.V12 >>> 8);
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -194,8 +194,12 @@ public final class InstrSupport {
 	 * @param reader
 	 *            reader to get information about the class
 	 * @return major version of bytecode
+	 * @see ClassReader#ClassReader(byte[], int, int)
+	 * @see #getMajorVersion(byte[])
 	 */
 	public static int getMajorVersion(final ClassReader reader) {
+		// relative to the beginning of constant pool because ASM provides API
+		// to construct ClassReader which reads from the middle of array
 		final int firstConstantPoolEntryOffset = reader.getItem(1) - 1;
 		return reader.readUnsignedShort(firstConstantPoolEntryOffset - 4);
 	}
@@ -267,6 +271,7 @@ public final class InstrSupport {
 	public static ClassReader classReaderFor(final byte[] b) {
 		final int originalVersion = getMajorVersion(b);
 		if (originalVersion == Opcodes.V12 + 1) {
+			// temporarily downgrade version to bypass check in ASM
 			setMajorVersion(Opcodes.V12, b);
 		}
 		final ClassReader classReader = new ClassReader(b);
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -167,7 +167,7 @@ public final class InstrSupport {
 	 */
 	public static int getVersionMajor(final ClassReader reader) {
 		final int firstConstantPoolEntryOffset = reader.getItem(1) - 1;
-		return reader.readShort(firstConstantPoolEntryOffset - 4);
+		return reader.readUnsignedShort(firstConstantPoolEntryOffset - 4);
 	}
 
 	/**
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -159,6 +159,36 @@ public final class InstrSupport {
 	static final int CLINIT_ACC = Opcodes.ACC_SYNTHETIC | Opcodes.ACC_STATIC;
 
 	/**
+	 * Gets major version number from given bytes of class (unsigned two bytes
+	 * at offset 6).
+	 *
+	 * @return major version of bytecode
+	 * @see <a href=
+	 *      "https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-4.html#jvms-4.1">Java
+	 *      Virtual Machine Specification §4 The class File Format</a>
+	 * @see #setMajorVersion(int, byte[])
+	 * @see #getVersionMajor(ClassReader)
+	 */
+	static int getMajorVersion(final byte[] b) {
+		return ((b[6] & 0xFF) << 8) | (b[7] & 0xFF);
+	}
+
+	/**
+	 * Sets major version number in given bytes of class (unsigned two bytes at
+	 * offset 6).
+	 *
+	 * @param majorVersion
+	 *            major version of bytecode to set
+	 * @param b
+	 *            bytes of class
+	 * @see #getMajorVersion(byte[])
+	 */
+	static void setMajorVersion(final int majorVersion, final byte[] b) {
+		b[6] = (byte) (majorVersion >>> 8);
+		b[7] = (byte) majorVersion;
+	}
+
+	/**
 	 * Gets major of bytecode version number from given {@link ClassReader}.
 	 *
 	 * @param reader
@@ -235,15 +265,12 @@ public final class InstrSupport {
 	 * @return {@link ClassReader}
 	 */
 	public static ClassReader classReaderFor(final byte[] b) {
-		final byte[] originalVersion = new byte[] { b[4], b[5], b[6], b[7] };
-		if (b[6] == 0x00 && b[7] == 0x39) {
-			b[4] = (byte) (Opcodes.V12 >>> 24);
-			b[5] = (byte) (Opcodes.V12 >>> 16);
-			b[6] = (byte) (Opcodes.V12 >>> 8);
-			b[7] = (byte) Opcodes.V12;
+		final int originalVersion = getMajorVersion(b);
+		if (originalVersion == Opcodes.V12 + 1) {
+			setMajorVersion(Opcodes.V12, b);
 		}
 		final ClassReader classReader = new ClassReader(b);
-		System.arraycopy(originalVersion, 0, b, 4, originalVersion.length);
+		setMajorVersion(originalVersion, b);
 		return classReader;
 	}
 
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -162,6 +162,8 @@ public final class InstrSupport {
 	 * Gets major version number from given bytes of class (unsigned two bytes
 	 * at offset 6).
 	 *
+	 * @param b
+	 *            bytes of class
 	 * @return major version of bytecode
 	 * @see <a href=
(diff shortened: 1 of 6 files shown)


Godin:
@marchof this is because you picked PR which is still in "implementation" 😆 btw hard to predict url for changelog before creation of PR 😉

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -158,18 +158,16 @@ private InstrSupport() {
...
-	 * @return version of bytecode
+	 * @param reader
+	 *            reader to get information about the class
+	 * @return major version of bytecode
 	 */
-	public static int getVersionMajor(final byte[] b) {
-		return (short) (((b[MAJOR_VERSION_INDEX] & 0xFF) << 8)
-				| (b[MAJOR_VERSION_INDEX + 1] & 0xFF));
+	public static int getVersionMajor(final ClassReader reader) {
+		final int firstConstantPoolEntryOffset = reader.getItem(1) - 1;
```
marchof: Why are you going relative to the contant pool entry? Isn't it always at the same absolute position? I'm asking because we have another method `classReaderFor()` in the same class which uses a fixed lookup index. Should be consistent.

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -238,7 +236,7 @@ public static void push(final MethodVisitor mv, final int value) {
 	 */
 	public static ClassReader classReaderFor(final byte[] b) {
 		final byte[] originalVersion = new byte[] { b[4], b[5], b[6], b[7] };
-		if (getVersionMajor(b) == Opcodes.V12 + 1) {
+		if (b[6] == 0x00 && b[7] == 0x39) {
```
marchof: Magic number? The intend was more clear with `Opcodes.V12 + 1`.

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: @marchof recall #850 - there is constructor [`ClassReader(byte[] classFileBuffer, int classFileOffset, int classFileLength)`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L165-177).

Godin: @marchof so relative to the constant pool entry seems safer to me, even if currently our code does not pass such instances of `ClassReader` into this method.

Godin: I can update unit test to show/cover this case.

Or we can create subclass of `ClassReader` that will returned by `classReaderFor` and required here.

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: @marchof inlining of former implementation of `getVersionMajor` into this method results in

```java
final int majorVersion = (short) (((b[MAJOR_VERSION_INDEX] & 0xFF) << 8)
		| (b[MAJOR_VERSION_INDEX + 1] & 0xFF));

if (majorVersion == Opcodes.V12 + 1)
```

which IMO contains far more magic and far less readable than

```
if (b[6] == 0x00 && b[7] == 0x39)
```

Can also add comment.

Readable and non-magical alternative

```
final int majorVersion = java.nio.ByteBuffer.wrap(b)
		.order(java.nio.ByteOrder.BIG_ENDIAN).getShort(6);
```

looks like a bit of overkill.

Godin: BTW numbers `6`, `7` and `0x39` are even shown in https://en.wikipedia.org/wiki/Java_class_file 😉

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
marchof: @Godin ASM API is really wired here: it has public `readX()` methods but keeps the offset private. My proposals to resolve this:

* Use absolute positions as the offset constructor cannot be used any more due to #850 or
* Add a inline comment to explain why it is done this way

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
marchof: So why inlining? If we keep the previous method (in addition to the new Reader version) we have testable and readable code ;)

marchof:
@Godin If you like we can merge as is and check in a separate PR how we can streamline version reading/writing. Ok?

marchof:
@Godin What about requesting a new public API `ClassReader.getVersion()` from the ASM project?

marchof:
@Godin With explicit methods to read versions the code becomes way more understandable:

	public static ClassReader classReaderFor(final byte[] b) {
		final int originalVersion = getVersion(b);
		if (getVersionMajor(b) == Opcodes.V12 + 1) {
			// temporary downgrade version for ASM
			setVersion(Opcodes.V12, b);
		}
		final ClassReader classReader = new ClassReader(b);
		setVersion(originalVersion, b);
		return classReader;
	}

Godin:
@marchof 

> What about requesting a new public API ClassReader.getVersion() from the ASM project?

Feel free to ask them :wink: :innocent:

----

Speaking about understandability, testability and magic :crystal_ball: - actually implementation of `getVersionMajor(byte[])` in `0.8.3`

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java#L170-L173

is incorrect, as well as

```
final int majorVersion = java.nio.ByteBuffer.wrap(b)
		.order(java.nio.ByteOrder.BIG_ENDIAN).getShort(6);
```

because [according to JVMS](https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-4.html#jvms-4.1) `major_version` is **unsigned** :wink:

For example let's consider `major_version` that with current temps (2 per year) will be approximately in 16393 years from now :smiling_imp: - `32786` (`0x8012`) and thus following test

```
final byte[] bytes = new byte[] { //
		(byte) 0xCA, (byte) 0xFE, (byte) 0xBA, (byte) 0xBE, // magic
		0x00, 0x00, // minor_version
		(byte) 0x80, (byte) 0x12 // major_version
};
assertEquals(0x8000, InstrSupport.getVersionMajor(bytes));
```

which fails

```
Expected :32786
Actual   :-32750
```

IMO such mistake is impossible in case of proposed exact comparison of bytes

```
if (b[6] == 0x80 && b[7] == 0x12)
```

:wink:

However proposed implementation of `getVersionMajor(ClassReader)`

https://github.com/jacoco/jacoco/blob/c7e2176bef48044cbf6dfd425f84a2dd012128fa/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java#L168-L171

has the same mistake because of [`readShort` which reads signed value](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L3366-3376).

Both come from the fact that [ASM itself uses same signed `readShort` to read `major_version`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L193-196) :laughing: :rofl: This also means that instead of "downgrade" trick, we can actually do "upgrade" trick :rofl: :rofl:

----

Moreover `major_version` is **2 bytes** and so implementation of `needsFrames(int)`

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java#L182-L185

is also incorrect and will fail / require change earlier - in about 122 years :smiling_imp:

```
assertEquals(true, InstrSupport.needsFrames(0x0100));
```

Morever similarly to https://github.com/jacoco/jacoco/pull/851#discussion_r262829574 about `getVersionMajor(ClassReader)`, `needsFrames` currently always receives `major_version` and so doesn't need to "consider 1.1 anomaly" ? :wink:

----

I added correct implementations of `getMajorVersion(byte[])` and `setMajorVersion(int, byte[])`,
fixed, renamed and commented `getMajorVersion(ClassReader)`. And let's fix `needsFrames` separately.

marchof:
@Godin Thanks for fixing this right away!

> will fail / require change earlier - in about 122 years

Ok, we document this as know limitation -- along with > 32 dimensional arrays 🤣

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -158,18 +158,50 @@ private InstrSupport() {
...
+	}
+
+	/**
+	 * Gets major of bytecode version number from given {@link ClassReader}.
+	 *
+	 * @param reader
+	 *            reader to get information about the class
+	 * @return major version of bytecode
+	 * @see ClassReader#ClassReader(byte[], int, int)
+	 * @see #getMajorVersion(byte[])
```
marchof: This give me a warning because public API links to package private API (which is not part of JavaDoc). Remove link or make methods public?

Godin: Done

marchof: @Godin Now as it is public we have a new warning as parameter `b` is not documented 🙈

Godin: Fixed

PR #850, opened 2019-03-04
Remove `instrument(ClassReader)` and `analyzeClass(ClassReader)`

Godin:
There are several problems with methods [`Instrumenter.instrument(ClassReader)`](https://github.com/jacoco/jacoco/blob/v0.8.3/org.jacoco.core/src/org/jacoco/core/instr/Instrumenter.java#L72-L82) and [`Analyzer.analyzeClass(ClassReader)`](https://github.com/jacoco/jacoco/blob/v0.8.3/org.jacoco.core/src/org/jacoco/core/analysis/Analyzer.java#L104-L112) :

* their implementations use
[field `org.objectweb.asm.ClassReader.b` which was marked as deprecated in ASM 7.1](https://gitlab.ow2.org/asm/asm/commit/5ed1a4abbbbd4ba416a5f3d99031535e89a2e88f)

* they don't work in case when field `ClassReader.b` contains more than just bytes of one class (see [`ClassReader(byte[] classFileBuffer, int classFileOffset, int classFileLength)`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L165-177)) and seems that there is no easy way to fix this

So I think that we have only two options:
1. mark them as deprecated and remove later
2. remove them

@marchof WDYT?

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/perf/ExecuteInstrumentedCodeScenario.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/perf/ExecuteInstrumentedCodeScenario.java
@@ -18,7 +18,6 @@ import org.jacoco.core.runtime.IRuntime;
 import org.jacoco.core.runtime.LoggerRuntime;
 import org.jacoco.core.runtime.RuntimeData;
 import org.jacoco.core.test.TargetLoader;
-import org.objectweb.asm.ClassReader;
 
 /**
  * This scenario runs a given scenario twice and reports the execution time:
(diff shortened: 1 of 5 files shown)


Godin:
For fun and future readers of this thread:
* to read bytes of class from `ClassReader` one can use [`ClassReader.readByte(int offset)`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L3343-3352)
* to obtain `classFileOffset` one can use [`ClassReader.getItem(int constantPoolEntryIndex)`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L3318-3330), i.e. `classFileOffset = getItem(0) - 11`
* however to determine `classFileLength` ([which is even not used](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L175)) one need [logic to skip `fields`, `methods` and `attributes`](https://gitlab.ow2.org/asm/asm/blob/ASM_7_1/asm/src/main/java/org/objectweb/asm/ClassReader.java#L3186-3231)

PR #853, opened 2019-03-08
Fix unlikely yet theoretically possible NPE

Godin:


marchof:
What about adding an new API to `AbstractMatcher`?

    void nextIsLdc(Object value)

Can be unit tested and makes filters more readable.

Alternatively

    Object nextIsLdc()

which returns the value. This method would also work for the other use cases.

Godin:
> `void nextIsLdc(Object value)`

Unfortunately there won't be code-reuse - value is checked only in two places, here and in

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilter.java#L52-L56

where check is `startsWith` instead of `equals`.

So far we've been adding methods to `AbstractMatcher`, when they have two or more consumers, and I'd like prefer to continue this practice.

> `Object nextIsLdc()`

if I understand your proposal correctly, then this method can return null, in this case how this prevents same mistake of accidental dereference without null-check ?

IMO better direction will be `nextIsLdc(Matcher<LdcInsnNode>)` or `nextIsLdc(Matcher<Object>)`
* name `Matcher` picked to not clash with `java.util.Predicate` in JDK 8, also maybe `boolean Predicate.test(T)` interface won't be sufficient for our needs
* then can imagine `nextIs(Matcher<AbstractInsnNode>)` and implementations of `Matcher<AbstractInsnNode>` which check multiple instructions via `getNext`

In any case for clarity I prefer to separate this PR with fix from possible refactorings/consolidations.

Godin:
`nextIs(Matcher<AbstractInsnNode>)` already will be useful in places where different opcodes can be encountered, for example
https://github.com/jacoco/jacoco/blob/23d835d5eb0f62a7ff036c5e74d933f0de407e7c/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java#L80
and
https://github.com/jacoco/jacoco/blob/23d835d5eb0f62a7ff036c5e74d933f0de407e7c/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchJavacFilter.java#L80

marchof:
@Godin Separate PR is ok! Remember we're still in Java 5... We don't have non-capturing lambdas at hand and should avoid creating too many matcher instances. Of course we can add constants for matcher, not sure whether this makes the code more readable.

Godin:
Note to myself: another places where predicate might be useful are

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/AbstractMatcher.java#L118-L128

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesEcjFilter.java#L210-L223

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavacFilter.java#L231-L244

https://github.com/jacoco/jacoco/blob/6cd3f0bd8e348f8fba7bffec5225407151f1cc91/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavac11Filter.java#L120-L133

PR #857, opened 2019-03-13
`nextIsInvokeStatic` and `nextIsInvokeVirtual` should check method descriptor

Godin:


marchof:
@Godin Now we have three methods with almost identical bodies:

* `nextIsInvokeSuper()`
* `nextIsInvokeVirtual()`
* `nextIsInvokeStatic()`

What about delegating from this method to a new internal method?

    private void nextIsInvoke(int opcode, String owner, String name, String descriptor)

Godin:
@marchof Exactly! And this was one of the goals of this change - to make these methods "almost identical" and more importantly similar to `nextIsInvokeSuper` to be able to factor out common part and reduce duplication. Before making next change, I was wondering if you'll notice this :wink:

Instead of

> delegating from this method to a new internal method

I'd like to propose to get rid of intermediate methods and directly call

```
void nextIsInvoke(int opcode, String owner, String name, String descriptor)
```

WDYT?

marchof:
@Godin I see, you laid the bait for me ... is there something else in the PR you want me to distract from? 😆 

Sure, using `nextIsInvoke()` directly is good too.

Godin:
> is there something else in the PR

nope

> Sure, using `nextIsInvoke()` directly is good too.

done

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/AbstractMatcherTest.java
```diff
@@ -137,89 +137,41 @@ public void nextIsVar() {
...
-		matcher.cursor = m.instructions.getFirst();
-		matcher.nextIsInvokeSuper("another_owner", "()V");
+		matcher.nextIsInvoke(Opcodes.INVOKEVIRTUAL, "owner", "another_name",
+				"()V");
 		assertNull(matcher.cursor);
 
 		// should set cursor to null when descriptor mismatch
 		matcher.cursor = m.instructions.getFirst();
-		matcher.nextIsInvokeSuper("owner", "(I)V");
+		matcher.nextIsInvoke(Opcodes.INVOKEVIRTUAL, "owner", "another_name",
```
marchof: Should read `"name"` here, otherwise we not testing the descriptor here and miss a branch.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
```diff
@@ -37,7 +37,8 @@ public void filter(final MethodNode methodNode,
 		private boolean match(final MethodNode methodNode,
 				final String superClassName) {
 			firstIsALoad0(methodNode);
-			nextIsInvokeSuper(superClassName, CONSTRUCTOR_DESC);
+			nextIsInvoke(Opcodes.INVOKESPECIAL, superClassName, "<init>",
```
marchof: We have a contant `CONSTRUCTOR_NAME` in this filter. Either we use it consistently or we remove it at all.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/AbstractMatcherTest.java
Godin: ouch, fixed

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java
Godin: fixed

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/AbstractMatcherTest.java
marchof: Try using "JaCoCo". Pretty cool tool to identify whether tests do what you think they do :stuck_out_tongue_winking_eye:

Godin: Yeah, one day should finally try it 🤣

PR #861, opened 2019-03-21
Add validation test for groovy.lang.Delegate

jwloka:
Added integration test case for groovy support during Hackergarten Zurich.

Test fails, issue reported at https://issues.apache.org/jira/browse/GROOVY-9051

### DIFF
--- a/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyDelegateClassTarget.groovy
+++ b/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyDelegateClassTarget.groovy
@@ -13,14 +13,24 @@ package org.jacoco.core.test.validation.groovy.targets
 
 class GroovyDelegateClassTarget { // assertEmpty()
 
-    @Delegate
-    Date target = new Date() // assertEmpty()
+    static class D {
+        void m1() {
+        } // assertFullyCovered()
 
-    long getTime() {
-        return target.time  // assertFullyCovered()
+        void m2() {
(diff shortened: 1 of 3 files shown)


org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyDelegateClassTarget.groovy
```diff
@@ -0,0 +1,33 @@
...
+* - a suitable implementation of after(Date)
+* - a suitable implementation of compareTo(Date)
+* - a suitable implementation of toInstant()
+* - a suitable implementation of getTarget()
+* - a suitable implementation of setTarget(Date)
+*/
+
+class GroovyDelegateClassTarget { // assertEmpty()
+
+    @groovy.lang.Delegate
```
Godin: According to documentation of `groovy.lang.Delegate`

>  All public instance methods present in the delegate type **and not present in the owner class** will be added to owner class at compile time.

so I think that for completeness of test will be nice to add hand-written method that is present in `java.lang.Date` and verify that compiler won't mark it is as `Generated`.

Godin:
@jwloka thank you for your contribution! and @aalmiray thank you for Hackergarten! ❤️

Godin:
Build fails because
* `gmavenplus-plugin` executes Groovy compiler with JDK 8 used for Maven instead of JDK 7 that is specified via `maven-toolchains-plugin`
* so that Groovy compiler generates delegate for `java.time.Instant java.util.Date#toInstant()` which was added in JDK 8
* test is executed with JDK 7 where `java.time.Instant` is absent

😢

PR #864, opened 2019-03-21
Add validation test for groovy.transform.AutoClone

ollin:
implemented during hackergarten Zurich.

Test fails because of https://issues.apache.org/jira/browse/GROOVY-9051

### DIFF
--- a/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/GroovyAutoCloneClassTest.java
+++ b/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/GroovyAutoCloneClassTest.java
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Andres Almiray - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation.groovy;
+
+import org.jacoco.core.test.validation.ValidationTestBase;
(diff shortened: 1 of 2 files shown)


org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/GroovyAutoCloneClassTest.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Contributors:
+ *    Andres Almiray - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation.groovy;
+
+import org.jacoco.core.test.validation.ValidationTestBase;
+import org.jacoco.core.test.validation.groovy.targets.GroovyAutoCloneClassTarget;
+import org.jacoco.core.test.validation.groovy.targets.GroovyDataClassTarget;
```
Godin: Unused import.

org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyAutoCloneClassTarget.groovy
```diff
@@ -0,0 +1,25 @@
...
+ *******************************************************************************/
+package org.jacoco.core.test.validation.groovy.targets
+
+import groovy.transform.AutoClone
+
+@AutoClone
+class GroovyAutoCloneClassTarget { // assertEmpty()
+
+    String name // assertEmpty()
+    Date age     // assertEmpty()
```
Godin: To me seems that fields are irrelevant for this test and can be removed.

Godin:
@ollin thank you for your contribution! and @aalmiray thank you for Hackergarten! ❤️

org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyAutoCloneClassTarget.groovy
Godin: because `groovy.transform.AutoClone` causes generation of `clone` method even in absence of fields.

aalmiray:
Thank you! :smile:

PR #865, opened 2019-03-21
Add validation test for groovy.transform.builder.Builder

jwloka:
Integration test for groovy support with @Builder

Test fails, issue submitted to https://issues.apache.org/jira/browse/GROOVY-9054


org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovySimpleStrategyBuilderClassTarget.groovy
```diff
@@ -0,0 +1,44 @@
...
+* - a suitable implementation of setter method setFirst(), setLast(), setBorn()
+* - a suitable implementation of getFirst()
+* - a suitable implementation of getLast()
+* - a suitable implementation of getBorn()
+*/
+
+@groovy.transform.builder.Builder
+class GroovySimpleStrategyBuilderClassTarget { // assertEmpty()
+
+    @groovy.transform.builder.Builder(builderStrategy= SimpleStrategy)
```
Godin: for consistency with other tests (e.g. https://github.com/jacoco/jacoco/blob/7c90df8e8b294f07c0bd4c79cbdd5bd5119713af/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyAutoCloneClassTarget.groovy#L16) let's use `import` instead of qualification

Godin: for consistency with other tests (e.g. https://github.com/jacoco/jacoco/blob/7c90df8e8b294f07c0bd4c79cbdd5bd5119713af/org.jacoco.core.test.validation.groovy/src/org/jacoco/core/test/validation/groovy/targets/GroovyAutoCloneClassTarget.groovy#L16) let's use `import` instead of qualification

Godin:
@jwloka thank you for one more contribution! and @aalmiray thank you again for Hackergarten! ❤️

aalmiray:
https://github.com/apache/groovy/pull/929

Godin:
@aalmiray why there is no tests regarding generation of `@Generated` annotation on Groovy side? nor for https://github.com/apache/groovy/pull/929 , nor for previously resolved tickets? both as a confirmation of fix and for prevention of regressions :wink:

aalmiray:
@Godin I think we left out those checks due to time, an oversight I expect to correct in a future Hackergarten or at http://hack-commit-pu.sh/ 😉

Issue #869, opened 2019-04-05
Shouldn't report-aggregate MOJO be an aggregator mojo?


JWGmeligMeyling:
Currently, the `report-aggregate` goal from the `jacoco-maven-plugin` is not specified as an [`aggregator`](https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/org/apache/maven/plugins/annotations/Mojo.html#aggregator--) MOJO. Aggregator executions run once and last, after the submodules have already executed their plugins. Having `aggregator` disabled means that, for multi-module projects this results in the goal being executed for every single module, which is often not what you want. As a work around, examples usually create a separate [`coverage-report` module](https://github.com/jacoco/jacoco/wiki/MavenMultiModule#usage), which is dependent on any other submodule relevant for the results, and only define a plugin execution on that project. This however feels like a workaround to me, and it seems much more logical to me to just declare the `report-aggregate` MOJO as an aggregator and have it only be run once on the entire project anyway.

If we decide to stick with the current behaviour of `report-aggregate`, perhaps we could introduce a new goal `report-aggregator` in a non-breaking fashion. I suppose it would eliminate the need of a separate coverage report module in most cases, and drastically simplify setup.

### Steps to reproduce

JaCoCo version:
Operating system: OSX
Tool integration: Maven

### Actual behaviour

Godin:
Mojos marked as `aggregator` can only be called directly like for example `release:prepare` and `release:perform` ( see https://issues.apache.org/jira/browse/MNG-6336 ), so can't cover all use-cases which `report-aggregate` already covers. Addition of another goal just for direct invocation is counterproductive and misleading.

JWGmeligMeyling:
Ah I see, I actually misread that section of the documentation. I do agree that the current `report-aggregate` goal is more versatile than achievable through aggregator tasks. I do however not agree that having another goal for direct invocation is counterproductive or misleading. It will make the trivial case of a single report for a multi module project a lot easier. Even though the current proces is well documented, there is confusion about it on the internet and it is just tedious to set up the entire module to even get started with multi module coverage reporting.

I've patched the plugin in order to include an aggregator module as I was tidying up my modules anyway. I've opened up a PR with the work I've done, don't feel the urge to integrate it, just wanted to put it out there 😄

Godin:
[...]

PR #872, opened 2019-04-22
Added aggregator mojo

JWGmeligMeyling:
Implements #869

I understand the philosophy behind the ReportAggregate MOJO, but for some use cases a simple Maven aggregator report is sufficient and easier to use, as it does not require a separate coverage report module.

I have implemented an aggregator MOJO and decided to at least open up a PR to see if theres interest in integrating it as an additional type of reporting.

Usage:

In your main module pom:

```xml
<plugin>
    <groupId>org.jacoco</groupId>
    <executions>
        <execution>
            <id>agent-for-ut</id>
            <goals>
                <goal>prepare-agent</goal>
            </goals>
        </execution>
    </executions>
    <configuration>
        <append>true</append>
        <destFile>${maven.multiModuleProjectDirectory}/target/jacoco.exec</destFile>
    </configuration>
</plugin>
```

Then execute with: `mvn test jacoco:report-aggregator`.

It's basically the same as using the existing `report-aggregate` task with a dependency on every reactor module.


Godin:
Closing as per explanations in #869

PR #877, opened 2019-05-11
InstrumentingLoader should use findLoadedClass

Godin:
Because execution of `java.lang.ClassLoader.defineClass` multiple
times with same class name causes `java.lang.LinkageError` with
message "duplicate class definition".


Godin:
@marchof FYI this was discovered during attempt to rework https://github.com/jacoco/jacoco/pull/861#issuecomment-491467461 so decided to merge without pr-review to unblock it 😛 will do adjustments on top, if you have some

marchof:
Looks good to me. Thx for fixing this!

PR #887, opened 2019-05-31
Filter unreachable branches in Kotlin open functions with default arguments

### LINKED ISSUE #882: Default arguments in open Kotlin function cause untestable branches
### Steps to reproduce
Test this class:
````kotlin
open class TestClass {

    open fun testFun(arg1: String, arg2: String = "abc"): String {
        return arg2
    }
}
````
with this junit test code:

````kotlin
class TestClassTest {

    @Test
    fun testFun() {
        val testClass = TestClass()
        assertEquals("abc", testClass.testFun("123"))
        assertEquals("def", testClass.testFun("123", "def"))
        assertEquals("abc", testClass.testFun("123", "abc"))
    }
}
````
and examine the coverage of `TestClass`

### Environment
JaCoCo version: 0.8.4
Operating system: Macos
Tool integration: Gradle

### Expected behaviour
`TestClass` should show complete coverage

### Actual behaviour
The synthetic function `testFun$default(TestClass, String, String, int, Object)` which implements default arguments is shown as missing coverage of one branch and 5 instructions. See image below.

The issue occurs only with open functions. It appears Kotlin  inserts additional code which is presumably invoked only when the function is overridden. The particular use-case which led to this is in a Spring Boot environment where annotations such as `@Transactional` automatically make classes open, rather than the Kotlin default of final.

Screenshot 2019-05-15 17 25 02

### PR DESCRIPTION
**Pull request by Godin**
For open functions with default arguments

```
open class Open {
    open fun f(a: String = "a") {
    }
}
```

Kotlin compiler generates check that default arguments are not used for super calls - https://github.com/JetBrains/kotlin/blob/v1.3.31/compiler/backend/src/org/jetbrains/kotlin/codegen/FunctionCodegen.java#L1341-L1359 , even if currently such calls not allowed - compiler rejects following

```
class Derived : Open() {
    override fun f(a: String) {
        super.f() // Error: Super-calls with default arguments are not allowed. Please specify all arguments of 'super.f' explicitly
    }
}
```

I suppose that check is introduced for future versions of language, where such calls might be allowed.

This PR adds handling of open functions to `KotlinDefaultArgumentsFilter`.

Fixes #882

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
@@ -124,6 +124,17 @@ public abstract class ValidationTestBase {
 	}
 
 	@Test
+	public void all_missed_instructions_should_have_line_number() {
+		CounterImpl c = CounterImpl.COUNTER_0_0;
+		for (Line line : source.getLines()) {
+			c = c.increment(line.getCoverage().getInstructionCounter());
+		}
+		assertEquals(
+				"sum of missed instructions of all lines should be equal to missed instructions of file",
+				source.getCoverage().getInstructionCounter().getMissedCount(), c.getMissedCount());
+	}
+
+	@Test
 	public void all_branches_should_have_line_number() {
 		CounterImpl c = CounterImpl.COUNTER_0_0;
 		for (Line line : source.getLines()) {
--- a/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
@@ -123,6 +123,17 @@ public abstract class ValidationTestBase {
 				source.getCoverage().getLastLine() <= source.getLines().size());
 	}
 
+	@Test
+	public void all_branches_should_have_line_number() {
+		CounterImpl c = CounterImpl.COUNTER_0_0;
+		for (Line line : source.getLines()) {
+			c = c.increment(line.getCoverage().getBranchCounter());
+		}
+		assertEquals(
+				"sum of branch counters of all lines should be equal to branch counter of file",
+				source.getCoverage().getBranchCounter(), c);
+	}
+
 	/*
 	 * Predefined assertion methods:
 	 */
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java [hunk 1]
```diff
@@ -123,6 +123,17 @@ public void last_line_in_coverage_data_should_be_less_or_equal_to_number_of_line
 				source.getCoverage().getLastLine() <= source.getLines().size());
 	}
 
+	@Test
+	public void all_branches_should_have_line_number() {
```
Godin: @marchof here is some notes about this new test:

Unfortunately in our validation tests all assertions are about code that has line numbers, so validation test was green without this - see first commit.

Similar generic assertion for instructions counter is not possible, because it does not hold for KotlinInlineTarget.

marchof: Nice, I like this!

I actually tried with instructions and it fails with most Kotlin tests.

I wonder whether we should add this test to the base class and explicitly overwrite it for those tests where it does not hold true (with comments why this is).

Godin: @marchof 

> I actually tried with instructions and it fails with most Kotlin tests.

You're right - instruction counters won't be equal for most Kotlin targets, because first entry in `LineNumberTable` is past non-null checks of parameters:

```
    Code:
      stack=8, locals=2, args_size=1
         0: aload_0
         1: ldc           #10                 // String args
         3: invokestatic  #16                 // Method kotlin/jvm/internal/Intrinsics.checkParameterIsNotNull:(Ljava/lang/Object;Ljava/lang/String;)V
         6: new           #18                 // class org/jacoco/core/test/validation/kotlin/targets/KotlinDataClassTarget$DataClass
         ...
      LineNumberTable:
        line 40: 6
```

And in fact I wasn't precise enough about description of "similar generic assertion":  what holds for every target except `KotlinInlineTarget` - is equality of number of missed instructions with line number and without, i.e. that all missed instructions are visible in source, while covered might be invisible, which is ok since they are covered 😉

> explicitly overwrite it for those tests where it does not hold true (with comments why this is)

Haven't thought about override. Nice idea! 👍 Added it.

PR #888, opened 2019-06-03
Do not filter out synthetic constructors that contain values of default arguments in Kotlin

Godin:
Similarly to #773 constructors in Kotlin can have default arguments with branches, what unfortunately was overlooked during implementation of #774 - currently we filter them out, because they gets compiled differently than functions:

```
class Example() {
    constructor(a: Boolean, b: String = if (a) "a" else "b") : this()
}
```

results in constructor with following signature

```
public synthetic <init>(ZLjava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```

This was discovered during implementation of #887.

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
@@ -48,24 +48,38 @@ public final class KotlinDefaultArgumentsFilter implements IFilter {
 		return methodName.endsWith("$default");
 	}
 
+	static boolean isDefaultArgumentsConstructor(final MethodNode methodNode) {
+		if (!methodNode.name.equals("<init>")) {
+			return false;
+		}
+		final Type[] argumentTypes = Type.getMethodType(methodNode.desc)
+				.getArgumentTypes();
+		if (argumentTypes.length < 2) {
+			return false;
+		}
+		return "kotlin.jvm.internal.DefaultConstructorMarker"
+				.equals(argumentTypes[argumentTypes.length - 1].getClassName());
+	}
+
 	public void filter(final MethodNode methodNode,
 			final IFilterContext context, final IFilterOutput output) {
 		if ((methodNode.access & Opcodes.ACC_SYNTHETIC) == 0) {
-			return;
-		}
-		if (!isDefaultArgumentsMethodName(methodNode.name)) {
 			return;
 		}
 		if (!KotlinGeneratedFilter.isKotlinClass(context)) {
 			return;
 		}
 
-		new Matcher().match(methodNode, output);
+		if (isDefaultArgumentsMethodName(methodNode.name)) {
+			new Matcher().match(methodNode, output, false);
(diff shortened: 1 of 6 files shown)


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -48,24 +48,38 @@ static boolean isDefaultArgumentsMethodName(final String methodName) {
...
 		}
-		if (!isDefaultArgumentsMethodName(methodNode.name)) {
-			return;
-		}
 		if (!KotlinGeneratedFilter.isKotlinClass(context)) {
 			return;
 		}
 
-		new Matcher().match(methodNode, output);
+		if (isDefaultArgumentsMethodName(methodNode.name)) {
```
marchof: For consistency with the new `isDefaultArgumentsConstructor()` method I would rename this method to `isDefaultArgumentsMethod()`. Also passing a `MethodNode` both all callers do `methodNode.name` anyways.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java
```diff
@@ -35,6 +35,11 @@ public void filter(final MethodNode methodNode,
 				return;
 			}
 
+			if (KotlinDefaultArgumentsFilter
```
marchof: Side note: When I see this I more and more like @Godin 's idea to separate filters for different compilers.

Godin: @marchof don't know about whom you're talking 😆

Not sure that separation can remove this dependency between filters, i.e. not every but some `synthetic` methods anyway should be filtered in Kotlin.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
Godin: Done

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -48,24 +48,38 @@ static boolean isDefaultArgumentsMethodName(final String methodName) {
 		return methodName.endsWith("$default");
 	}
 
+	static boolean isDefaultArgumentsConstructor(final MethodNode methodNode) {
+		if (!methodNode.name.equals("<init>")) {
```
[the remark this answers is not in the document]
Godin: I doubt that `methodNode.name` can be null, and `isDefaultArgumentsMethod` anyway performs `methodNode.name.endsWith` without null-check 😉 but flipped for consistency within `isDefaultArgumentsConstructor`.

Godin:
@marchof fixed, also updated javadoc of filter according to changes, please re-review

PR #893, opened 2019-06-07
Add safety slot before variable for probe array

### LINKED ISSUE #767: Probes inserted at locals that will be overridden

Godin:
Fixes #767

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
@@ -62,7 +62,7 @@ class ProbeInserter extends MethodVisitor implements IProbeInserter {
 		for (final Type t : Type.getArgumentTypes(desc)) {
 			pos += t.getSize();
 		}
-		variable = pos;
+		variable = pos + 1;
 	}
 
 	public void insertProbe(final int id) {
@@ -118,14 +118,14 @@ class ProbeInserter extends MethodVisitor implements IProbeInserter {
 		// stack size is an absolute maximum, as the accessor code is inserted
 		// at the very beginning of each method when the stack size is empty.
 		final int increasedStack = Math.max(maxStack + 3, accessorStackSize);
-		mv.visitMaxs(increasedStack, maxLocals + 1);
+		mv.visitMaxs(increasedStack, maxLocals + 2);
 	}
 
 	private int map(final int var) {
-		if (var < variable) {
+		if (var < variable - 1) {
 			return var;
 		} else {
-			return var + 1;
+			return var + 2;
 		}
 	}
 
@@ -138,12 +138,15 @@ class ProbeInserter extends MethodVisitor implements IProbeInserter {
 					"ClassReader.accept() should be called with EXPAND_FRAMES flag");
 		}
 
-		final Object[] newLocal = new Object[Math.max(nLocal, variable) + 1];
+		final Object[] newLocal = new Object[Math.max(nLocal + 2, variable + 1)];
 		int idx = 0; // Arrays index for existing locals
 		int newIdx = 0; // Array index for new locals
 		int pos = 0; // Current variable position
 		while (idx < nLocal || pos <= variable) {
-			if (pos == variable) {
+			if (pos == variable - 1) {
+				newLocal[newIdx++] = Opcodes.TOP;
+				pos++;
+			} else if (pos == variable) {
 				newLocal[newIdx++] = InstrSupport.DATAFIELD_DESC;
 				pos++;
 			} else {
--- a/org.jacoco.core.test/src/org/jacoco/core/instr/MethodParameterOverwriteTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/instr/MethodParameterOverwriteTest.java
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Marc R. Hoffmann - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.instr;
+
+import org.jacoco.core.runtime.IRuntime;
+import org.jacoco.core.runtime.RuntimeData;
+import org.jacoco.core.runtime.SystemPropertiesRuntime;
+import org.jacoco.core.test.TargetLoader;
+import org.junit.Test;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+public class MethodParameterOverwriteTest {
+
+	@Test
+	public void run_original()
+			throws IllegalAccessException, InstantiationException {
+		final byte[] original = generateMethod();
+
+		new TargetLoader().add("Sample", original).newInstance();
+	}
+
+	@Test
+	public void run_instrumented() throws Exception {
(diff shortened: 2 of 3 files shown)


org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
```diff
@@ -62,7 +62,7 @@
 		for (final Type t : Type.getArgumentTypes(desc)) {
 			pos += t.getSize();
 		}
-		variable = pos;
+		variable = pos + 1;
```
marchof: I would add some documentation about the "magic" +1. Either here and maybe in addition in the class documentation to explain the general approach.

Jacks0N23:
Please merge it, it's block me af

AlexeyKorshun:
can you merge this fix?

Godin:
@Jacks0N23 @AlexeyKorshun 

> can you merge it?

unfortunately not yet - see https://github.com/jacoco/jacoco/pull/893#pullrequestreview-1403399584

Godin:
@marchof

> has probably no performance impact

I'm not so sure, are you? 😉 

> This will increase the byte code size

And method size is limited 😉 
 
> simply adding it to the end

I wouldn't call it "simply" - `ProbeInserter#visitMaxs` can not be used to determine position, because it is invoked after other `ProbeInserter` visit methods (such as `ProbeInserter#visitVarInsn`), which already should know position, so we will need to find another place/way to compute position. Which to me sounds more complex than a modification of this PR that will fix https://github.com/jacoco/jacoco/pull/893#discussion_r1178755535, and especially since I already have this modification and will publish it shortly 😉

Godin:
@marchof I rebased this PR and fixed the problem described in https://github.com/jacoco/jacoco/pull/893#discussion_r1178755535, it still lacks in-code comments requested in https://github.com/jacoco/jacoco/pull/893#discussion_r291450313 and the changelog entry, but maybe you already can review what else might be missing in the implementation?

----

Fun fact: the issue can be reproduced by using Kotlin compiler versions starting from released 7 years ago `1.0` 🙈 😆  up to the latest as of today released `1.8.21` - enough to declare an `inline` function with a default value for a parameter and body that immediately declares a variable of `Long` or `Double` type

```
inline fun example(x: Int = 0): Double {
    val y: Double = 0.0
    return x + y
}

}
```

org.jacoco.core.test/src/org/jacoco/core/internal/instr/SafetySlotTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.objectweb.asm.ClassReader;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+public class SafetySlotTest {
+
+	@Test
+	public void original() throws Exception {
```
marchof, line 31: As there is no assertion in the test maybe a descriptive test name can describe the purpose of the test:

```suggestion
	public void should_verify_original_class_without_errors() throws Exception {
```

org.jacoco.core.test/src/org/jacoco/core/internal/instr/SafetySlotTest.java
```diff
@@ -0,0 +1,118 @@
...
+		final byte[] original = createClass();
+
+		new TargetLoader().add("Sample", original).newInstance();
+	}
+
+	@Test
+	public void instrumented() throws Exception {
```
marchof, line 38:
```suggestion
	public void should_verify_instrumented_class_without_errors() throws Exception {
```

marchof (approved):

org.jacoco.core.test/src/org/jacoco/core/internal/instr/SafetySlotTest.java
Godin:
```suggestion
	public void jvm_should_verify_original_class_without_errors() throws Exception {
```

org.jacoco.core.test/src/org/jacoco/core/internal/instr/SafetySlotTest.java
Godin:
```suggestion
	public void jvm_should_verify_instrumented_class_without_errors() throws Exception {
```

Godin:
@marchof I added comments and updated changelog, so IMO this can be merged 🎉

PR #908, opened 2019-07-19
KotlinDefaultArgumentsFilter should not assume that all parameters consume one slot

### LINKED ISSUE #907: Some branches are not covered if the type of a default argument is Long.
### Steps to reproduce

JaCoCo version: 0.8.5-SNAPSHOT
Operating system: Linux pop-os 5.0.0-20-generic
Tool integration: gradle

Test this production code:

```kotlin
data class Foo(val x: Long = 0)
```

by this test:

```kotlin
class FooTest {

    @Test
    fun test() {
        assertEquals(1, Foo(1).x)
        assertEquals(0, Foo().x)
    }
}
```

### Expected behaviour

All instructions and branches are covered

### Actual behaviour

All instructions are covered but 50% of branches are NOT covered.

image

Godin:
Fixes #907

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
@@ -129,12 +129,17 @@ public final class KotlinDefaultArgumentsFilter implements IFilter {
 			}
 		}
 
-		private static int maskVar(final String desc, final boolean constructor) {
+		private static int maskVar(final String desc,
+				final boolean constructor) {
+			int slot = 0;
+			if (constructor) {
+				// one slot for reference to current object
+				slot++;
+			}
 			final Type[] argumentTypes = Type.getMethodType(desc)
 					.getArgumentTypes();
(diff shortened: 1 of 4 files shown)


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -129,6 +128,17 @@ public void match(final MethodNode methodNode,
 				output.ignore(i, i);
 			}
 		}
+
+		private static int maskVar(final String desc, final boolean constructor) {
+			final Type[] argumentTypes = Type.getMethodType(desc)
+					.getArgumentTypes();
+			int slot = 0;
+			for (int i = 0; i < argumentTypes.length
+					- (constructor ? 1 : 2); i++) {
```
marchof: What do the magic numbers here mean?

Maybe we can enlight future readers of the code with a local variable:

    int descripiveName = constructor ? 1 : 2;

Godin: @marchof please re-review

PR #911, opened 2019-07-26
Add validation test for Kotlin control structures

Godin:


Godin:
We have such test for Java, so I think we should add similar for Kotlin too. @marchof could you please review?

marchof:
@Godin Cool!

PR #922, opened 2019-08-18
Add detection of Scala classes

Godin:


marchof:
@Godin Shouldn't this PR add a integration test for Scala that demonstrates the improvement?

Godin:
@marchof There is no improvement for Scala, since all Scala classes supposed to have these attributes 😉Also don't know cases of such synthetic methods in other languages, but would like to have this detection of Scala classes as safeguard and for possible future filters.

PR #923, opened 2019-08-21
add filter: GetterAndSetterFilter

donhui:
add filter: GetterAndSetterFilter

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterContextMock.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterContextMock.java
@@ -11,6 +11,8 @@
  *******************************************************************************/
 package org.jacoco.core.internal.analysis.filter;
 
+import org.objectweb.asm.tree.FieldNode;
+
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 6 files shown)


marchof:
@donhui Thanks for this contribution! Filtering user code is very controversial. Therefore we decided not to include them in JaCoCo. It will simply cause too many discussions with our users which we can't handle with our limited resources. Therefore we currently only focus on filtering compiler generated synthetic code.

Also as you can see from the broken build, getting filters right is a cumbersome task. Your filter breaks an Kotlin integration test.

Thanks for your understanding!

donhui:
I wonder why my filter breaks an Kotlin integration test. 
Because I never change the Kotlin code.

marchof:
Probably because your filter has side effects on filtering of classfiles generated by the Kotlin compiler?

We have lots of integration tests for different JVM languages. Actually there should also be one for your filter.

But as said before: We will curretnly not include user code filtering.

donhui:
> Probably because your filter has side effects on filtering of classfiles generated by the Kotlin compiler?
> 
> We have lots of integration tests for different JVM languages. Actually there should also be one for your filter.
> 
> But as said before: We will currently not include user code filtering.

I see. Thank you！

asolntsev:
We would like to filter out getters/setters because of Kotlin. 
In Kotlin, we have `data` classes which have generated getters/setters. We don't want to test them neither see those "uncovered" lines in JaCoCo report. 

@donhui

donhui:
@asolntsev 
you can reference the code of this PR, then custom your jacoco jar and related plugins.

asolntsev:
@donhui thank you, I know that in open-source world it's always possible to fork&customize :)

But for most of users, it's too hard way.

PR #926, opened 2019-08-24
Fix substitution in `org.jacoco.ant/about.html`

### LINKED ISSUE #924: Keywords in org.jacoco.ant/about.html not expanded
### Steps to reproduce

[Download](https://search.maven.org/remotecontent?filepath=org/jacoco/org.jacoco.ant/0.8.4/org.jacoco.ant-0.8.4.jar) latest JaCoCo release from Maven Central.

### Expected behaviour

File about.html should show build date.

### Actual behaviour

The keyword `@build.date@` is not replaced.

Godin:
Fixes #924

### DIFF
--- a/org.jacoco.doc/docroot/doc/maven.html
+++ b/org.jacoco.doc/docroot/doc/maven.html
@@ -91,7 +91,7 @@
 &lt;plugin&gt;
   &lt;groupId&gt;org.jacoco&lt;/groupId&gt;
   &lt;artifactId&gt;jacoco-maven-plugin&lt;/artifactId&gt;
-  &lt;version&gt;@project.version@&lt;/version&gt;
+  &lt;version&gt;${project.version}&lt;/version&gt;
 &lt;/plugin&gt;
 </pre>
 
@@ -152,8 +152,8 @@ mvn help:describe -Dplugin=org.jacoco:jacoco-maven-plugin -Ddetail
 
 </div>
 <div class="footer">
-  <span class="right"><a href="@jacoco.home.url@">JaCoCo</a> @qualified.bundle.version@</span>
-  <a href="license.html">Copyright</a> &copy; @copyright.years@ Mountainminds GmbH &amp; Co. KG and Contributors
+  <span class="right"><a href="${jacoco.home.url}">JaCoCo</a> ${qualified.bundle.version}</span>
+  <a href="license.html">Copyright</a> &copy; ${copyright.years} Mountainminds GmbH &amp; Co. KG and Contributors
 </div>
 
 </body>
--- a/org.jacoco.agent.rt.test/about.html
+++ b/org.jacoco.agent.rt.test/about.html
@@ -9,7 +9,7 @@
 <h2>About This Content</h2>
 
 <p>
-  @build.date@
+  ${build.date}
 </p>
 
 <h3>License</h3>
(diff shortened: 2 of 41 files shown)


marchof:
Thanks for picking this up quickly! Here is something strange: When I run the build locally, replacement does not work at all (neither in *.html or in *.properties) and finally Ant tests fail:

     [java] Exception in thread "main" java.lang.NoClassDefFoundError: ${jacoco/runtime/package/name}/Offline
     [java] 	at org.jacoco.ant.TestTarget.$jacocoInit(TestTarget.java)
     [java] 	at org.jacoco.ant.TestTarget.main(TestTarget.java)
     [java] Caused by: java.lang.ClassNotFoundException: ${jacoco.runtime.package.name}.Offline
     [java] 	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
     [java] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
     [java] 	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
     [java] 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
     [java] 	... 2 more

My local environment was:

* Java 8u181
* Maven 3.5.3
* mvn clean install

Any proposal what I can do on my side to investigate?

marchof:
In line 94 of `maven.html`there I found one more `@`substitution.

Godin:
> In line 94 of `maven.html` there I found one more `@` substitution.

Fixed and double-checked all remaining occurrences of regex `@[a-z.]++@` and so also fixed some in `org.jacoco.examples/`.

> When I run the build locally, replacement does not work at all
> Any proposal what I can do on my side to investigate?

This is very weird - I just checked with `Maven 3.5.3` and `JDK 8u181` and it works on my side.
So for the time being I don't have ideas. Maybe you can try without Eclipse from command line after cleaning local Maven repository and with fresh clone into new directory

```
rm -rf ~/.m2/repository/org/jacoco/
git clone https://github.com/jacoco/jacoco.git
cd jacoco
mvn install
```

?

marchof:
@Godin Ok, I found the cause: When I build in `./org.jacoco.build`as working directory keyword replacement fails. When repo root `./` is the working directory (and I use `/pom.xml`) Everything is good.

Do you wan't to investigate (and fix) this or do we require workspace root as working directory for build?

Godin:
@marchof weird again, because

```
cd org.jacoco.build
mvn install
```

works for me. So my best wild guess remains the same - are you using Eclipse Maven Launcher or command line?

marchof:
Looks like this some local problem on my side only.

org.jacoco.doc/docroot/doc/maven.html
```diff
@@ -91,7 +91,7 @@ <h2>Usage</h2>
 &lt;plugin&gt;
   &lt;groupId&gt;org.jacoco&lt;/groupId&gt;
   &lt;artifactId&gt;jacoco-maven-plugin&lt;/artifactId&gt;
-  &lt;version&gt;@project.version@&lt;/version&gt;
+  &lt;version&gt;@{project.version}&lt;/version&gt;
```
marchof: Should read: `${project.version}` (dollar)

marchof:
@Godin Ok, looks like running Eclipse kicks-in and copies original resources to target.

org.jacoco.doc/docroot/doc/maven.html
Godin: Ouch. Fixed!

PR #927, opened 2019-08-25
Add validation test for Scala synchronized block

Godin:


Godin:
We have such test for Java, so I think we should add similar for Scala too. @marchof could you please review?

PR #931, opened 2019-08-30
Report invalid coverage ratio configuration

### LINKED ISSUE #783: limit > 1 breaks rule config
### Steps to reproduce
I created a rule 

```
<rule implementation="org.jacoco.maven.RuleConfiguration">
                                    <element>PACKAGE</element>
                                    <limits>
                                        <limit implementation="org.jacoco.report.check.Limit">
                                            <counter>INSTRUCTION</counter>
                                            <value>COVEREDRATIO</value>
                                            <!-- Note: if you put in a value > 1 it will report 0% coverage! I'm reporting this bug to Jacoco. -->
                                            <minimum>90</minimum>
                                        </limit>
                                    </limits>
                                </rule>
```

This blows up saying every package in the project has 0% coverage. When I look at the html reports they still have real looking coverage info. This caused me to waste a lot of time and great frustration. I almost gave up.

JaCoCo version: jacoco-maven-plugin:0.8.2
Operating system: MacOS
Tool integration: Maven

### Expected behaviour
If a value greater than 1 is entered it should either report an invalid limit or just show the package level as being below the impossible value of 1.

### Actual behaviour
Currently blows up saying 0% coverage even though the html report shows the accurate value.

marchof:
Fixes #783 

To avoid common misconfigurations of coverage limits where users provide
ratio as percentage value (without "%") we now issue a rule violation in
such situations.

* Update documentation to make clear that the range for ratios is [0,1]
* Use new naming conventions in corresponding unit tests
* Remove unnecessary config elements from Ant tests
* Fix local variable name


marchof:
While the rules could be validated once I decided to include it in the existing check methods. This leads to repeated execution but on the other hand allows to keep the existin `IViolationOutput` API.

jacoco-maven-plugin/src/org/jacoco/maven/CheckMojo.java
```diff
@@ -161,8 +161,8 @@ private boolean canCheckCoverage() {
 	}
 
 	@Override
-	public void executeMojo() throws MojoExecutionException,
-			MojoExecutionException {
```
Godin: Duplicate in throws clause - [fixed in master](https://github.com/jacoco/jacoco/commit/b71209adbf45b454cdf2ebda6f144f2687430c9e).

Godin:
@marchof we have really nice statement on [Development Conventions page](https://www.jacoco.org/jacoco/trunk/doc/conventions.html)

> Pull requests handling various topics ("I fixed this and that") are typically difficult in handling

So I cherry-picked non-functional changes as separate commits into master - https://github.com/jacoco/jacoco/commit/02682c9fe38f681ad24670c2f7d944712275c8de and https://github.com/jacoco/jacoco/commit/9f1efb2b591b306869a20c83ac60cb92153e23b3 This really helps to focus on functional changes 😉

org.jacoco.doc/docroot/doc/ant.html
```diff
@@ -857,11 +857,11 @@ <h3>Element <code>check</code></h3>
...
-      <td>Expected minimum value. If the minimum refers to a ratio the range is
-          from 0.0 to 1.0 where the number of decimal places will also determine
-          the precision in error messages. A limit ratio may optionally be
-          declared as a percentage where 0.80 and 80% represent the same value,
-          the value must end with %.</td>
+      <td>Expected minimum value. If the minimum refers to a ratio it must be
+          in the range from 0.0 to 1.0 where the number of decimal places will
+	      also determine the precision in error messages. A limit ratio may
+	      optionally be declared as a percentage where 0.80 and 80% represent
+	      the same value.</td>
```
Godin: This and other HTML files use spaces for indentation, so for consistency tab characters should be removed.

org.jacoco.report/src/org/jacoco/report/check/Limit.java
```diff
@@ -80,10 +80,11 @@ public CounterEntity getEntity() {
 
 	/**
 	 * Sets the counter entity to check.
+	 * 
+	 * TODO: use CounterEntity directly once Maven 3 is required.
```
Godin: Not only this Maven-ism affected API, but also this [`TODO`-comment is visible in documentation](https://www.jacoco.org/jacoco/trunk/doc/api/org/jacoco/report/check/Limit.html#setCounter(java.lang.String)) 😞 And while we already require Maven 3 (#821), I propose to just move this `TODO`-comment outside of javadoc.

org.jacoco.report/src/org/jacoco/report/check/Limit.java
```diff
@@ -98,10 +99,11 @@ public CounterValue getValue() {
 
 	/**
 	 * Sets the value to check.
+	 * 
+	 * TODO: use CounterValue directly once Maven 3 is required.
```
Godin: Same as above - let's move this `TODO`-comment outside of javadoc.

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -62,6 +62,14 @@ <h3>Non-functional Changes</h3>
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/910">#910</a>).</li>
 </ul>
 
+<h3>API Changes</h3>
+<ul>
+    <li>The coverage check API and tools (Ant, Maven) now report an error, when
+        a coverage ratio limit is configured outside the range [0,1] to avoid
+        common configuration mistakes.
```
Godin: There is dot is after reference to ticket as like in all existing places, so this one should be removed for consistency.

org.jacoco.report/src/org/jacoco/report/check/Limit.java
```diff
@@ -186,4 +195,29 @@ private String message(final String minmax, final BigDecimal v,
...
+			if (maxmsg != null) {
+				return maxmsg;
+			}
+		}
+		return null;
+	}
+
+	private String checkRatioLimit(final String minmax, final BigDecimal v) {
+		if (v != null && (v.compareTo(BigDecimal.ZERO) == -1
+				|| v.compareTo(BigDecimal.ONE) == 1)) {
```
Godin: While `BigDecimal.compareTo` clearly specifies return values as `-1, 0, or 1`,
`< 0` (less than zero) and `> 0` (greater than 0) look clearer to my taste,
also `Comparable.compareTo` specifies return values as `a negative integer, zero, or a positive`.

Godin:
@marchof merged! Nice trick to not change API 👍Hope you don't mind that I fixed little things mentioned above.

Also given that people do not read documentation wanna hope that such message will be enough without `maybe you forgot "%"` 😆

marchof:
@Godin Many thanks for cleaning up my mess! 😺

PR #932, opened 2019-08-30
Add FAQ entry about class filtering in reports

marchof:


Godin:
@marchof addition of such entry into FAQ is nice since this indeed seems to be a quite frequent source of questions, and so I'm wondering - WDYT maybe text on other page(s) (such as [page about agent](https://www.jacoco.org/jacoco/trunk/doc/agent.html)) can be somehow improved? I'm ok to merge this PR as it is if you don't have ideas for other pages.

Godin:
Rephrasing a bit my question - maybe source of questions is misunderstanding due to not very clear text of text on other pages?

marchof:
@Godin Agreed, FAQ entry is a symptom for lack of documentation. I'll give it a try.

marchof:
@Godin I added hints at the respective doc pages (only to the exclude options). Is this what you were thinking about?

Godin:
@marchof yup! :thumbsup:

PR #934, opened 2019-09-02
Remove unnecessary about.html files

marchof:
These modules are not distributed.


marchof:
@Godin Ups, removed now.

Godin:
@marchof sorry that I noticed this only after first review - [we also don't distribute `org.jacoco.example`](https://github.com/jacoco/jacoco/blob/v0.8.4/org.jacoco.examples/pom.xml#L28) , only  [sources from it](https://github.com/jacoco/jacoco/blob/v0.8.4/org.jacoco.examples/assembly.xml) are [unpacked into zip distribution](https://github.com/jacoco/jacoco/blob/v0.8.4/jacoco/assembly.xml#L97-L104), so maybe also remove `org.jacoco.example/about.html`?  WDYT?

marchof:
@Godin Sure, removed now. I wonder whether it is possible to automatically check that every published artifact contains a `about.html` file?

PR #941, opened 2019-09-20
Check whether source reference actually is a file

marchof:
A corner case has been reported where the source reference points to an folder. Instead of ignoring such a reference (as we do for missing files) in this case the report generation process fails. This fix checks that the source reference actually points to a file.

The issue war reported on the mailing list: https://groups.google.com/forum/?fromgroups=#!topic/jacoco/cx9sTR3KMr4


Godin:
@marchof I'm wondering if this deserves changelog entry? 😉

marchof:
@Godin Sure. Done

Godin:
@marchof thanks! 👍

PR #943, opened 2019-09-20
Update license to EPL version 2.0

Godin:
After nice cleanup of information about licenses by #926 , #918 and #934, I propose to update Eclipse Public License from version 1.0 to version 2.0.

This requires

* following modification of headers in source files

```diff
  * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
  *
+ * SPDX-License-Identifier: EPL-2.0
```

* following modification in `about.html` files

```diff
   All Content in this distribution is made available by Mountainminds GmbH &amp; Co.
   KG, Munich. Unless otherwise indicated below, the Content is provided to you
-  under the terms and conditions of the Eclipse Public License Version 1.0
+  under the terms and conditions of the Eclipse Public License Version 2.0
   (&quot;EPL&quot;). A copy of the EPL is available at
-  <a href="http://www.eclipse.org/legal/epl-v10.html">http://www.eclipse.org/legal/epl-v10.html</a>.
+  <a href="https://www.eclipse.org/legal/epl-2.0/">https://www.eclipse.org/legal/epl-2.0/</a>.
   For purposes of the EPL, "Program" will mean the Content.
```

* replacement of file `org.jacoco.doc/docroot/doc/epl-v10.html` that contains text of license
* and update of metainformation and checks in `org.jacoco.build/pom.xml`

At the same time I propose to also remove [optional `Contributors` field](https://www.eclipse.org/projects/handbook/#ip-copyright-headers) from headers.

Changes are in separate commits to simplify review, but prior to merge should be squashed into single one.


org.jacoco.core.test.validation.java7/src/org/jacoco/core/test/validation/java7/targets/TryWithResourcesTarget.java
```diff
@@ -32,6 +32,7 @@ public void close() {
 	 */
 	private static Object test() throws Exception {
 		nop(); // assertFullyCovered()
+
 		try ( // assertTry()
```
Godin: Change of header revealed following strange behavior of ECJ `3.12.1` that we currently use in tests:

compilation of

```
/* comment with 20 lines */
import java.io.Closeable;

class Example {
    private static class Resource implements Closeable {
        @Override
        public void close() {
        }
    }

    private static void nop() {
    }

    private static void example() throws Exception {
        nop();
        try( // line 35
            Resource r1 = new Resource();
            Closeable r2 = new Resource();
            AutoCloseable r3 = new Resource();
        ) {
            nop();
        }
    }
}
```

results in bytecode with following `LineNumberTable`:

```
        line 34: 0
        line 36: 7
        line 37: 16
        line 3: 25
        line 40: 35
        line 41: 38
        line 42: 176
```

however if first comment contains one less/more lines than `20`, i.e. `try` keyword is not on line `35`, then `LineNumberTable` is different - for example when method starts at line `35`:

```
        line 35: 0
        line 36: 3
        line 37: 7
        line 38: 16
        line 39: 25
        line 41: 35
        line 42: 38
        line 43: 176
```

notice absence of entry for offset `3` in the first case.

marchof:
@Godin I hope our former contributors are o.k. with removing their names from source files. Do you know of a reasoning behind the removal at eclipse.org?

Godin:
@marchof AFAIK this "field" was always optional - here is the the oldest entry of "handbook" I can find: http://web.archive.org/web/20151106184324/https://www.eclipse.org/projects/handbook/ Projects are not forced to keep/remove it.

However AFAIK it was well maintained only a long-long time ago, let's say in pre-Git history, and nowadays usually not maintained properly, i.e. not updated over time.

I have feeling that the same in our case - value is only correct at a time of file creation, important and big file updates don't result in update of this field, and so that its current value is simply outdated in many cases. I propose removal because of this, and because all authors of all changes are anyway already properly recorded by Git, and also we do already mention contributors in changelog - search by "contrib" in it. And this way in future we only will need to pay attention on mentioning contributions in changelog, instead of reviewing header of each individual file.

Godin:
Can also make analogy with archaic javadoc tag `@author` - https://github.com/jacoco/jacoco/commit/2ccb2fc041b228cec72a256a64bc992f4f7bc331 😉

marchof:
@Godin Ok, let's make things simple and remove contributors from individual source files. Git history and change log should be enough.

With some grepping

    grep -R -A3 -h "Contributors:" **/* | grep " - " | sort | uniq

I come to the following contributors list in our sources:

```
Andres Almiray
Brock Janiczak
Chas Honton
Dominik Stadler
Evgeny Mandrikov
Fabian Mastenbroek
Jan Wloka
John Keeping
John Oliver
Kyle Lieber
Mads Mohr Christensen
Marc R. Hoffmann
Mirko Friedenhagen
Nikolay Krasko
Oliver Nautsch 
Sonatype, Inc.
```

Out of which the following people are not mentioned in change log (beside us):

```
Brock Janiczak
Chas Honton
Jan Wloka
John Keeping
John Oliver
Mirko Friedenhagen
Oliver Nautsch 
Sonatype, Inc.
```

I wonder whether we might create a separate contributors.html page.

What about cleaning up contributors information in a separate PR?

Godin:
> What about cleaning up contributors information in a separate PR?

@marchof fine :thumbsup: I excluded [remove_optional_field.patch.txt](https://github.com/jacoco/jacoco/files/3640053/remove_optional_field.patch.txt) from this PR.

BTW don't know how/where you found as contributor `Sonatype, Inc.` company that hosts Maven repository :laughing: - see lines that was removed by patch:

```
cat remove_optional_field.patch.txt | grep "^\- " | sort | uniq
```

marchof:
@Godin I should have excluded `target` folders... :see_no_evil:

    ./jacoco-maven-plugin.test/target/local-repo/org/eclipse/aether/aether/0.9.0.M2/aether-0.9.0.M2.pom

PR #947, opened 2019-09-29
Upgrade ASM to 7.2

Godin:


marchof:
@Godin ASM has now official support for Java 14. Are you planning to remove our current "downgrade trick" for Java 14?

Godin:
@marchof no, I don't plan to remove it - in order to avoid the need to re-add it later for Java 15 😉

PR #948, opened 2019-09-30
Update issue templates

marchof:
* Add CLI
* Make clear that other tools are not supported here
* Ask for executable reproducer
* Ask for full stack trace
* Clarify rule about closing issues as invalid

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,17 +1,3 @@
-This is a issue tracker. Please use our mailing list for general questions: 
-https://groups.google.com/forum/?fromgroups=#!forum/jacoco
+PLEASE USE ONE OF THE FOLLOWING TEMPLATES
 
-Also check FAQ before opening an issue: http://www.jacoco.org/jacoco/trunk/doc/faq.html
-
-
-### Steps to reproduce
-
-JaCoCo version:
-Operating system:
-Tool integration: Maven/Ant/API/Other
-
-
-### Expected behaviour
-
-
-### Actual behaviour
+https://github.com/jacoco/jacoco/issues/new/choose
--- a/.github/ISSUE_TEMPLATE/1_question.md
+++ b/.github/ISSUE_TEMPLATE/1_question.md
@@ -1,0 +1,14 @@
+---
+name: 'Question'
+about: 'Questions and requests for support'
+title: 'PLEASE DO NOT USE GITHUB FOR QUESTIONS - CHECK FAQ OR USE FORUM'
+labels: 'declined: question :x:'
+assignes:
+
+---
+
+https://www.jacoco.org/jacoco/trunk/doc/faq.html
+
+https://groups.google.com/forum/?fromgroups=#!forum/jacoco
+
+Thank you!
--- a/.github/ISSUE_TEMPLATE/2_bug_report.md
+++ b/.github/ISSUE_TEMPLATE/2_bug_report.md
@@ -1,0 +1,32 @@
+---
+name: 'Bug report'
+about: 'Report a bug in JaCoCo'
+title:
+labels: 'type: bug :bug:'
+assignes:
+
+---
+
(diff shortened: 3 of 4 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,17 +1,22 @@
...
 
-JaCoCo version:
-Operating system:
-Tool integration: Maven/Ant/API/Other
-
+* JaCoCo version:
+* Operating system:
+* Tool integration: Maven/Ant/CLI/API (for other report to the respective project)
+* Minimal Executable Reproducer: (complete setup, e.g. GitHub Repo)
+* Steps: (what exactly your you doing with the reproducer above?)
```
marchof: Typo: your --> are

Godin:
@marchof since some time GitHub allows to create [multiple templates](https://help.github.com/en/articles/about-issue-and-pull-request-templates#issue-templates) and by using them we can have following, which hopefully will be more understandable for people:

choose

bug

question

I can amend this PR or open new one. WDYT?

marchof:
@Godin Cool, I wasn aware of it. Yes please add commits to this PR.

In this case I wonder whether we should create a feature request template asking for a exact description of the use case.

marchof:
Hi @Godin, what is the difference between `.github/ISSUE_TEMPLATE.md` and `.github/ISSUE_TEMPLATE/bug_report.md`?

Why do we not explicitly ask for reproducer steps etc. in the latter?

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,14 @@
+---
+name: 'Question'
+about: 'Questions and requests for support'
+title: 'PLEASE DO NOT USE GITHUB FOR QUESTIONS - CHECK FAQ OR USE FORUM'
+labels:
```
marchof: Can we directly add "declined: question" here? 🤣

Godin:
> what is the difference between `.github/ISSUE_TEMPLATE.md` and `.github/ISSUE_TEMPLATE/bug_report.md`?
> Why do we not explicitly ask for reproducer steps etc. in the latter?

@marchof I made only "draft" as stated in commit message and then wanted to discuss/polish exact content with you. And that's also why existing file was not touched.

However there is also indeed difference between these files:

templates

And so we need to adjust content of all three. Last one can be blank or removed.

----

> I wonder whether we should create a feature request template asking for a exact description of the use case.

Do you have proposal of exact content? It also can be added later - IMO most important here is a clear separation of questions from everything else.

.github/ISSUE_TEMPLATE/question.md
Godin: I wanted to discuss this 😆 because yes - we can as shown on last screenshot in https://github.com/jacoco/jacoco/pull/948#issuecomment-538687538 🤣however this looks a bit weird given that it will be open for a while. I can imagine GitHub Action that will be closing them automatically, but fear that this might cause that people will be creating multiple ones - anyway this is a bit premature and better be discussed separately later after introduction of clear separation.
 
Also this looks a bit inconsistent with labels `type: bug` and `type: enhancement`. So I'm wondering what do you think about renaming of `declined: question` into just `type: question` for consistency and usage of it here? or without jokes use `declined: question` here?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,27 @@
...
+http://www.jacoco.org/jacoco/trunk/doc/faq.html
+
+https://groups.google.com/forum/?fromgroups=#!forum/jacoco
+
+Please understand that
+ISSUES WITHOUT FOLLOWING INFORMATION WILL BE CLOSED WITHOUT COMMENTS!
+
+### Steps to reproduce
+
+* JaCoCo version:
```
Godin: Got an idea - here we can easily exclude third-party tools such as SonarQube and Jenkins by requesting

```suggestion
* JaCoCo version: (from right bottom corner of JaCoCo report)
```

@marchof WDYT?

marchof: 👍

PR #953, opened 2019-10-04
Find source files also in shortened package directories

### LINKED ISSUE #939: Support for Kotlin Convention packaging for report
When setting up JaCoCo for a Kotlin project using the [Coding Conventions recommended directory structure](Coding Conventions recommended directory structure) (putting files at the root if they share the same common package), the test coverage is calculated as expected but the report does not show the source files due to the following problem:

> Source file "com/example/controller/ExampleController.kt" was not found during generation of report.

This appears to me to be because of the way [https://github.com/jacoco/jacoco/blob/master/org.jacoco.report/src/org/jacoco/report/InputStreamSourceFileLocator.java](InputStreamFileLocator) assumes the packaging will be in a folder structure matching the full package name.

It would be beneficial to have a way to configure JaCoCo to pick up files from the correct directory such as in this example from src/main/kotlin/controller rather than src/main/kotlin/com/example/controller by perhaps providing a package prefix similarly to IntelliJ's `packagePrefix` which corresponds to the root of the source directory. This allows it to support Kotlin projects following the convention.

### Steps to reproduce
JaCoCo version: 0.8.4
Operating system: Mac OSX
Tool integration: Gradle

### Expected behaviour
File source should be shown highlighted with test coverage in the report, possibly by customising a root package name

### Actual behaviour
File source is not shown highlighted with test coverage in the report

marchof:
Fixes #939 

The PR is missing a test case for Maven. Not sure whether this is possible with a Java project.

@Godin Do you think we can add a Kotlin project to the Maven IT tests? Or do you have another proposal how to test this?

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
@@ -38,6 +38,7 @@ import org.jacoco.report.IReportGroupVisitor;
 import org.jacoco.report.IReportVisitor;
 import org.jacoco.report.ISourceFileLocator;
 import org.jacoco.report.MultiReportVisitor;
+import org.jacoco.report.ShortenedPackageSourceFileLocator;
 import org.jacoco.report.check.IViolationsOutput;
 import org.jacoco.report.check.Rule;
 import org.jacoco.report.check.RulesChecker;
@@ -182,7 +183,8 @@ final class ReportSupport {
 			final List<String> includes, final List<String> excludes,
 			final String srcEncoding) throws IOException {
 		processProject(visitor, bundeName, project, includes, excludes,
-				new SourceFileCollection(project, srcEncoding));
+				new ShortenedPackageSourceFileLocator(
+						new SourceFileCollection(project, srcEncoding)));
 	}
 
 	private void processProject(final IReportGroupVisitor visitor,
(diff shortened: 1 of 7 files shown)


Godin:
> Do you think we can add a Kotlin project to the Maven IT tests? Or do you have another proposal how to test this?

@marchof let's deprecate `jacoco-maven-plugin` in favor of `maven-antrun-plugin` with already tested JaCoCo Ant Tasks 😆More seriously - [here is example of Maven project](https://github.com/jacoco/jacoco/files/3691382/example.zip) that produces following report before this change

before

and following after

after

Godin:
Here is example of how people usually generate aggregated report using Gradle JaCoCo Plugin, it contains two modules with Kotlin source files, where packages are different (`a` and `b`), but names of files are the same (`Example.kt`) and directories denoting packages are completely omitted - 
[example.zip](https://github.com/jacoco/jacoco/files/3714999/example.zip) And in this case report incorrectly shows the same source file for both packages.

milgner:
Changelog has been added but now it seems like there's a conflict with the master branch. I just stumbled over this when trying to integrade CodeClimate with Jacoco in a Kotlin project. Merging this would be highly appreciated! :+1:

Fleshgrinder:
Any chance to get this revived, this is still a pita. 😊

marchof:
Die due the ambiguity that @Godin identified we will probably drop this.

Fleshgrinder:
Which would bring us back to exposing a configuration property with which this could be enabled, as already discussed in the https://github.com/jacoco/jacoco/issues/939. Let's continue the discussion in the issue then. 😊

PR #954, opened 2019-10-04
Format code base consistently

marchof:
I would like to propose a consistent formatting of our code base in a single commit.

With this PR all *.java files now comply with the Eclipse 2019-09 formatter. I also applied the formatter to the test code snippets in `jacoco-maven-plugin.test` which are not autoformatted because they are not in Eclipse source folders.

### Reasoning

From the very beginning we had "auto formatting" configured for JaCoCo resulting in a consistent formatting of all source files. A couple of years ago the Eclipse formatter was reworked. Basically it does now IMHO a better job when in comes to line wrapping. This has introduced inconsistencies in our source base between "old" files and newly created source files. Also when working on files formatted the old way autoformating  introduces unrelated changes.

### DIFF
--- a/jacoco-maven-plugin.test/it/it-check-fails-halt/src/main/java/Example.java
+++ b/jacoco-maven-plugin.test/it/it-check-fails-halt/src/main/java/Example.java
@@ -13,8 +13,8 @@
  *******************************************************************************/
 public class Example {
 
-  public void sayHello() {
-    System.out.println("Hello world");
-  }
+	public void sayHello() {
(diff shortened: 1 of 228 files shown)


Godin:
As you might know, I use `org.jacoco.core/.settings/org.eclipse.jdt.core.prefs` as a config for Eclipse Code Formatter in IntelliJ IDEA 😜 so as a check reformatted all Java files in `master` branch and diff with this PR is empty 👍 Also, looking at the changes, seems that in many places the new formatter indeed works better.

PR #956, opened 2019-10-05
Check code formatting during build

Godin:
@marchof thanks to your explanation ❤️ in https://github.com/jacoco/jacoco/pull/954 

> A couple of years ago the Eclipse formatter was reworked.

I  finally understood why Eclipse Code Formatter in IntelliJ at some moment in time started to change formatting when applied on old code. And why [Spotless](https://github.com/diffplug/spotless) was doing the same during my previous attempt to use it for checking of formatting during build.

Given that after https://github.com/jacoco/jacoco/pull/955 formatting will be completely consistent, I propose to add check into build. This will also allow contributors to easily fix formatting even without proper IDE setup - following message is printed when check fails: `Run 'mvn spotless:apply' to fix these violations.` 🎉

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -474,6 +474,11 @@
           <artifactId>maven-bundle-plugin</artifactId>
           <version>3.5.1</version>
         </plugin>
+        <plugin>
+          <groupId>com.diffplug.spotless</groupId>
+          <artifactId>spotless-maven-plugin</artifactId>
+          <version>1.24.3</version>
+       </plugin>
       </plugins>
     </pluginManagement>
 
@@ -549,6 +554,31 @@
                 </requireReleaseDeps>
               </rules>
             </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <configuration>
+          <java>
+            <lineEndings>UNIX</lineEndings>
+            <includes>
+              <include>src/**/*.java</include>
+            </includes>
+            <eclipse>
+              <version>4.12.0</version>
+              <file>../org.jacoco.core/.settings/org.eclipse.jdt.core.prefs</file>
+            </eclipse>
+          </java>
+        </configuration>
+        <executions>
+          <execution>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
           </execution>
         </executions>
       </plugin>
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.java text eol=lf


org.jacoco.build/pom.xml
```diff
@@ -553,6 +558,30 @@
...
       </plugin>
 
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <configuration>
+          <java>
+            <includes>
+              <include>src/**/*.java</include>
+            </includes>
```
marchof: I would like to propose the follwing configuration:

    <includes>
      <include>**/*.java</include>
    </includes>
    <excludes>
      <exclude>target/**/*</exclude>
    </excludes>

This would basically check every `*.java` file in our tree, especially code in `jacoco-maven-plugin.test`which is currently not covered.

Godin: @marchof I added `.gitattributes` and configured Spotless to check that Java files use `LF` as line endings.

Unfortunately
```
<include>**/*.java</include>
```
causes weird reformatting of files in `jacoco-maven-plugin.test` - for example
```
jacoco-maven-plugin.test/it/it-multi-module/child/src/main/java/Example.java
```
but not
```
jacoco-maven-plugin.test/it/it-multi-module/child/src/main/java/somepackage/Example.java
```
My best guess is that in absence of `package` declaration first comment is considered as javadoc rather than as header. Maybe we can move files and add `package` declaration or add some javadoc. Not sure what will better. So I propose to merge this as is and investigate formatting of `jacoco-maven-plugin.test/it` later separately.

org.jacoco.build/pom.xml
```diff
@@ -553,6 +558,31 @@
...
+            <eclipse>
+              <version>4.12.0</version>
+              <file>../org.jacoco.core/.settings/org.eclipse.jdt.core.prefs</file>
+            </eclipse>
+          </java>
+        </configuration>
+        <executions>
+          <execution>
+            <phase>prepare-package</phase>
```
marchof: The license check we do in phase `verify`. Why do we use different phase here? Shouldn't be both in `verify` oder even `validate`?

Godin: @marchof in fact maybe we should reconsider phase used for check of license header - my rationale for usage of `prepare-package`:

Localy I often use `mvn test` just to do a quick test of something, what not necessarily will be committed and might be removed right after. In this case failure of license check seems unnecessary and bothers sometimes.

Also localy I often use `mvn package` to not execute time consuming Maven integration tests. 

So decided to use `prepare-package` phase in order to not complicate first case by check of formatting and in order to not miss it in second case.

marchof: @Godin Then we move the license check in a separate PR?

Godin: @marchof yup

PR #957, opened 2019-10-07
Normalize line endings in HTML, CSS and XSL source files

Godin:
Similarly to #955.

Also adds `*.html`, `*.css` and `*.xsl` to `.gitattributes`


Godin:
> Should also hold for style sheets

@marchof fixed too

PR #959, opened 2019-10-09
Remove trailing whitespaces in Java source files

marchof:


Godin:
@marchof I tried to enable `Spotless` check on top of this, but seems that you forgot `org.jacoco.core.test` :wink: I'll fix this.

PR #970, opened 2019-10-22
Add current project to report aggregate

gaellalire:
Current project is not included with report-aggregate goal.
If it is a neglect you can merge this branch, otherwise you should add an option to allow the inclusion of current project


lion7:
It would help if the option to include the current project to be configurable.
That way the current behavior won't change (which might in some cases break things).
See also the comment in https://github.com/jacoco/jacoco/pull/680#issuecomment-387501358

pertsodian:
Hi, should we improve this PR to incorporate the above comment to make this behaviour configurable (and disabled by default)? The fact that this is the 3rd PR submitted to add this change, a lot of people are looking to use this configuration (myself included)

marchof:
Superseded by configurable #1007

Issue #771, opened 2018-10-22
Add filter for exhaustive switch-expression


Godin:
Currently for exhaustive switch-expression ([JEP 361](https://openjdk.java.net/jeps/361))

```java
public class SwitchExpressionTarget {

	private enum E {
		TRUE, FALSE,
	}

	private static int target(E e) {
		return switch (e) {
			case TRUE -> 1;
			case FALSE -> 0;
		};
	}

	public static void main(String[] args) {
		target(E.TRUE);
		target(E.FALSE);
	}

}
```

with JDK 14 EA b29

```
$ java --version
openjdk version "14-ea" 2020-03-17
OpenJDK Runtime Environment (build 14-ea+29-1384)
OpenJDK 64-Bit Server VM (build 14-ea+29-1384, mixed mode, sharing)

$ javac --release 14 SwitchExpressionTarget.java

$ java -javaagent:jacoco-0.8.5/lib/jacocoagent.jar SwitchExpressionTarget 

$ java -jar jacoco-0.8.5/lib/jacococli.jar \
    report jacoco.exec \
    --classfiles SwitchExpressionTarget.class \
    --sourcefiles . \
    --html report
```

we produce

![report](https://user-images.githubusercontent.com/138671/47281473-91794c80-d5db-11e8-8c93-72024b4e25b0.png)

branch that throws `IncompatibleClassChangeError` should be filtered out

```
  private static int target(SwitchExpressionTarget$E);
    descriptor: (LSwitchExpressionTarget$E;)I
    flags: (0x000a) ACC_PRIVATE, ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: getstatic     #2                  // Field SwitchExpressionTarget$1.$SwitchMap$SwitchExpressionTarget$E:[I
         3: aload_0
         4: invokevirtual #3                  // Method SwitchExpressionTarget$E.ordinal:()I
         7: iaload
         8: lookupswitch  { // 2
                       1: 36
                       2: 41
                 default: 46
            }
        36: iconst_1
        37: istore_1
        38: goto          54
        41: iconst_0
        42: istore_1
        43: goto          54
        46: new           #4                  // class java/lang/IncompatibleClassChangeError
        49: dup
        50: invokespecial #5                  // Method java/lang/IncompatibleClassChangeError."<init>":()V
        53: athrow
        54: iload_1
        55: ireturn
      LineNumberTable:
        line 8: 0
        line 9: 36
        line 10: 41
[...]

PR #987, opened 2019-12-19
Add validation tests for Java 14 switch expressions

marchof:


Godin:
@marchof One of the features of switch expressions - is automatic insertion of implicit default clause in case when all enum values are listed. See #771. So I'm wondering if this case should/can be included into this PR?

org.jacoco.core.test.validation.java14/src/org/jacoco/core/test/validation/java14/SwitchExpressionsTest.java
```diff
@@ -0,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2019 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
```
Godin: @marchof copy-paste? 😉 that's [one of the reasons why I think that `Contributors` field doesn't make sense nowadays](https://github.com/jacoco/jacoco/pull/943#issuecomment-533833196)

marchof: Grrr, yep let's remove them soon. Fixed.

marchof:
@Godin Good idea, exhaustive example added.

PR #990, opened 2020-01-06
Add filter for Records

Godin:
I propose to filter methods `toString`, `hashCode` and `equals` that compiler generates for Records ([JEP 359](https://openjdk.java.net/jeps/359)), leaving accessor methods untouched, which is consistent with what we do for Kotlin `data class`es.

----

For

```
record Point(int x, int y) {
        new Point(0, 0);
    }
}
```

Using  `java -version`

```
openjdk version "14-ea" 2020-03-17
OpenJDK Runtime Environment (build 14-ea+30-1385)
OpenJDK 64-Bit Server VM (build 14-ea+30-1385, mixed mode, sharing)
```

```
javac \
    --enable-preview \
    --release 14 \
    -d classes \
    Point.java

java \
    --enable-preview \
    -javaagent:jacocoagent.jar \
    -cp classes \
    Point
```

Generation of report

```
java -jar jacococli.jar \
    report jacoco.exec \
    --classfiles classes \
    --sourcefiles src \
    --html report
```

before this change produces

before

and after

after


org.jacoco.core.test.validation.java14/pom.xml
```diff
@@ -35,4 +35,25 @@
...
+          <compilerArgs>
+            <arg>--enable-preview</arg>
+          </compilerArgs>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>--enable-preview</argLine>
```
marchof: Out of curiosity: Why does surefire enable this flag?

Godin: @marchof enablement of Preview Features by `--enable-preview` flag is required both at compile and at run time - see [Preview Language and VM Features (JEP 12)](https://openjdk.java.net/jeps/12), otherwise we'll get

```
java.lang.UnsupportedClassVersionError: Preview features are not enabled for org/jacoco/core/test/validation/java14/RecordsTest (class file version 58.65535). Try running with '--enable-preview'
```

marchof: Right, thanks!

ice1000:
Should we also ignore the accessors?

ice1000:
See #1215

PR #1005, opened 2020-01-30
Add validation test for Kotlin lambda expressions

Godin:


Godin:

@qwwdfsad FYI an example of our validation tests :wink:

PR #1006, opened 2020-01-30
Add filter for plain getter and setter

DanielKraft:
# Plain getter and setter filter
This filter is inspired by the [FilteringOptions: Plain getters and setters](https://github.com/jacoco/jacoco/wiki/FilteringOptions#filters-for-code-where-test-execution-is-questionable-or-impossible-by-design)

## Filter criteria
- Method name starts with `get`, `is` or `set`
- Method name contains the name of a field
- Method body contains only standard logic like `return name` or `this.name = name`

### Examples
```java
private String name;
private boolean valid;

// Ignored because of the name 'getName' and the body
public String getName() {
   return name;  
}

// Ignored because of the name 'setName' and the body
public void setName(String name) {
   this.name = name; 
}

// Not ignored because of the custom body
public void setName(String name) {
   if (name.length() > 0) {
      this.name = name; 
   }
}

// Ignored because of the name 'isValid' and the body
public boolean isValid() {
   return valid; 
}

// Not ignored because of the custom name 'valid'
public boolean valid() {
   return valid; 
}
```

## API
The filter can be disabled via the following APIs.

#### report
```shell
java -jar jacococli.jar report --disablePlainGetterSetter
```

#### report
```xml
<properties>
   <jacoco.disablePlainGetterSetter>true</jacoco.disablePlainGetterSetter>
</properties>
```
#### check
```xml
<execution>
   <id>check</id>
   <goals>
      <goal>check</goal>
   </goals>
   <configuration>
      <disablePlainGetterSetter>true</disablePlainGetterSetter>
   </configuration>
</execution>
```

#### report
```xml
<jacoco:report>
   <structure name="root">
      <filters disablePlainGetterSetter="true" />
   </structure>
</jacoco:report>
```


marchof:
Thanks for the contribution!

As we already decided earlier (see PR #923) we will not add filters for user written source code.

Also as you can see configurable filters are really painful to maintain in the various integrations we have. This approach will obviously not scale if we add more configuration options. Also think of other integrations which are not in this repository (EclEmma, Gradle etc).

Please understand that due to our limited ressources to maintain this project we will cannot add additional complexity to it.

Issue #812, opened 2018-12-31
report-aggregate does not include report from current project


sidgate:
### Steps to reproduce

Create a multi-module maven project with one module as aggregator. Following is a sample jacoco plugin configuration

```
<plugin>
                <groupId>org.jacoco</groupId>
                <artifactId>jacoco-maven-plugin</artifactId>
                <version>0.8.2</version>
                <executions>
                    <execution>
                        <id>prepare-agent</id>
                        <goals>
                            <goal>prepare-agent</goal>
                        </goals>
                    </execution>
                    <execution>
                        <id>report</id>
                        <phase>prepare-package</phase>
                        <goals>
                            <goal>report</goal>
                        </goals>
                    </execution>
                    <execution>
                        <id>report-aggregate</id>
                        <phase>verify</phase>
                        <goals>
                            <goal>report-aggregate</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

https://stackoverflow.com/questions/53983849/jacoco-report-aggregate-include-report-of-aggregate-project

JaCoCo version: 0.8.2
Operating system: Windows
Tool integration: Maven

### Expected behaviour
report-aggregate should include test report from current aggregate module

### Actual behaviour
report is generated only with dependent modules

The plugin does [load](https://github.com/jacoco/jacoco/blob/9c45dab3e9d9ed888481005095b30d3e5b55b338/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java#L109) the execution data for current project, but creates report only from [dependencies](https://github.com/jacoco/jacoco/blob/9c45dab3e9d9ed888481005095b30d3e5b55b338/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java#L135) . May we should just add following before the for loop

`support.processProject(group, getProject().getArtifactId(),
					getProject(), getIncludes(), getExcludes(), sourceEncoding);`

Godin:
There are already two PRs #430 and #680 (so please search before creation of ticket in future). They are almost identical and contain your proposal, as well as description of why they was not accepted as is. In case if this description is unclear:

current behavior is there since the inception of `report-aggregate` and IMO [clearly stated in documentation](https://www.jacoco.org/jacoco/trunk/doc/report-aggregate-mojo.html)

> The report is created from all modules this project **depends on**.

[...]

PR #1007, opened 2020-01-30
Add parameter to include the current project in the aggregated report

lion7:
This PR tries to solve the issue described in https://github.com/jacoco/jacoco/issues/812.
It is similar to https://github.com/jacoco/jacoco/pull/430, https://github.com/jacoco/jacoco/pull/680 and https://github.com/jacoco/jacoco/pull/970 but with the https://github.com/jacoco/jacoco/pull/680#issuecomment-387501358 of @marchof and https://github.com/jacoco/jacoco/issues/812#issuecomment-452140634 of @Godin taken into account.


ramachandranms:
@marchof, @Godin 

gentle bump, if this has fallen through cracks. would be great to have this feature as the workarounds are not easier to understand and get through. greatly appreciate a review to merge this. thanks in advance.

pertsodian:
Hi @marchof & @Godin, can we review this PR? I think there are already quite a number of requests for this feature.

kakawait:
Since the way that PR is dev, is not breaking changes. So by reviewing it and merging it, you'll make many devs happy no? :)

lordjaxom:
+1
Would save us the hassle of introducing aggregator projects in 50+ repositories.

lordjaxom:
Just to state a report, we are currently using jacoco 0.8.5 + issue-812 (built locally) successfully within 50+ repositories to have full and accurate coverage reports with SonarQube without having to fuss with any projects' pom (both report-aggregate and SonarQube are configured in a global parent pom of our project tree).

marchof:
@lordjaxom Can you describe the structure of your repos and shortly explain why this modification helps? I want to understand the use case. Thx!

lordjaxom:
We mostly have multi-module-projects for EJB components which are separated into shared and server modules. The shared part contains interfaces, DTOs and utility classes, while the server part contains entities and implementations. 

Since structurally sometimes the tests are located within the server part (even though the test covers DTO and utility classes) or integration tests cover multiple components at once, we would need aggregate reports over two or more modules. Ideally those would be generated within the server part, covering both server and shared.

Without this flag, we would either have to aggregate ourselves the standard report (covering only the server part) and the aggregate report (covering only the shared part since only dependencies are scanned) or introduce a third module as suggested in various documents to cover classes from both modules. Since "report" only actually reports coverage for classes present in the current module (not dependencies) and "report-aggregate" only reports coverage for classes from the dependencies, not the project itself.

zuza-str:
@marchof are there any other concerns to merge this change? It would really make the life easier for many JaCoCo users :)

henrik242 (approved):
lgtm

twoodhouse (approved):

twoodhouse:
My company is being forced to use aggregator modules in lots of projects to get around this. We are looking forward to this option being merged!

kakawait (approved):

cai3178940:
LGTM

bogaertg:
up

jycr:
@kakawait , @henrik242, @twoodhouse: what is missing now to be able to merge this PR into master and add this feature in the next version?

lyoumi:
@lion7 @henrik242 @kakawait  do we have any estimates for merging this PR to master?

kakawait:
I'm not maintainer at all of Jacoco. I can't help you

jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java
```diff
@@ -131,6 +138,10 @@ File getOutputDirectory() {
 	void createReport(final IReportGroupVisitor visitor,
 			final ReportSupport support) throws IOException {
 		final IReportGroupVisitor group = visitor.visitGroup(title);
+		if (includeCurrentProject) {
```
marchof, line 141: May I propose to extract the following method:


	private void processProject(final ReportSupport support,
			final IReportGroupVisitor group, final MavenProject project)
			throws IOException {
		support.processProject(group, project.getArtifactId(), project,
				getIncludes(), getExcludes(), sourceEncoding);
	}

This will avoid code duplication because we can simply call it here with

    processProject(support, group, project);

and the original call becomes

    processProject(support, group, dependency);

lion7: Changed this per your suggestion

lyoumi:
@marchof @lion7 just to follow up this review. Do we have any more restrictions to not PR?

marchof:
@lyoumi My concern is that all builds are broken due to a source formatting issue. I can help to fix this when I‘m back from vacation next week.

lion7:
> @lyoumi My concern is that all builds are broken due to a source formatting issue. I can help to fix this when I‘m back from vacation next week.

Oh, I totally missed that in my last commit. I will try to fix the build asap 🙂

marchof (approved):

marchof:
@Godin May you also have a look at this?

Godin (approved):

froque:
Any chance of releasing 0.8.9 with this change ?

m-herold:
Can't wait to start using the new `includeCurrentProject` parameter. Are there already plans when the new version `0.8.9` will be released?

PR #1010, opened 2020-01-31
Add filter for bridge methods

### LINKED ISSUE #1002: Kotlin: auto-generated methods for implementation with delegation are taken into account in coverage report
### Steps to reproduce

* JaCoCo version:  0.8.5.201910111838
* Operating system: macOS Catalina 10.15
* Tool integration: JaCoCo Gradle plugin
* Complete executable reproducer: https://github.com/qwwdfsad/coverage-test, package `coverage.delegates`, run command ` ./gradlew test jacocoTestReport`, report directory: `build/jacocoHtml`
* Steps: 
Implement any interface with delegation. Example from the reproducer:
```
class DelegatedList : List<Int> by ArrayList() {
    override val size: Int
        get() = 42
}
```

and use it as 
```
val list = DelegatedList()
```

### Expected behaviour
Coverage reports only one missing method, `size` 

### Actual behaviour
All `List` methods are reported as not covered

Godin:
Fixes #1002


org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -27,6 +27,9 @@ <h3>New Features</h3>
   <li>Methods <code>toString</code>, <code>hashCode</code> and <code>equals</code>
       generated by compiler for records are filtered out during generation of report
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/990">#990</a>).</li>
+  <li>Bridge methods not marked as <code>synthetic</code> by Kotlin compiler
```
marchof: Even if we discovered this because of the Kotlin compiler we filter bridge methods in any case (regardless of the compiler). Maybe we just say: _Bridge methods are filtered out during report generation_?

Godin: @marchof done

PR #1011, opened 2020-01-31
Adjust test to reflect fix of JDK-8237528 in JDK 15 EA b08

Godin:


Godin:
@marchof could you please review? The next build of master in Travis will fail without this, unless I'll switch it back on usage of previous JDK 15 EA b07.

marchof:
I wonder how many bugs/questions regarding `instanceof` we will get from people using Java 14 with JaCoCo... just curious.

PR #1013, opened 2020-02-02
Documentation should advise usage of "mvn verify" instead of "mvn install"

Godin:
Because last one pollutes local Maven repository.


Godin:
@marchof could you please review? I don't use `install` for ages and forgot that we have it in documentation - who read docs, especially own one :laughing:

Issue #977, opened 2019-11-20
Lambdas inside @Generated methods are not ignored


ntkoopman:
THIS IS A BUG TRACKER ONLY. FOR QUESTIONS PLEASE CHECK FAQ OR USE FORUM:

http://www.jacoco.org/jacoco/trunk/doc/faq.html

https://groups.google.com/forum/?fromgroups=#!forum/jacoco

Please understand that
ISSUES WITHOUT FOLLOWING INFORMATION WILL BE CLOSED WITHOUT COMMENTS!
Thank you for filling bug report!

### Steps to reproduce

* JaCoCo version:  0.8.5.201910111838
* Operating system: macOS 10.15.1 (19B88)
* Tool integration: Gradle
* Steps:

```java
  @Generated
  public Stream<String> generate() {
    List<String> data = new ArrayList<>();
    data.add("foo");
    data.add("bar");
    data.add("qux");
    return data.stream()
        .map(
            string -> {
              if (string.isEmpty()) return "";
              return string.substring(0, 1).toUpperCase() + string.substring(1);
            });
  }
```

### Expected behaviour

Everything inside the method is ignored by Jacoco

### Actual behaviour

Any statements in the method itself are ignored correctly, but any lambdas inside the method still show up and count towards to code coverage limits.

marchof:
Currenty JaCoCo processes methods independently from each other. To track the relationships between synthetic methods for lambda bodies and the methods where the lambdas are created we would need to introduce a second pass over all methods.

rammygit:
Is @nocoverage way to go

PR #1014, opened 2020-02-06
Added @NoCoverage

uweschaefer:
In order to consciously skip code from coverage measurement, @Generated can be used currently. We have the requirement of occasionally excluding code that is not exactly generated, but should be ignore for other reasons (Framework specific code etc). 
This is why we extended the behavior of @Generated also to code annotated with @NoCoverage.

Maybe this is useful for others as well.


marchof:
Please see [PR675](https://github.com/jacoco/jacoco/pull/675#issuecomment-608075757) for a explanation why we cannot add this feature to JaCoCo.

As a side note @Godin suggested to use `@NoCoverageShouldBeGenerated` as a workaround as a distinct annotation 😉

uweschaefer:
thank you for taking the time @marchof 

I understand and respect your decision on that. One thing though: I did not get the reference to @NoCoverageShouldBeGenerated. I don't see this annotation in jacocos codebase and i am a bit puzzled. Could you please point me to the code or discussion you are referenceing?

Thanks.

Godin:
@uweschaefer quoting https://www.jacoco.org/jacoco/trunk/doc/changes.html

> Classes and methods annotated by annotation whose retention policy is runtime or class and whose simple name contains "Generated" (previously equality was required) are filtered out during generation of report (GitHub #822).

uweschaefer:
thanks a lot, that'll help.

Godin:
@uweschaefer please also note that `NoCoverageShouldBeGenerated` is an abuse of the feature that was developed for filtering of code generated by Lombok and Groovy and not for manual exclusions. Such abuse has the same bad consequences as https://github.com/jacoco/jacoco/pull/675#issuecomment-608075757 including #977 "does not work for lambda bodies in the annotated methods". As such any ticket/request related to such abuse might be closed as invalid.

uweschaefer:
got that. However, it scratches an itch many users seem to have (looking at #675), so as long as it works (for the majority of cases), we'll probably make use of it, taking the consequences into account.
Thanks again.

PR #1017, opened 2020-02-09
Update to ASM 7.3.1

marchof:
This pull request only updates the dependency. It does not yet utilize the new `V15` constant or adds experimental support for Java 16.

### DIFF
--- a/jacoco/pom.xml
+++ b/jacoco/pom.xml
@@ -111,7 +111,7 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>4400000</maxsize>
+                  <maxsize>4500000</maxsize>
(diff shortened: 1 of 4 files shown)


Godin:
@marchof see my comment https://bugs.eclipse.org/bugs/show_bug.cgi?id=559150#c1 , which explains why tests are failing in `org.jacoco.core.test.validation.java14`. I doubt that there is an easy way to upgrade ASM to 7.3.1 without losing support for Java 14 preview features.

marchof:
@Godin Thanks for the info! Ups, looks like future versions of ASM will not allow us to support preview features any more :(  Maybe we have a chance to discuss this with ASM team.

PR #1019, opened 2020-02-10
Report coverage for Unit-returning suspend funs in Kotlin 1.3.60 correctly

poseidon-mysugr:
When the Kotlin 1.6.30 compiler encounters a `Unit` - returning `suspend fun` that forwards to calling another `Unit` - returning suspend fun as its last statement, and doesn't have other calls to `suspend fun`s, then it generates a ternary operator that can often not be covered in testing:
```kotlin
suspend fun a() {}
suspend fun b() {
  a() // last statement is another suspend fun returning `Unit`
}
```
The Bytecode generated for `b()` is:
```
L0
  LINENUMBER 3 L0
  ALOAD 0
  INVOKESTATIC my/package/Class.a (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
  DUP
  INVOKESTATIC kotlin/[...]/IntrinsicsKt.getCOROUTINE_SUSPENDED ()Ljava/lang/Object;
  IF_ACMPNE L1
  ARETURN
L1
  POP
L2
  LINENUMBER 4 L2
  GETSTATIC kotlin/Unit.INSTANCE : Lkotlin/Unit;
  ARETURN
L3
```
which decompiles to Java as follows
```java
public static final Object b(@NotNull Continuation $completion) {
  Object var10000 = a($completion);
  return var10000 == IntrinsicsKt.getCOROUTINE_SUSPENDED() ? var10000 : Unit.INSTANCE;
}
```
The ternary operator that can be seen in the decompiled Java code can often not be covered; so the whole ternary operator should be excluded from coverage.

This PR addresses this issue by searching for calls to `IntrinsicsKt.getCOROUTINE_SUSPENDED()`, followed by the instructions `IF_ACMPNE L1`, `ARETURN`, `L1`, `POP` with `L1` being the Label the `IF_ACMPNE` instructions jumps to. It then ignores all four instructions (`IF_ACMPNE` to `POP`, inclusively).
It furthermore ignores the `return Unit.INSTANCE` part, being the `GETSTATIC` and the following `ARETURN` instruction, because returning `Unit` can not be asserted in Kotlin anyway.
*Note:* the `return Unit.INSTANCE` part may or may not follow directly after the `POP` like in the given example; there may be a `GOTO` instruction to jump to it, as can be seen in the tests.


poseidon-mysugr:
I updated the PR from master and reworked the fix based on Evgeny Mandrikov's fix for the same problem - the only thing I found missing is that the `return Unit.INSTANCE` part should also be ignored.

Godin:
@poseidon-mysugr 

> `return Unit.INSTANCE` part should also be ignored

It corresponds to non-suspended exit from a function, so either will be executed or function infinitely suspends. In both cases coverage is consistent with non-suspending functions, while won't after your change:

```
import kotlinx.coroutines.GlobalScope
import kotlinx.coroutines.launch
import kotlinx.coroutines.yield

object KotlinCoroutineTarget {

    private suspend fun a1() {
        b1()
    } // assertNotCovered()

    private suspend fun b1() {
        while (true) yield()
    }

    private fun a1_without_suspend() {
        b1_without_suspend()
    } // assertNotCovered()

    private fun b1_without_suspend() {
        while (true) {}
    }

    private suspend fun a2() {
        b2()
    } // assertFullyCovered()

    private suspend fun b2() {
    }

    private fun a2_without_suspend() {
        b2_without_suspend()
    } // assertFullyCovered()

    private fun b2_without_suspend() {
    }

    @JvmStatic
        GlobalScope.launch { a1() }
        GlobalScope.launch { a1_without_suspend() }
        GlobalScope.launch { a2() }
        GlobalScope.launch { a2_without_suspend() }
        Thread.sleep(2000L)
    }

}
```

poseidon-mysugr:
@Godin You're right about indefinitely suspending functions, but there is one case you've missed: The `return Unit` instructions are never executed if the `suspend fun` that is being called _always_ suspends, but _not indefinitely_. For instance, if I modify your example in a very simple way:
```kotlin
    private suspend fun b2() {
        yield()
    }
```
then the return of `a2` is not fully covered anymore, and cannot be, as the actual return of `b2` will always be `IntrinsicsKt.getCOROUTINE_SUSPENDED()`, and never `Unit.INSTANCE`.

In some instances it's possible to write tests that trigger both paths, but if there's no control over the `suspend fun` that is being called (like in the example above), this may in fact be an impossible-to-reach branch, and we have a number of such examples in our code base.

I'm not sure how to filter this case while maintaining consistent behaviour with non-suspending functions, filtering the return altogether is the only thing I could come up with... Do you have a better idea?

PR #1032, opened 2020-03-28
Upgrade ASM to 8.0

Godin:


marchof:
@Godin Thx! There is now `Opcodes.V15`. Should we prepare for Java 16 support? 😆

PR #1040, opened 2020-04-02
WiP: Log total coverage

MeyerNils:
I'm facing the same use case as @mauriceqch in #488 and understood and can use different solution options like the XSLT transformation or awk. Nevertheless I'd prefer if I could avoid extra tool usage (being another thing all co-workers need to understand) but have a way of simply print out some summary information easily within the Maven Build. I was thinking about something like sketched in this pull request. I would extend this (see below) but did not want to put any effort into this before getting your opinion on this topic. 

I would add a Maven property (e.g. jacoco.printSummary) to control, whether details should be logged. If set to true, I'd log all counters in a similar way as below.

What do you think about this?

### DIFF
--- a/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
+++ b/jacoco-maven-plugin/src/org/jacoco/maven/ReportSupport.java
@@ -214,6 +214,12 @@ final class ReportSupport {
 		log.info(format("Analyzed bundle '%s' with %s classes",
 				bundle.getName(),
 				Integer.valueOf(bundle.getClassCounter().getTotalCount())));
+		log.info(format("%s of %s instructions have been covered (%s %%)",
+				Integer.valueOf(
+						bundle.getInstructionCounter().getCoveredCount()),
+				Integer.valueOf(bundle.getInstructionCounter().getTotalCount()),
+				Double.valueOf(bundle.getInstructionCounter().getCoveredRatio()
+						* 100)));
 		if (!nomatch.isEmpty()) {
 			log.warn(format(
 					"Classes in bundle '%s' do no match with execution data. "


marchof:
@MeyerNils As already explained in #488 JaCoCo already has several machine readable outputs of coverage data. We don't consider log files as an appropriate data source. Moreover change seems to be very specific to your needs and makes the tools inconsistent:

* beside Maven goal there is a Ant task and the CLI
* Beside instruction coverage we have a couple of more counters

MeyerNils:
Thanks for your promt reply @marchof. GitLab is using the log output of the build as datasource that can be used directly to find one summary number for the coverage using regex. That's why I was suggesting an additional log entry to save a parsing and printing step for GitLab users. I admit I'm not aware of another use case, but a least there are 4000 projects alone on GitLab.com for which this could be beneficial, not yet counting the on premise installations. 

As I wrote, I'm aware that the change would not yet be complete and need to be generalized, e.g. to be made consistent regarding other entry points, I could do so but did not want to spend the effort before knowing your thoughts.

marchof:
Hi @MeyerNils , do you have a link where the the Gitlab integration is specified?

MeyerNils:
@marchof, the documentation describes how to use it, see https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-parsing. This is than influencing the coverage bage, see https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-report-badge

I'm not aware of a further specification. 

#### Side information, not directly related to my suggestion here
Besides this overall "coverage summary" GitLab allows to visualize more details regaring coverage of the code suggested by a merge request, see https://docs.gitlab.com/ee/user/project/merge_requests/test_coverage_visualization.html for an example. This means the topic I'm describing here, might not be the only area where the JaCoCo / GitLab integration can be eased. This second visualization relies on Cobertura XML files, so either an adaption on GitLab or JaCoCo side or using a converter like https://github.com/rix0rrr/cover2cover in between will be needed here, if JaCoCo results should be visualized here, too. But this is a separate topic for me.

MeyerNils:
@marchof, log output can't be converted into coverage highlighting. They use log output for the "summary number" that is e.g. used in badges and Cobertura XML for coverage highlighting. I do not know why they're going two totally independent approaches here, perhaps because different teams implemented things. It's not perfectly consistent on their side from my point of view and I'm even not sure if it will stay like this forever. 

So even if I contributed a JaCoCo XML parser for GitLab it would not solve the summary number issue that I was tackling here with the log output.

Besides contributing to GitLab is a bit more difficult for me, because I'm not deeply into their technology stack, so it's not an option in the near future for me.

marchof:
@MeyerNils Just because another product has a crappy API this does not justify to do the same for JaCoCo 😉 

Please understand that we with our limited time budget for this free time project we cannot afford to add conceptually broken and incomplete features. The follow-ups on feature requests and bug reports will kill us.

MeyerNils:
I understand your view, no worries. I'll find a solution for me and will share it here just if somebody else is interested. Thanks for the discussion (and all the work on this project!)

PR #1052, opened 2020-05-19
Fix validation test for Java 15

marchof:
With second preview the constructors of records now get the same access
modifier than the class (see JDK-8242479).

For "record WithoutFields" this leads to an private empty default
constructor which is filtered out by PrivateEmptyNoArgConstructorFilter.

Removing private modifiers from records to have consistent results
between first and second preview.


marchof:
Build with latest JDK-15 build is green: https://pici.beachhub.io/#/jdk-jacoco/20200519-191453

PR #1055, opened 2020-06-08
Fix build with Java >= 12 when option -Dbytecode.version is not provided

andrioli:
Fixes #1054

PR #1056, opened 2020-06-08
Hello, GitHub Actions

andrioli:


andrioli:
Sorry for the noise, I want to create this PR in my fork :(

andrioli:
@marchof if you are okay with that I can re-open this PR and remove the Java 7 build. After merge https://github.com/jacoco/jacoco/pull/1055 the build will be greeen

marchof:
@andrioli Hm, we already have a build setup for all supported JREs (1.5 - 15). What would be the point to maintain another (incomplete) pipeline?

andrioli:
> @andrioli Hm, we already have a build setup for all supported JREs (1.5 - 15). What would be the point to maintain another (incomplete) pipeline?

I was just thinking that probably is a good idea to also run the build in CI with only the command `mvn verify` for the supported Java versions (i.e., Java >= 8 / no-toolchain). But to be honest I don't have a **strong** point for that.  😄

PR #1057, opened 2020-06-08
Improve support for multiple JaCoCo runtimes in the same VM

### LINKED ISSUE #331: java.nio.channels.OverlappingFileLockException for Jacoco offline
Hi, recently I used Jacoco 0.7.2 to test our AppServer, Encountered such a problem:
```
java.nio.channels.OverlappingFileLockException 
at sun.nio.ch.SharedFileLockTable.checkList(FileLockTable.java:255) 
at sun.nio.ch.SharedFileLockTable.add(FileLockTable.java:152) 
at sun.nio.ch.FileChannelImpl.lock(FileChannelImpl.java:1030) 
at java.nio.channels.FileChannel.lock(FileChannel.java:1053) 
at org.jacoco.agent.rt.internal_e6e56f0.output.FileOutput.openFile(FileOutput.java:69) 
at org.jacoco.agent.rt.internal_e6e56f0.output.FileOutput.writeExecutionData(FileOutput.java:53) 
at org.jacoco.agent.rt.internal_e6e56f0.Agent.shutdown(Agent.java:137) 
at org.jacoco.agent.rt.internal_e6e56f0.Agent$1.run(Agent.java:54) 
```

In Jacoco Offline, are there different threads to synchronisation/write locking for jacoco.exec in one JVM? In order to avoid concurrent writes from different threads running in parallel in one JVM, I changed some code, it can solve our problem temporarily, as below:

```
private OutputStream openFile() throws IOException {
  final FileOutputStream file = new FileOutputStream(destFile, append);
  int retries = 5;//retries 5 times
  // Avoid concurrent writes from different agents running in parallel:
  for (int t = 0; t < retries; t++)
  try{
      file.getChannel().lock();
      break;
  }catch(OverlappingFileLockException ofe){
      System.out.println("[JACOCO][RT] Got OverlappingFileLockException, re-trying.");
      try {
                Thread.sleep(1000);
            } catch (InterruptedException e) {
                //ignore
            }
  }
  return file;
 }
```

Please help to check the issue.

marchof:
FileChannel.lock() throws OverlappingFileLockException when the file is
locked from within the same JVM. This can happen when multiple JaCoCo
agents are loaded by different classloaders.

Fixes #331


marchof:
In Java 5 `OverlappingFileLockException` is not thrown, therefore unit tests fail for Java 5.

https://www.oracle.com/technetwork/java/javase/compatibility-137541.html

famod:
@marchof I've just tested this PR with [my Quarkus reproducer](https://github.com/famod/quarkus-issue-9837) and there is no `OverlappingFileLockException` anymore. :+1:

j-be:
@marchof many thanks for this.

For what its worth: I was hit by the issue using Quarkus 1.6.1 and can confirm your fix.

henrik242 (approved):
lgtm

tisoft:
I am using this branch in my quarkus project now since a week and it fixes all(*) my issues. 

Would be great if it would make it into 0.8.7.

(*) Well, all my coverage related issues, anyway.

jorsol (approved):
Thanks! confirmed the fix locally in a Quarkus application.

Please target this to 0.8.7

pilhuhn:
Please include this in the next JaCoCo release, hopefully soon

marchof:
Hi @Godin I try to pick up this PR again.

Should I cherry-pick the commit "Add org.jacoco.core.test to dependencyManagement" into a separate PR to make this PR smaller and more specific?

Godin:
> Should I cherry-pick the commit "Add org.jacoco.core.test to dependencyManagement" into a separate PR to make this PR smaller and more specific?

@marchof 👍 and done - https://github.com/jacoco/jacoco/commit/1c113ffbf7018962b8837719279ba0ca8c052de2

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/FileOutputTest.java
```diff
@@ -72,4 +78,48 @@ public void testInvalidDestFile() throws Exception {
 		controller.startup(options, new RuntimeData());
 	}
 
+	@Test(expected = OverlappingFileLockException.class)
```
Godin, line 81: We can deal with already existing usages of `@Test(expected = ...)` separately, but for new ones I prefer try/catch idiom - see https://github.com/junit-team/junit4/wiki/Exception-testing#trycatch-idiom

Godin (changes requested):

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/FileOutputTest.java
marchof: Wasn't the try/catch idiom for JUnit 3.x? Why do you prefer this? I would rather move to the JUnit 5 idiom `assertThrows(extype, sut)` once we have a Java 8 baseline.

Godin: @marchof For the reasons explained on page https://github.com/junit-team/junit4/wiki/Exception-testing#trycatch-idiom Among which: With `@Test(expected = ...)` and `ExpectedException` easy to overlook which code in test is executed and which is not due to exception, while try/catch allows to more precisely specify at which place exception is expected. When switching to `assertThrows` will you prefer to wrap entire method into `assertThrows` or just concrete part? :wink: so use of try/catch now will also kinda help to migrate on `assertThrows` one day. try/catch also allows to test message of exception if needed, and we already have places where we need and do use try/catch, thus also for consistency.

Godin (changes requested):

marchof:
@Godin Thx for the review! I applied your proposals.

Godin (approved):

PR #1059, opened 2020-06-16
Add experimental support for Java 16 class files

marchof:


marchof:
@Godin Java 16 build added as discussed

PR #1060, opened 2020-06-17
Upgrade Groovy to 3.0.6 to support latest Java versions

marchof:
Since GMavenPlus 1.9.0 Java 14 and Java 15 is supported. This version of
GMavenPlus requires Groovy 3.x, so also update Groovy to latest version.

For other JVM languages I propose that we focus on the last stable
release, so Groovy 3.x should be ok as a test target.

As Java 7 is not supported any more therefore Groovy tests are now
executed for >= Java 8

This closes #1022 being a rebased/updated version of it.

### DIFF
--- a/org.jacoco.core.test.validation.groovy/pom.xml
+++ b/org.jacoco.core.test.validation.groovy/pom.xml
@@ -26,8 +26,8 @@
   <name>JaCoCo :: Test :: Core :: Validation Groovy</name>
 
   <properties>
-    <gmaven.version>1.6.2</gmaven.version>
-    <groovy.version>2.5.8</groovy.version>
+    <gmaven.version>1.11.0</gmaven.version>
+    <groovy.version>3.0.6</groovy.version>
   </properties>
 
   <dependencies>
--- a/org.jacoco.core.test.validation/pom.xml
+++ b/org.jacoco.core.test.validation/pom.xml
@@ -90,13 +90,9 @@
           <value>7</value>
         </property>
       </activation>
-      <properties>
-        <groovy.targetBytecode>1.7</groovy.targetBytecode>
-      </properties>
       <modules>
         <module>../org.jacoco.core.test.validation.kotlin</module>
         <module>../org.jacoco.core.test.validation.java7</module>
-        <module>../org.jacoco.core.test.validation.groovy</module>
       </modules>
     </profile>
 
@@ -108,13 +104,9 @@
           <value>7</value>
         </property>
       </activation>
-      <properties>
-        <groovy.targetBytecode>1.7</groovy.targetBytecode>
-      </properties>
       <modules>
         <module>../org.jacoco.core.test.validation.kotlin</module>
         <module>../org.jacoco.core.test.validation.java7</module>
-        <module>../org.jacoco.core.test.validation.groovy</module>
       </modules>
     </profile>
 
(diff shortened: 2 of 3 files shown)


org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -26,8 +26,8 @@
   <name>JaCoCo :: Test :: Core :: Validation Groovy</name>
 
   <properties>
-    <gmaven.version>1.6.2</gmaven.version>
-    <groovy.version>2.5.8</groovy.version>
+    <gmaven.version>1.9.0</gmaven.version>
+    <groovy.version>3.0.4</groovy.version>
```
henrik242, line 30: This can be bumped to 3.0.5

org.jacoco.core.test.validation.groovy/pom.xml
```diff
@@ -26,8 +26,8 @@
   <name>JaCoCo :: Test :: Core :: Validation Groovy</name>
 
   <properties>
-    <gmaven.version>1.6.2</gmaven.version>
-    <groovy.version>2.5.8</groovy.version>
+    <gmaven.version>1.9.0</gmaven.version>
```
henrik242, line 29: GmavenPlus can be bumped to  1.10.0

henrik242 (approved):

org.jacoco.core.test.validation/pom.xml
```diff
@@ -108,13 +104,9 @@
           <value>7</value>
         </property>
       </activation>
-      <properties>
-        <groovy.targetBytecode>1.7</groovy.targetBytecode>
-      </properties>
       <modules>
         <module>../org.jacoco.core.test.validation.kotlin</module>
         <module>../org.jacoco.core.test.validation.java7</module>
-        <module>../org.jacoco.core.test.validation.groovy</module>
```
Godin, line 117: I was wondering about the statement

> As Java 7 is not supported any more therefore Groovy tests are now executed for >= Java 8

"not supported" by whom? Apparently it is about Groovy 3 - https://groovy-lang.org/releasenotes/groovy-3.0.html

> Groovy 3.0 requires JDK9+ to build and JDK8 is the minimum version of the JRE

then in `org.jacoco.doc/docroot/doc/build.html` in row `org.jacoco.core.test.validation.groovy` we should replace `7` by `8` and empty cell about `JDK 7`.

Godin (changes requested):
> This version of GMavenPlus requires Groovy 3.x

@marchof To me seems that `gmavenplus-plugin` version `1.9.0` and later actually work for latest Groovy 2.x version `2.5.13`, while compilation into bytecode version `14` and above require Groovy 3.x

```
[ERROR] Failed to execute goal org.codehaus.gmavenplus:gmavenplus-plugin:1.9.0:compile (compile) on project org.jacoco.core.test.validation.groovy:
Execution compile of goal org.codehaus.gmavenplus:gmavenplus-plugin:1.9.0:compile failed:
Target bytecode 14 requires Groovy 3.0.0-beta-2 or newer.
```

However I like your proposal

> For other JVM languages I propose that we focus on the last stable release, so Groovy 3.x should be ok as a test target.

So I'll clean it up and then do merge.

Godin (approved):

marchof:
@Godin Thanks for updating & cleaning-up!

PR #1068, opened 2020-07-07
improve english grammer

imbilltucker:
change language from "..do no match with execution data" to "..do not
match with execution data.

This is a simple change in grammer both to the code and the tests.
It can be seen when the classes that jacoco is looking at do not match the classes that were used by the tests.


marchof:
Thanks @imbilltucker !

PR #1079, opened 2020-08-06
Correct location of version in report

marchof:


Godin:
@marchof JaCoCo report shows version (`Created with`) on the right and not on the left - see https://github.com/jacoco/jacoco/blob/v0.8.5/org.jacoco.report/src/org/jacoco/report/internal/html/page/ReportPage.java#L151-L155

Screenshot 2020-08-06 at 09 55 02

or I miss something?

marchof:
Correct 🤦

PR #1098, opened 2020-09-30
Remove empty timestamp column in test report

marchof:
While reviewing #1091 I discovered that we show an always empty column in the JUnit test report.


Godin:
It was not empty in JaCoCo `0.5.2`

jacoco-0 5 2

----

And become empty in JaCoCo `0.5.3`

jacoco-0 5 3

---

So I guess relates to the migration of build from Ant to Maven.

marchof:
I don't think this information is really relevant. For it is ok to just remove it.

Godin:
Agree, I was just curious when/why it disappeared.

Godin (approved):

PR #1105, opened 2020-10-15
improve clarity of violation message

monperrus:
The current message is confusing because the two numbers compared are different (number of things versus metric)

This proposes the following change:

    classes missed count is 3, but expected maximum is 1

becomes

    unacceptable classes is 3, because missed count does not meet maximum 1

### DIFF
--- a/org.jacoco.report/src/org/jacoco/report/check/Limit.java
+++ b/org.jacoco.report/src/org/jacoco/report/check/Limit.java
@@ -190,9 +190,9 @@ public class Limit {
 	private String message(final String minmax, final BigDecimal v,
 			final BigDecimal ref, final RoundingMode mode) {
 		final BigDecimal rounded = v.setScale(ref.scale(), mode);
-		return String.format("unacceptable %s %s, because %s does not meet %s",
+		return String.format("unacceptable %s %s, because %s does not meet %s %s",
 				ENTITY_NAMES.get(entity),
-				rounded.toPlainString(), VALUE_NAMES.get(value), minmax);
+				rounded.toPlainString(), VALUE_NAMES.get(value), minmax, ref.toPlainString());
 	}
 
 	private String checkRatioLimit() {
--- a/org.jacoco.report/src/org/jacoco/report/check/Limit.java
+++ b/org.jacoco.report/src/org/jacoco/report/check/Limit.java
@@ -190,9 +190,9 @@ public class Limit {
 	private String message(final String minmax, final BigDecimal v,
 			final BigDecimal ref, final RoundingMode mode) {
 		final BigDecimal rounded = v.setScale(ref.scale(), mode);
-		return String.format("%s %s is %s, but expected %s is %s",
-				ENTITY_NAMES.get(entity), VALUE_NAMES.get(value),
-				rounded.toPlainString(), minmax, ref.toPlainString());
+		return String.format("unacceptable %s %s, because %s does not meet %s",
+				ENTITY_NAMES.get(entity),
+				rounded.toPlainString(), VALUE_NAMES.get(value), minmax);
 	}
 
 	private String checkRatioLimit() {


marchof:
Hi, not sure whether the re-wording works for all combinations (like ratio). If you fix the unit tests you should see the different wordings that result from your change.

Godin:
Hi, in addition to what was written by @marchof,
* not sure that word `unacceptable` is acceptable, because sometimes violation is expected and just change of limits was not done 😉 
* not sure that there are no flaws in the proposed message - IMO unclear what means `classes is 3` and how it is related to `missed count`, maybe it is just me, or maybe better wording possible

monperrus:
Thanks for your feedback. Here is the source of the confusion.

```
<limit implementation="org.jacoco.report.check.Limit">
<counter>CLASS</counter>
<value>MISSEDCOUNT</value>
<maximum>6</maximum>
</limit>
```

Does `<maximum>` apply to the number of classes or on the `MISSEDCOUNT` value?

Is that at most 6 classes with `MISSEDCOUNT == 0`? Or at most 0 classes with `MISSEDCOUNT <= 0`?

Godin:
> Does `<maximum>` apply to the number of classes or on the `MISSEDCOUNT` value?

Because of

```
limit implementation="org.jacoco.report.check.Limit"
```

I assume that you use Maven, so can check the documentation at https://www.jacoco.org/jacoco/trunk/doc/check-mojo.html
The same is also described in documentation for Ant at https://www.jacoco.org/jacoco/trunk/doc/ant.html

> Is that at most 6 classes with `MISSEDCOUNT == 0`? Or at most 0 classes with `MISSEDCOUNT <= 0`?

Here is how to read

```
<rule>
  <element>BUNDLE</element>
  <limits>
    <limit implementation="org.jacoco.report.check.Limit">
      <counter>CLASS</counter>
      <value>MISSEDCOUNT</value>
      <maximum>6</maximum>
    </limit>
```

for each `BUNDLE` `MISSEDCOUNT` of `CLASS`es should not be greater than `maximum` of `6`,
in other words there should be no more than 6 uncovered classes in bundle.

marchof:
And here are the counter definitions: https://www.jacoco.org/jacoco/trunk/doc/counters.html

monperrus:
>  in other words there should be no more than 6 uncovered classes in bundle.

OK that's clear.

Is there a way to say the other, "at most 0 classes with `MISSEDCOUNT <= 6`"?

Godin:
> Is there a way to say the other, "at most 0 classes with `MISSEDCOUNT <= 6`"?

Sorry, but to me not clear what you wanna achieve - `MISSEDCOUNT` of what? Class is either missed or not ( please see the link about counters provided by @marchof ), so for class `MISSEDCOUNT` is either `0` or `1`. And your `<= 6` is unclear to me.

Maybe you want "at most 0 classes with missedcount of instructions <= 6" or "at most 0 classes with missedcount of methods <= 6"  or "at most 0 classes with missedcount of lines <= 6"? they all possible.

monperrus:
> Class is either missed or not ( please see the link about counters provided by @marchof ), so for class MISSEDCOUNT is either 0 or 1.

Ahah, now I understand, per the doc, I thought `MISSEDCOUNT` was only for instructions.

> "at most 0 classes with missedcount of lines >= 6" 

Yes, exactly, this is what I'm trying to specify with Maven (which is equivalent to "all classes with missedcount of lines < 6" )

marchof:
Documentation states:

> Each limit applies to a certain counter (INSTRUCTION, LINE, BRANCH, COMPLEXITY, METHOD, CLASS) 

That is what the `counter` property is for. If that can be expressed more clearly please propose a rewording of the documentation. Also documentation contains a example with `LINE`as coverage counter.

monperrus:
OK, I need to replace `<element>BUNDLE</element>` by `<element>CLASS</element>`

```
<rule>
  <element>CLASS</element>
  <limits>
    <limit implementation="org.jacoco.report.check.Limit">
      <counter>LINE</counter>
      <value>MISSEDCOUNT</value>
      <maximum>6</maximum>
    </limit>
```

The maximum applies to the counter, not to the element.

Correct?

> If that can be expressed more clearly please propose a rewording of the documentation. 

Ack

marchof:
Correct. The `element` property defined the scope of the rule. Your example reads:

- For every class check the following limits:
  - the number of missed lines must not be more than 6

monperrus:
That's great. One more question :)

When we have the message `classes missed count is 3` on a remote CI, we don't have access to jacoco.exec.

What's the configuration parameter to list in the build log the non-covered classes?

marchof:
Messages are printed on a per limit base. It does not detail how the respective counter was calculated. There is no option to enable this.

Anyways I highly recommend to make the coverage report (not exec file) available to the developers. From my point of view the check goal alone is worthless.

monperrus:
Thanks for the answer.

Interesting, I don't have the listing of faulty classes with `<element>BUNDLE</element>`, but I have one listing with `<element>CLASS</element>`

```
[WARNING] Rule violated for class spoon.support.reflect.code.CtForImpl: lines missed count is 8, but expected maximum is 0
[WARNING] Rule violated for class spoon.support.reflect.code.CtUnaryOperatorImpl: lines missed count is 4, but expected maximum is 0
[WARNING] Rule violated for class spoon.support.reflect.code.CtCatchVariableImpl: lines missed count is 20, but expected maximum is 0
```

monperrus:
It seems I understand my confusion, sorry: we have
ELEMENT in (BUNDLE, PACKAGE, CLASS, SOURCEFILE or METHOD) 
and
COUNTER in (INSTRUCTION, LINE, BRANCH, COMPLEXITY, METHOD, CLASS)

while BUNDLE, PACKAGE, SOURCEFILE are unique for ELEMENT and  INSTRUCTION, LINE, BRANCH, COMPLEXITY are unique for COUNTER;

CLASS and METHOD are the same labels for both ELEMENT and COUNTER

marchof:
Correct, element and counter overlaps: CLASS and METHOD can server as check scope as well as a coverage counter.

monperrus:
#1107 is an attempt to clarify the doc. hope you'll like it.

PR #1111, opened 2020-10-20
adds reconnectMs for tcpclient mode

nathanwray:
This change allows tcpclient to reconnect after a socket failure. It is intended to have minimal impact to the signature and behavior of TcpClientOutput. It replaces the Thread "worker" with a TimerTask runnable. If the connection.run() exits with an IOException indicating a socket failure, the TimerTask schedules a replacement to start after a reconnectMs delay. 

Minimal changes have been made to AgentOptions to support a reconnectMs parameter. The default value disables reconnect.

The Timer is run in daemon mode and is canceled on shutdown. Since connection may initially be null, care is taken not to throw any new Exceptions (null pointer) by testing for null in shutdown() and writeExecutionData(). 

Since disconnections may be transient, they are only logged as errors if reconnection is not enabled.

This change handles the tcpserver not being active at startup as well as the tcpserver disconnecting and reconnecting during execution. 

Tests and Documentation have not been updated yet pending feedback.


marchof:
Hi @nathanwray, beside my concerns raised at the mailing list this PR seems to completely miss testing:

1) Exiting tests fail
2) No new tests that verify the new options and intended behavior

Please understand that we cannot manually test JaCoCo and therefore have unit and integration tests for alle features.

nathanwray:
Marc, if you think this has any chance of proceeding I'll add tests and address Evgeny's concerns from the mailing list. I was hoping for either encouragement or a definitive "no" before going further.

nathanwray:
The tests appear to be failing because the initialization of the Socket and Connection are async (TimerTask with a zero second delay). Experimentally, adding a sleep(100) in the junit class setup() resolves the issue although it's not very satisfying since it feels nondeterministic. I think in the real world you'd not often see an immediate call to shutdown during initialization. I'm not sure what the preferred solution would be here, but if the reconnect logic is intended to support a server that is not "up" at initialization we'd have to acknowledge that the connection object might stay null until the server becomes available. So it would not be possible to wait in the startup() method until connection is not null for example.

PR #1119, opened 2020-11-06
report-aggregate should scan also transitive dependencies for reactor projects

tobiasstadler:


marchof:
Can you please describe the use case for this? 

From a first look I see the following issues: 

* Is this a breaking change of the current behaviour?
* No tests added or adjusted

tobiasstadler:
In my project we have Module1, Module2 and Module3 (and a couple more). Module2 has Module1 as a dependency, Module3 has Module1 as a dependency. There are (integration) tests in Module3 which access classes from Module1. I want to include the coverage of classes from Module1 to be included in the report of Module3. Right now this is not possible, as only direct dependencies are scanned for classes.

tobiasstadler:
If this is something you would like to support, I can put the logic behind a config property. Also I can write test for it.

marchof:
It is currently possible. The goal is designed do be used in a separate project for aggregate report generation which lists all modules that should be considered in the aggregate report.

Using the aggregate-report goal within a regular code module is not adviced and removes exactly the flexibility to define the content of the report.

tobiasstadler:
Ok, so your solution would be to create a Module4 and add Module3 and Module1 as Dependency?

marchof:
Yes, this gives you fine grained control over:

1) The modules that should show up in the report
2) The modules that should contribute execution data (e.g. in case of pure IT test modules)

tobiasstadler:
Ok, That works, but is quite uncomfortable in our setup (Unfortunately, we have a lot of modules).

Is there a chance to get this in, if it is behind a config property?

marchof:
Ok, let's assume we consider transitive dependencies. Question:

If you have lot of modules: Which of the module(s) are you going to use for the aggregate-report goal(s)?

tobiasstadler:
Well, in our case I would like every integration test module to do its own aggregation. Actually I just want to create a report for very integration test module. But the report should include the classes of all dependencies. Classes I don't want to see are already excluded via jacoco agent parameters.

marchof:
Note that when you exclude classes from the agent they still show up in the report -- as not covered. See "Why do I see classes in the coverage report although I excluded them in the JaCoCo agent configuration?" in our [FAQ](https://www.jacoco.org/jacoco/trunk/doc/faq.html).

tobiasstadler:
Yes, but this is not a problem for us, as we use the report only to import it in SonarQube.

marchof:
While JaCoCo report generation is fast this means you create many redundant reports. Depending on the number of modules and dependencies this could result in some overhead.

Do you know what SonarQube does when the same class has data in multiple reports? What is your expectation?

laurentgo:
Can you please have a look at #1115, the end goal is similar but the approach is slightly different as it is an aggregator mojo capturing all the projects at once

tobiasstadler:
#1115, is also fine for me

tobiasstadler:
Any chance #1115 or this will be merged in the future?

PR #1136, opened 2020-12-16
Scala case class filtering

gergelyfabian:
Added support for filtering out generated methods of Scala case classes:
ScalaCaseClassGeneratedMethodsFilter.
This filters both the generated Java class `Abc` and `Abc$`.

ScalaLambdaGeneratedMethodsFilter filters for the <init> method of Scala
<=2.11 lambda classes.

Line coverage should stay the same for case classes and lambdas but
branch coverage should improve with these changes.
In validation test: number of missed branches for case class dropped from
31/0 to 0/0.

Note: canEqual() is not filtered yet as filtering it could cause line
coverage drop (detected when using jacoco with Bazel).
Note 2: only Scala 2.12 compilation is validated currently as org.jacoco.core.test.validation.scala has Scala 2.12. Part of the filtering done here is applicable only for Scala 2.11 (could consider validating both Scala versions).

Extended IFilterContext with getInterfaceNames().


gergelyfabian:
This is my first PR for Jacoco, so I'd be very grateful for any comments, proposals or guidance.

marchof:
@gergelyfabian Thanks for your contribution which looks really clean and also comes with tests! ❤️ 

As this is all spare time work please give us some time for the review.

gergelyfabian:
Thanks! Of course, take your time.
I'll use my fork in the meantime.

Marc R. Hoffmann <notifications@github.com> ezt írta (időpont: 2020. dec.
16., Sze 20:31):

gergelyfabian:
I've created also PR #1138, that could be merged before this one (only test changes), so that we can see how the (new) tests change with this PR (TDD).

marchof:
@gergelyfabian To keep the scope of this project manageable for us we decided to support only latest releases for non-Java languanges. So we can upgrade the filters and and tests for Scale 2.12 (or even 2.13?) but we will not have two versions in parallel.

marchof:
In general I recommend keeping PRs small and specific so we can discuss and decide on individual aspects.

gergelyfabian:
> @gergelyfabian To keep the scope of this project manageable for us we decided to support only latest releases for non-Java languanges. So we can upgrade the filters and and tests for Scala 2.12 (or even 2.13?) but we will not have two versions in parallel.

Ok, I see. Opened new PR #1139 with only initial Scala validation test for case classes, and will follow that with an implementation PR.

Issue #905, opened 2019-07-11
Coverage of default methods


realdadfish:
### Steps to reproduce

JaCoCo version: 0.8.3
Operating system: macOS
Tool integration: Gradle

```kotlin
interface Callback {
     fun onFoo() = Unit
     fun onBar() = Unit
     fun onBaz() = Unit
}

class CallbackImpl : Callback {
      fun onFoo() { /* ... */ }
}
```

### Expected behaviour

I'd like to see that Jacoco reports 100% coverage, because all methods in `CallbackImpl` are in fact covered and the default method implementations might technically be part of the class, but serve no purpose.

### Actual behaviour

For a test of `CallbackImpl`, Jacoco will tell me that 66% of the class' methods are not covered (and one instruction per missed method as well).

One could argue that one could test to call the other methods to check if they lead to no behaviour change, but being default implementations from some base class with no actual implementation in this class I don't see why I should unit-test them here.

Godin:
In case of `Example.kt`

```
interface Callback {
  fun onFoo() = Unit // line 2
  fun onBar() = Unit // line 3
}

class CallbackImpl : Callback { // line 6
  override fun onFoo() { } // line 7
}
```

`kotlinc` version `1.3.60` for non-overridden method `onBar` generates method
(`javap -v -p CallbackImpl.class`)

```
  public void onFoo();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=0, locals=1, args_size=1
         0: return
      LineNumberTable:
        line 7: 0

  public CallbackImpl();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokespecial #13                 // Method java/lang/Object."<init>":()V
         4: return
      LineNumberTable:
        line 6: 0

  public void onBar();
    descriptor: ()V
    flags: ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: invokestatic  #19                 // Method Callback$DefaultImpls.onBar:(LCallback;)V
         4: return
      LineNumberTable:
        line 6: 0
```

which delegates to default implementation (`javap -v -p Callback$DefaultImpls`)

```
  public static void onFoo(Callback);
    descriptor: (LCallback;)V
    flags: ACC_PUBLIC, ACC_STATIC
    Code:
      stack=0, locals=1, args_size=1
         0: return
      LineNumberTable:
[...]

PR #1137, opened 2020-12-16
KotlinDefaultMethodsFilter should also filter methods with parameters

Godin:
Filter added by #1012 in version `0.8.6` doesn't take into account methods with parameters.

See https://github.com/jacoco/jacoco/issues/905#issuecomment-746204180


mirabilos:
Can you please move this to 0.8.9 and release 0.8.8 in the meantime? I don’t use Kotlin, but I _do_ need the `assert` filtering… Thanks in advance!

alisonthemonster:
Would love to have this in jacoco 0.8.8! Any updates here?

marchof:
@alisonthemonster We're working in our spare time and completely transparently: All information and the current status of the implementation can be seen in this pull request.

PR #1139, opened 2020-12-17
Scala validation tests for case classes

gergelyfabian:
Tests for validating scala case classes.

### DIFF
--- a/org.jacoco.core.test.validation.scala/src/org/jacoco/core/test/validation/scala/targets/ScalaCaseClassTarget.scala
+++ b/org.jacoco.core.test.validation.scala/src/org/jacoco/core/test/validation/scala/targets/ScalaCaseClassTarget.scala
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2020 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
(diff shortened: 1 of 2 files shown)


org.jacoco.core.test.validation.scala/src/org/jacoco/core/test/validation/scala/targets/ScalaCaseClassTarget.scala
```diff
@@ -0,0 +1,31 @@
...
+package org.jacoco.core.test.validation.scala.targets
+
+import org.jacoco.core.test.validation.targets.Stubs.{exec, noexec, nop}
+
+/**
+ * Test target for anonymous functions.
+ */
+object ScalaCaseClassTarget {
+
+  case class Foo(a: Int, b: String, c: Set[String], d: Double) // assertPartlyCovered(31, 0)
```
gergelyfabian, line 22: 31 uncovered branches are caused by the methods generated by the Scala compiler. My next PR will fix this:

https://github.com/gergelyfabian/jacoco/blob/83c1332614a7652d2991f1be0fb6940b75d88432/org.jacoco.core.test.validation.scala/src/org/jacoco/core/test/validation/scala/targets/ScalaCaseClassTarget.scala#L22

gergelyfabian: With Scala 2.13 it's 36 branches uncovered for a single case class.

gergelyfabian:
Updated test for Scala 2.13.

gergelyfabian:
@Godin, @marchof Is this change on its own ok, or would you prefer creating already a PR together with the Scala filtering changes for case classes?

Issue #1126, opened 2020-11-25
Android "viewModelScope.launch" block partially covered for Kotlin 1.4.20


ugurcany:
### Steps to reproduce

* JaCoCo version: 0.8.6.202009150832
* Operating system: MacOS
* Steps: I'm using jacoco tool for code coverage reporting in my Android project. After I incremented the Kotlin version from 1.4.10 to 1.4.20, jacoco reports started showing coverages for `viewModelScope.launch { ... }` blocks as "partially covered". Whereas in version 1.4.10, they were fully covered.

### Expected behaviour

`viewModelScope.launch { ... }` should be marked as fully covered by jacoco for Kotlin 1.4.20.

### Actual behaviour

Jacoco report says "2 of 3 branches missed" for`viewModelScope.launch { ... }` block. So it's only partially covered:

![image](https://user-images.githubusercontent.com/9681668/100284025-91e52e00-2f7f-11eb-9adf-99e7e320350f.png)

PR #1143, opened 2020-12-18
Update filter for unsafe cast operator for Kotlin 1.4

### LINKED ISSUE #1134: Kotlin 1.4 uses NullPoinerException instead of TypeCastException, KotlinUnsafeCastOperatorFilter no more filters 'unsafe' branch
### Summary
Similarly to #754, sometimes when `as` is used for casting to nun-nullable type, kotlin adds an additional null check in byte code, which is filtered out by jacoco for kotlin 1.3, but this is no more the case for kotlin 1.4.

### Steps to reproduce
With this code snippet:
```kotlin
fun bar(f: () -> Unit) = f()

fun foo(x: Number) {
    bar {
        x as Int
    }
}
```
kotlin compiler <= 1.3 (or 1.4 with `apiLevel=1.3`) generates the following bytecode:
```
L0
    LINENUMBER 8 L0
    ALOAD 0
    GETFIELD FileKt$foo$1.$x : Ljava/lang/Number;
    DUP
    IFNONNULL L1
    NEW kotlin/TypeCastException
    DUP
    LDC "null cannot be cast to non-null type kotlin.Int"
    INVOKESPECIAL kotlin/TypeCastException.<init> (Ljava/lang/String;)V
    ATHROW
   L1
    CHECKCAST java/lang/Integer
    POP
   L2
    LINENUMBER 9 L2
    RETURN
   L3
    LOCALVARIABLE this LFileKt$foo$1; L0 L3 0
    MAXSTACK = 4
    MAXLOCALS = 1
```
and in kotlin 1.4 it generates the following:
```
L0
    LINENUMBER 8 L0
    ALOAD 0
    GETFIELD FileKt$foo$1.$x : Ljava/lang/Number;
    DUP
    IFNONNULL L1
    NEW java/lang/NullPointerException
    DUP
    LDC "null cannot be cast to non-null type kotlin.Int"
    INVOKESPECIAL java/lang/NullPointerException.<init> (Ljava/lang/String;)V
    ATHROW
   L1
    CHECKCAST java/lang/Integer
    POP
   L2
    LINENUMBER 9 L2
    RETURN
   L3
    LOCALVARIABLE this LFileKt$foo$1; L0 L3 0
    MAXSTACK = 4
    MAXLOCALS = 1
```
where `TypeCastException` is substituted with `NullPointerException`. I've found [an announcement](https://kotlinlang.org/docs/reference/compatibility-guide-14.html#unify-exceptions-from-null-checks) of this change.
Jacoco filters `TypeCastException` in the first case using `KotlinUnsafeCastOperatorFilter`, and doesn't filter in the latter case.

* JaCoCo version: 0.8.6
* Operating system: Linux and Windows
* Tool integration: Maven
* Complete executable reproducer: In [this PR](https://github.com/cqfn/diKTat/pull/623) setting `apiVersion` to 1.4 causes dramatic decrease in code coverage.
* Steps: code coverage reports are generated by jacoco and are sent to codecov. Coverage in pull request (kotlin apiVersion 1.4) is ~2.5% less than in main branch (with kotlin apiVersion 1.3).

### Expected behaviour
Same as for kotlin 1.3: branch with null check is filtered out.

### Actual behaviour
Code coverage decreases for all similar usages. Kotlin doesn't report such casts as unsafe, and even marks explicit null checks redundant (like adding `if (x != null)` to the first example).

Godin:
Fixes #1134

### DIFF
--- a/org.jacoco.core.test.validation.kotlin/pom.xml
+++ b/org.jacoco.core.test.validation.kotlin/pom.xml
@@ -25,7 +25,7 @@
   <name>JaCoCo :: Test :: Core :: Validation Kotlin</name>
 
   <properties>
-    <kotlin.version>1.3.61</kotlin.version>
+    <kotlin.version>1.4.0</kotlin.version>
   </properties>
 
   <dependencies>
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinUnsafeCastOperatorFilterTest.java
@@ -34,8 +34,8 @@ public class KotlinUnsafeCastOperatorFilterTest extends FilterTestBase {
 		final Label label = new Label();
 
 		m.visitInsn(Opcodes.DUP);
-		m.visitJumpInsn(Opcodes.IFNONNULL, label);
 		final AbstractInsnNode expectedFrom = m.instructions.getLast();
+		m.visitJumpInsn(Opcodes.IFNONNULL, label);
 		m.visitTypeInsn(Opcodes.NEW, "kotlin/TypeCastException");
 		m.visitInsn(Opcodes.DUP);
 		m.visitLdcInsn("null cannot be cast to non-null type kotlin.String");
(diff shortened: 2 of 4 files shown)


jcenturion18 (approved):

Godin:

Since for implementation of unsafe cast Kotlin compiler now uses generic `NullPointerException` I hardened filter to not run on non-Kotlin classes and to hopefully remove possibility of filtering hand-written code, i.e. so far can't imagine hand-written Kotlin code that compiles into **`dup` instruction** followed by `null`-check.

Merge of this PR will also allow further update of Kotlin to `1.4.20` in `org.jacoco.core.test.validation.kotlin` during resolution of #1126.

marchof (approved):

PR #1144, opened 2020-12-19
Change assert for CLI DumpTest to pass on non-English systems

gergelyfabian:
For a non-English system the error message may differ.

E.g. on a Hungarian system it says:
"[WARN] Kapcsolat elutasítva (Connection refused)."
This fails the test as it does not include "[WARN] Connection refused".

Change the assert to allow for the changed error message.

### DIFF
--- a/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/DumpTest.java
+++ b/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/DumpTest.java
@@ -87,7 +87,9 @@ public class DumpTest extends CommandTestBase {
 		} catch (IOException ignore) {
 		}
 
-		assertContains("[WARN] Connection refused", err);
+		// Locale independent parts of error message:
+		assertContains("[WARN]", err);
+		assertContains("Connection refused", err);
 	}
 
 	private int startMockServer() throws IOException {


marchof (approved):

marchof:
@Godin It looks like this is the only place where we assert the message of the underlying exception.

marchof:
@Godin Can you please have a look?

Godin:
A kind of mutation testing: after this change removal of `[WARN]` from
https://github.com/jacoco/jacoco/blob/29f6b48c5ae464ee47a1f4516957edc793ff24bc/org.jacoco.cli/src/org/jacoco/cli/internal/commands/Dump.java#L64
won't result in failed test, so maybe will be better to do

```
assertContains("[WARN]", err);
assertContains("Connection refused", err);
```

or maybe even change main code on something like

```
"[WARN] Connection failed: %s.%n"
```

and then assert on

```
assertContains("[WARN] Connection failed:", err);
assertContains("Connection refused", err);
```

Also, IMO in either case, easy to forget this point about locale-dependent message and not check git commits log during future changes in this test, so worth placing comment next to assert - maybe something like:

```
// Locale independent part of error message:
assertContains("Connection refused", err);
```

@marchof WDYT?

marchof:
@Godin Good point. I suggest a mixture of your proposals:

```
// Locale independent parts of error message:
assertContains("[WARN]", err);
assertContains("Connection refused", err);
```

gergelyfabian:
> @Godin Good point. I suggest a mixture of your proposals:
> 
> ```
> // Locale independent parts of error message:
> assertContains("[WARN]", err);
> assertContains("Connection refused", err);
> ```

I implemented this suggestion.

Godin (approved):

Godin:
@gergelyfabian thanks!

PR #1164, opened 2021-03-15
Fix parsing of SMAP for Kotlin 1.5

### LINKED ISSUE #1155: Support SMAP generated by Kotlin 1.5 compiler (Support JSR-45 spec)
Prior version 1.5 Kotlin compiler violate  JSR-45 specification in case of several stratum presence that was fixed in 1.5. There was a problem with `*E` presence between stratums which according to specification should be presented only at the end of SMAP (more details could be found in https://youtrack.jetbrains.com/issue/KT-37704).

Now proper SMAP output breaks JaCoCo parser with `Unexpected SMAP line: *S KotlinDebug` exception.
Repro project could be found in https://youtrack.jetbrains.com/issue/KT-44757

Godin:
Fixes #1155


marchof:
@Godin Same question: Do we have a failing validation test with Kotlin 1.5 (without this fix)?

Godin:
@marchof and the same answer 😄 yes, we already have validation test, otherwise I would add one 😉 but thank you for double-checking  👍 

Before this change when executing

```
mvn clean package -Dkotlin.version=1.5.0-M1
```

we'll get

```
Tests in error:
  execute_assertions_in_comments(org.jacoco.core.test.validation.kotlin.KotlinInlineTest): Error while analyzing org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget.
  all_branches_should_have_line_number(org.jacoco.core.test.validation.kotlin.KotlinInlineTest): Error while analyzing org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget.
  all_missed_instructions_should_have_line_number(org.jacoco.core.test.validation.kotlin.KotlinInlineTest): Error while analyzing org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget.
  last_line_in_coverage_data_should_be_less_or_equal_to_number_of_lines_in_source_file(org.jacoco.core.test.validation.kotlin.KotlinInlineTest): Error while analyzing org/jacoco/core/test/validation/kotlin/targets/KotlinInlineTarget.
```

among other differences that I currently investigate, and this test will pass after this change.

marchof (approved):

michaelbrewer:
@marchof @Godin  when will this be added to Gradle plugin?

marchof:
Probably after we released JaCoCo 0.8.7. Note that the Gradle plugin is not maintained by us.

michaelbrewer:
Looks like the new version is up for gradle is you do this:

```grooy
jacoco {
	toolVersion = "0.8.7"
}
```

Godin:
Yes - [JaCoCo `0.8.7` has been released](https://twitter.com/_godin_/status/1389728203342757891)

And yes - [Gradle's `jacoco.toolVersion` can be set to non-default value to use specific version including `0.8.7`](https://docs.gradle.org/7.0/dsl/org.gradle.testing.jacoco.plugins.JacocoPluginExtension.html),
default value will be updated - https://github.com/gradle/gradle/pull/17077

PR #1097, opened 2020-09-29
Use ASM9 API

Godin:
For following `Example.java`

```
sealed class Shape
    permits Circle
{
}

final class Circle extends Shape {
}
```

compiled as

```
javac --enable-preview --release 15 -d classes Example.java
```

by

```
openjdk version "15" 2020-09-15
OpenJDK Runtime Environment (build 15+36-1562)
OpenJDK 64-Bit Server VM (build 15+36-1562, mixed mode, sharing)
```

Execution of

```
java -jar jacococli.jar instrument classes --dest instrumented
```

succeeds when using released JaCoCo 0.8.6,
however fails when using snapshot after https://github.com/jacoco/jacoco/pull/1094

```
Exception in thread "main" java.io.IOException: Error while instrumenting /private/tmp/j/classes/Shape.class.
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrumentError(Instrumenter.java:160)
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrument(Instrumenter.java:110)
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrument(Instrumenter.java:135)
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrument(Instrumenter.java:155)
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrumentAll(Instrumenter.java:194)
        at org.jacoco.cli.internal.commands.Instrument.instrument(Instrument.java:89)
        at org.jacoco.cli.internal.commands.Instrument.instrumentRecursive(Instrument.java:78)
        at org.jacoco.cli.internal.commands.Instrument.instrumentRecursive(Instrument.java:74)
        at org.jacoco.cli.internal.commands.Instrument.execute(Instrument.java:61)
        at org.jacoco.cli.internal.Main.execute(Main.java:90)
        at org.jacoco.cli.internal.Main.main(Main.java:105)
Caused by: java.lang.UnsupportedOperationException: PermittedSubclasses requires ASM9
        at org.jacoco.cli.internal.asm.ClassVisitor.visitPermittedSubclass(ClassVisitor.java:266)
        at org.jacoco.cli.internal.asm.ClassReader.accept(ClassReader.java:673)
        at org.jacoco.cli.internal.asm.ClassReader.accept(ClassReader.java:394)
        at org.jacoco.cli.internal.core.instr.Instrumenter.instrument(Instrumenter.java:90)
[...]

PR #1165, opened 2021-03-19
Add module for Java 16 validation tests

marchof:
Also remove preview features from Java 14 validation tests.

### DIFF
--- a/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
+++ b/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
@@ -10,7 +10,7 @@
  *    Evgeny Mandrikov - initial API and implementation
  *
  *******************************************************************************/
-package org.jacoco.core.test.validation.java14.targets;
+package org.jacoco.core.test.validation.java16.targets;
 
 import static org.jacoco.core.test.validation.targets.Stubs.nop;
 
@@ -20,7 +20,7 @@ import static org.jacoco.core.test.validation.targets.Stubs.nop;
 public class InstanceofTarget {
 
 	private static void ifInstanceof(Object e) {
-		if (e instanceof String s) { // assertInstanceof()
+		if (e instanceof String s) { // assertFullyCovered(0, 2)
 			nop(s);
 		}
 	}
--- a/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
+++ b/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
@@ -15,7 +15,8 @@ package org.jacoco.core.test.validation.java16.targets;
 import static org.jacoco.core.test.validation.targets.Stubs.nop;
 
 /**
- * This target exercises pattern matching for instanceof (JEP 305).
+ * This target exercises pattern matching for instanceof
+ * (<a href="https://openjdk.java.net/jeps/394">JEP 394</a>).
  */
 public class InstanceofTarget {
 
--- a/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/RecordsTarget.java
+++ b/org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/RecordsTarget.java
@@ -13,7 +13,8 @@
 package org.jacoco.core.test.validation.java16.targets;
 
 /**
- * This target exercises records.
+ * This target exercises records
+ * (<a href="https://openjdk.java.net/jeps/395">JEP 395</a>).
  */
 public class RecordsTarget {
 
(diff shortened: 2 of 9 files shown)


Godin:
@marchof currently `InstanceofTest` and `RecordsTest` tested in following configurations

```
mvn package -Dbytecode.version=14 -Djdk.version=14
mvn package -Dbytecode.version=15 -Djdk.version=15
mvn package -Dbytecode.version=16 -Djdk.version=16
mvn package -Dbytecode.version=17 -Djdk.version=17
```

ie both as a preview feature in 14 and 15 and as a standard feature in 16+
so out of curiosity - why

> Also remove preview features from Java 14 validation tests

? I guess this is because strategy in IDEs is to support preview features of latest release and stop supporting preview features of previous releases? Like Eclipse probably already did, and IntelliJ started to warn. Right?

org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
```diff
@@ -20,7 +20,7 @@
 public class InstanceofTarget {
```
Godin, line 20: Unfortunately GitHub doesn't allow me to comment on lines too far from changed ones 😞 

----

Javadoc mentions first preview ([JEP 305](https://openjdk.java.net/jeps/305)), but not second ([JEP 375](https://openjdk.java.net/jeps/375)) and not standardisation ([JEP 394](https://openjdk.java.net/jeps/394)). So I think will be nice to update it, especially if we stop testing this as preview feature. Maybe also convert it into a link?

```diff
 /**
- * This target exercises pattern matching for instanceof (JEP 305).
+ * This target exercises pattern matching for instanceof
+ * (<a href="https://openjdk.java.net/jeps/394">JEP 305</a>).
  */
```

As a remark - to me seems that reference to JEP is kinda more insightful than [reference to JLS](https://docs.oracle.com/javase/specs/jls/se16/html/jls-15.html#jls-15.20.2).

org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/RecordsTarget.java
```diff
@@ -10,7 +10,7 @@
  *    Evgeny Mandrikov - initial API and implementation
  *
  *******************************************************************************/
-package org.jacoco.core.test.validation.java14.targets;
+package org.jacoco.core.test.validation.java16.targets;
 
 /**
  * This target exercises records.
```
Godin, line 16:
```suggestion
 * This target exercises records (<a href="https://openjdk.java.net/jeps/395">JEP 395</a>).
```

Godin (changes requested):

marchof:
@Godin Correct, the idea is to only test released features in the long run. To keep our test set maintainable we only test not-yet-released preview features primarily to prepare ourself and provide feedback to OpenJDK. I thought we discussed this once.

Godin:
>  I thought we discussed this once.

oh, I see - https://groups.google.com/g/jacoco-dev/c/edIrAG85jho/m/k4c8_ZzIDwAJ
from which I remember part

> I needed some time to wrap my mind around the fact that it is now compiling successfully.
> —> validation.java14 with JDK 15 is now compiled with to java 15 bytecode —> preview flag works :)
> So reconsidering my earlier suggestion I would actually keep switch expressions and records in validation.java14 for now.

but not

> Once features get official we can move them to the respective java version.

😆

marchof:
@Godin I don't insist on this. We can also keep testing preview features. But given the fact the the respective Java versions are already EOL and we had version specific switches in our validation code from my point of view we can drop those preview tests.

org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/InstanceofTarget.java
marchof: Good idea, done.

org.jacoco.core.test.validation.java16/src/org/jacoco/core/test/validation/java16/targets/RecordsTarget.java
marchof: Done!

marchof:
Rebased and Maven versions fixed to get a green build.

Godin:
@marchof Also this PR reminded me about absence of validation test with Sealed Classes - if it was present, we would not have missed https://github.com/jacoco/jacoco/pull/1097#issuecomment-700670200 and would have detected the problem automatically earlier 😉  Sealed Classes are preview feature in 15 and 16 and will be finalized in 17 - https://openjdk.java.net/jeps/409 And since for them there is a special handling in https://openjdk.java.net/jeps/406 will be indeed simpler to add them in/after/together with https://github.com/jacoco/jacoco/pull/1191 than alone in `org.jacoco.core.test.validation.java15`, so that `org.jacoco.core.test.validation.java17` will be the only one with `--enable-preview` 👍 

And indeed seems that in IntelliJ IDEA 2021.1 ability to use Java 14 with preview features already not available:





And seems that in upcoming Eclipse 2021-06 ability to use preview features available only for Java 16:

Godin (approved):

Godin:
@marchof oops, we almost both almost missed that it also should be added to `java17-bytecode` profile - #1194 😉

PR #1167, opened 2021-03-24
Add option exludesfile for jacoco agent

soufianenassih:
this option allows the agent to get a large list of excludes class names and avoid getting "argument too long" error as in our situation.

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/runtime/AgentOptions.java
+++ b/org.jacoco.core/src/org/jacoco/core/runtime/AgentOptions.java
@@ -12,9 +12,13 @@
  *******************************************************************************/
 package org.jacoco.core.runtime;
 
+import org.jacoco.core.tools.StringFileReader;
+
 import static java.lang.String.format;
 
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
@@ -66,6 +70,14 @@ public final class AgentOptions {
 	 * @see WildcardMatcher
 	 */
 	public static final String EXCLUDES = "excludes";
+
+	/**
+	 * File path location to get Wildcard expression for class names that should
+	 * be excluded from code coverage.
+	 *
+	 * @see WildcardMatcher
+	 */
+	public static final String EXCLUDESFILE = "excludesfile";
 
 	/**
 	 * Wildcard expression for class loaders names for classes that should be
@@ -190,7 +202,7 @@ public final class AgentOptions {
 	public static final String JMX = "jmx";
 
 	private static final Collection<String> VALID_OPTIONS = Arrays.asList(
-			DESTFILE, APPEND, INCLUDES, EXCLUDES, EXCLCLASSLOADER,
+			DESTFILE, APPEND, INCLUDES, EXCLUDES, EXCLUDESFILE, EXCLCLASSLOADER,
 			INCLBOOTSTRAPCLASSES, INCLNOLOCATIONCLASSES, SESSIONID, DUMPONEXIT,
 			OUTPUT, ADDRESS, PORT, CLASSDUMPDIR, JMX);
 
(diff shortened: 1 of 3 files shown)


marchof:
Dear @soufianenassih , can you please explain your use case for "a large list of excludes class names"?

soufianenassih:
@marchof The goal with the PR is to provide another option to pass the class exclusion value to the agent via a file
to support the cases where the size of the argument for the exclusion in the command line exceeds the limit of 128k.
Using the wildcard won't work in some cases where for example the list of classes is generated automatically.

marchof:
@soufianenassih I rephrase my question: **Why** do you need to exclude classes?

soufianenassih:
@marchof We exclude classes generated by the annotation processor that contains a lot of boilerplate code, that’s why we generated them instead of writing them manually. This is amplifying uncovered code paths. If some small part of the annotated code is uncovered, this might result in large parts of the generated code being uncovered. The result is that we see almost only generated code at the top of the coverage reports. But that’s not really useful to diagnose the problem, we'd rather see only the code that we wrote. If that is 100% covered, the generated code will be too (and even if not, that’s not our main concern but the concern of the code generator).

marchof:
@soufianenassih Then you only need to specify the exclude list when you create the report. No need to exclude classes for the agent.

Godin:
@soufianenassih exclusion of classes in agent configuration doesn't exclude them from report - see [description of agent option `excludes`](https://www.jacoco.org/jacoco/trunk/doc/agent.html)

> Except for performance optimization or technical corner cases this option is normally not required. If you want to exclude classes from the report please configure the respective report generation tool accordingly.

Since you anyway want to exclude them from report, you can **configure excludes only for report**, so you won't face "argument too long".

soufianenassih:
@Godin @marchof Thanks for the explanation: Indeed, we can exclude post-mortem at the report level but I forget to mention that we use the coverage data in SonarQube also

soufianenassih:
indeed found that SonarQube has an option for excluding.

marchof:
So in this case I close this PR.

soufianenassih:
Eventually, We still need to exclude from both instrumentation and report. the compiler cannot work correctly when jacoco instrumentation is injected into specific code. In particular:

- "implementation snippets" are a way to define some low-level behaviors using special Java code, if jacoco instrumentation is added there, those snippets are not valid anymore and lead to compiler errors.
- "tests snippets" are pieces of Java code that have a very specific shape which is asserted in the tests, if the shape is not as expected the tests can't run correctly and will fail.
@marchof  @Godin can we re-open this Pr or should I submit a new one.

Godin:
@soufianenassih which JVM version do you use? I'm asking because starting from version 9 there is following option (see output of `java -help`)

```
    @argument files
                  one or more argument files containing options
```

which allows to overcome "argument too long" without addition of new options to JaCoCo.

For example

with following `jacoco.args` file

```
-javaagent:jacoco-0.8.6/lib/jacocoagent.jar
```

execution of

```
java @jacoco.args -version
```

produces `jacoco.exec` indicating that JVM was executed with agent.

soufianenassih:
@Godin we are supporting JDK 8+.

marchof:
@soufianenassih To manage expectations: We're very reluctant with adding new config options as they complicate reproducing/debugging error reports. Also all down-stream tools have to be considered (Maven, Ant, Gradle etc). So given the fact that there is a solution with current Java versions we will not add this new option for a very specific use case.

But anyways I would be happy to help you with your case:

* Do the snippet classes follow a naming pattern that could be used for exclusion (for example specific package or marker characters in the class name)?
* What exact problem is caused by JaCoCo instrumentation?

soufianenassih:
@marchof thanks for your feedback , 
> Do the snippet classes follow a naming pattern that could be used for exclusion (for example specific package or marker characters in the class name)?

For the generated code, some have a unique Gen suffix, and some others have the no-so-unique Factory suffix. Unfortunately, that's API so it can't be changed easily.

> What exact problem is caused by JaCoCo instrumentation?

some tests will fail if jacoco instrumentation is injected into specific code. In particular:

- "implementation snippets" are a way to define some low-level behaviors using special Java code, if jacoco instrumentation is added there, those snippets are not valid anymore and lead to compiler errors.
- "tests snippets" are pieces of Java code that have a very specific shape which is asserted in the tests, if the shape is not as expected the tests can't run correctly and will fail.

more information about `implementation snippets` and `tests snippets` you can see http://www.christianwimmer.at/Publications/Simon15a/, it's a paper about this specific technique in the Graal compiler.

PR #1169, opened 2021-03-24
Add url to jacoco-maven-plugin

jakobbraun:
`jacoco-maven-plugin` does not define a project URL in the pom file. In that case, maven inherits the value of the parent pom and appends the artifact id (see maven documentation).
By that maven resolves the URL to http://jacoco-maven-plugin/, which does not exist.

This PR fixes that issue by adding an explicit URL to the project.


Godin:
@jakobbraun could you please clarify where this URL is visible?

jakobbraun:
@Godin sure.
In my case, it's visible using the Maven API. I'm using the maven API to generate a notice file that references all dependencies.
Accoring to the maven [docs](https://maven.apache.org/pom.html) it's a "nice gesture for projects users, however, so that they know where the project lives."

redcatbear:
Having this fix would be beneficial for us since we would love to have a correct list of project pages.
We want to properly acknowledge all OSS makers where we use their software as external dependencies. And part of that acknowledgement is to get the project URL right.

Please consider merging this PR.

marchof (approved):

PR #1175, opened 2021-04-12
Add maven parameter to control which reports to generate

### LINKED ISSUE #9: Report format selection
The maven report goal should allow to select the formats to be created. Maybe we can use a similar syntax as in Ant to make the report formats configurable.

troosan:
In some cases we do not need to generate all the report types. For instance when the report picked up by Sonar.
This pull request adds a new optional parameter that can be passed to the maven plugin configuration to control which reports should be generated.

Fixes #9


jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -186,12 +200,34 @@ private void executeReport(final Locale locale)
 		}
 	}
 
+	private void addFormatters(final ReportSupport support, final Locale locale)
+			throws IOException {
+		if (formats == null || formats.isEmpty()) {
```
marchof, line 205: Instead of this check can we define a default value for the formats parameter?

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -186,12 +200,34 @@ private void executeReport(final Locale locale)
...
+			support.addAllFormatters(getOutputDirectory(), outputEncoding,
+					footer, locale);
+		} else {
+			getOutputDirectory().mkdirs();
```
marchof, line 209: Related to the above comment: How can we avoid this code duplication with ReportSupport methods? Maybe replace 

    addAllFormatters(...)

with 

    addFormatters(Set<Formats> ...)

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
```diff
@@ -33,12 +34,23 @@
 public abstract class AbstractReportMojo extends AbstractMojo
 		implements MavenMultiPageReport {
 
+	static final String REPORT_XML = "XML";
```
marchof, line 37: Can we use a public Enum (top level) here instead?

jacoco-maven-plugin/src/org/jacoco/maven/AbstractReportMojo.java
marchof: Idea: implement formatter creation in Enums.

marchof:
This also fixes #9.

marchof (approved):

troosan:
@marchof thanks for the updates !

marchof:
@troosan Sure, thanks for picking up on this long outstanding feature request! I added your Github handle in the change log and file header. If you prefer your full name for attribution please feel free to add it. Also if you seen any issues with my updates please feel free to comment on it.

Godin (approved):

marchof:
Thanks @Godin !

troosan:
Thank you all @marchof @Godin

PR #1181, opened 2021-05-03
Upgrade Kotlin to 1.5.0

Godin:


org.jacoco.core.test.validation/pom.xml
```diff
@@ -90,6 +96,9 @@
           <value>7</value>
         </property>
       </activation>
+      <properties>
+        <kotlin.compiler.jvmTarget>1.6</kotlin.compiler.jvmTarget>
```
marchof, line 100: Is this on purpose? No Java 7 target?

org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinWhenExpressionTarget.kt
```diff
@@ -30,8 +30,8 @@ object KotlinWhenExpressionTarget {
     @Suppress("REDUNDANT_ELSE_IN_WHEN")
     private fun whenSealedRedundantElse(p: Sealed): Int = when (p) { // assertFullyCovered()
         is Sealed.Sealed1 -> 1 // assertFullyCovered(0, 2)
-        is Sealed.Sealed2 -> 2 // assertFullyCovered(1, 1)
-        else -> throw NoWhenBranchMatchedException() // assertNotCovered()
+        is Sealed.Sealed2 -> 2 // assertFullyCovered(0, 0)
+        else -> throw NoWhenBranchMatchedException() // assertEmpty()
```
marchof, line 34: Why don‘t we show instructions any more on this line? Or does this compile to something that our filter cannot distinguish from compiler generated code?

marchof:
@Godin Does the compilation target table in [build.html](https://www.jacoco.org/jacoco/trunk/doc/build.html) still correctly reflect the correct versions for Kotlin? Looks like Kotlin [supports](https://kotlinlang.org/docs/compiler-reference.html#jvm-target-version) a larger range. And I wonder why we stop by Java 7.

org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinWhenExpressionTarget.kt
Godin: @marchof Unfortunately without reliance on line numbers implicit `NoWhenBranchMatchedException` for exhaustive when-expression indistinguishable from explicit when using  Kotlin compiler version `1.5`.

org.jacoco.core.test.validation/pom.xml
Godin: @marchof This is kinda historical - `kotlin.compiler.jvmTarget` had default value `1.6` and in Kotlin `1.5` default was changed to `1.8`, thus this change doesn't change what we already had before.

marchof: In this case we need to update `build.html`?

Godin:
@marchof 

> Looks like Kotlin [supports](https://kotlinlang.org/docs/compiler-reference.html#jvm-target-version) a larger range. And I wonder why we stop by Java 7.

This like https://github.com/jacoco/jacoco/pull/1181#discussion_r625558104 is also historical - I think some time ago they didn't had such large range.

> Does the compilation target table in [build.html](https://www.jacoco.org/jacoco/trunk/doc/build.html) still correctly reflect the correct versions for Kotlin? 

Table should show what will be generated when `bytecode.version` not specified depending on JDK version.
And should be generated `bytecode.version` when it specified.
So you're right - both `pom.xml` and `build.html` should be adjusted.

I added such cleanup to this PR few seconds ago. Where I also made specification of `groovy.targetBytecode` consistent with `kotlin.compiler.jvmTarget` - more explicit and so IMO more readable/understandable.

org.jacoco.core.test.validation/pom.xml
Godin: See https://github.com/jacoco/jacoco/pull/1181#issuecomment-831841243

Godin:
@marchof Also Kotlin compiler doesn't support `1.7` 😉 

```
Caused by: org.jetbrains.kotlin.maven.KotlinCompilationFailureException: Compilation failure
Unknown JVM target version: 1.7
Supported versions: 1.6, 1.8, 9, 10, 11, 12, 13, 14, 15, 16
```

marchof (approved):

PR #1191, opened 2021-05-27
Add validation test for JEP 433: Pattern Matching for switch

Godin:
This requires OpenJDK 17 Early Access build with https://github.com/openjdk/jdk/pull/3863

### DIFF
--- a/org.jacoco.core.test.validation.java20/.classpath
+++ b/org.jacoco.core.test.validation.java20/.classpath
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-20">
+		<attributes>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="con" path="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER">
+		<attributes>
+			<attribute name="maven.pomderived" value="true"/>
+		</attributes>
+	</classpathentry>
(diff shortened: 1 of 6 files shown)


marchof:
@Godin Before we add new validation test project for Java 17 what about my proposed validation test cleanups up to Java 16 in #1165? Maybe we merge this first?

Godin:
@marchof 

> what about my proposed validation test cleanups up to Java 16 in #1165? Maybe we merge this first?

yup - I was about to look at it today 😉 
and this one is still draft - we can't merge it before https://github.com/openjdk/jdk/pull/3863 😉

mirabilos:
Can you please move this to 0.8.9 and release 0.8.8 in the meantime? I don’t use JDK 17, only JDK 8/11 with JRE 8, but I _do_ need the `assert` filtering… Thanks in advance!

Godin:
@marchof I updated this PR to the most recent https://openjdk.org/jeps/433

marchof:
Hi @Godin, I wonder whether we still want to provide Eclise .project and .classpath files like for the other projects (disclaimer: JavaSE-17) ist not yet supported by latest Eclipse).

Godin:
> I wonder whether we still want to provide Eclipse .project and .classpath files like for the other projects

Oops, I have them locally, but forgot to commit, because I use IntelliJ IDEA and my global `.gitignore` excludes everything starting with the dot 🤦 

> JavaSE-20 is not yet supported by latest Eclipse

I don't think that this should stop us from the addition of tests so that they can be executed at least in CI.
Thus IMO there are two options - I can create a ticket to not forget to add these files later when Eclipse will add support for JavaSE-20, or include them in this PR. Which option do you prefer?

marchof:
I would add execution environment `JavaSE-20` right now. It will be supported in 4 week with the upcoming Eclipse release.

Godin:
@marchof done!

PR #1198, opened 2021-06-13
Add experimental support for Java 18 class files

Godin:


Godin:
@marchof FYI here is quote of Remi Forax regarding support of Java 18 in ASM - https://gitlab.ow2.org/asm/asm/-/merge_requests/323 : 

> No classfile format change are currently schedule for Java 18, the only change is to add a new version V18.
>
> There are numerous changes/fixes scheduled for Java 18 https://openjdk.java.net/jeps/8267650 but none of them directly concern ASM (perhaps apart re-saying that ACC_SUPER and ACC_STRICT are dead flags)

marchof:
Hi @Godin , thanks for preparing Java 18 support! Do we want to update `org.jacoco.doc/docroot/doc/build.html` for Java 17 and 18?

marchof (approved):

Godin:
> Do we want to update org.jacoco.doc/docroot/doc/build.html for Java 17 and 18?

@marchof we also have list of supported versions for example in FAQ, so previously I was updating all these places once we add "official" support i.e. after their GA release. Will you prefer to update them all now or just `build.html` or wait for GA release?

marchof:
Adding it later if ok for me too!

PR #1221, opened 2021-09-09
Use Maven 3.8.1 in Azure Pipelines

Godin:
Seems that Maven 3.8.2 contains bug, which prevents us from using it.


marchof (approved):
Thanks!

Issue #1223, opened 2021-09-22
Missing line coverage for several examples


andre15silva:
### Steps to reproduce

* JaCoCo version:  `0.8.7.202105040129`
* Operating system: Linux
* Tool integration: Maven
* Complete executable reproducer: https://github.com/andre15silva/jacoco-bugs/tree/main/cm1
* Steps:
```bash
cd cm1/
mvn clean -Dtest="org.apache.commons.math.stat.univariate.rank.PercentileTest#testFailingExample" test
mvn jacoco:report
```

### Expected behaviour

Class `org.apache.commons.math.stat.univariate.rank.Percentile`'s coverage should include lines `143` to `151` (this excludes line `152` where an exception is thrown during the test execution).

### Actual behaviour

Coverage stops at line `142`.

![pic-selected-210922-1318-17](https://user-images.githubusercontent.com/31904862/134342101-d7a3b4b8-62fe-4655-a79c-5c1c1bb118dc.png)

We know that these lines should be covered, otherwise the exception that ends the test wouldn't be thrown.

marchof:
Are you sure `System.arraycopy()` does not throw an Exception?

andre15silva:
Hi!

> Are you sure `System.arraycopy()` does not throw an Exception?

Yes. Otherwise the test stack trace wouldn't include line `152`:
```
Running org.apache.commons.math.stat.univariate.rank.PercentileTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.011 sec <<< FAILURE!
testHighPercentile(org.apache.commons.math.stat.univariate.rank.PercentileTest)  Time elapsed: 0.009 sec  <<< ERROR!
java.lang.ArrayIndexOutOfBoundsException: 3
	at org.apache.commons.math.stat.univariate.rank.Percentile.evaluate(Percentile.java:152)
	at org.apache.commons.math.stat.univariate.rank.Percentile.evaluate(Percentile.java:105)
	at org.apache.commons.math.stat.univariate.AbstractUnivariateStatistic.evaluate(AbstractUnivariateStatistic.java:71)
	at org.apache.commons.math.stat.univariate.rank.PercentileTest.testHighPercentile(PercentileTest.java:105)
```

Adding to that, when I run the intellij-coverage agent through IntellIJ these lines are included. 
When I run the Jacoco agent either through IntellIJ or through Maven they are not included.

andre15silva:
[...]

PR #1226, opened 2021-10-01
Fix misleading FAQ

andre15silva:
A sequence of instructions between two probes may correspond to several lines.

In that case, as identified in #1223, an exception may "hide" the executed status of several lines, and not just one.


marchof (approved):

marchof:
Thanks @andre15silva !

PR #1229, opened 2021-10-05
Use Maven 3.8.3 in Azure Pipelines

### LINKED ISSUE #1218: JaCoCo Build fails with Maven 3.8.2
Since Maven 3.8.2 the JaCoCo build fails with the following error:

    org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-plugin-plugin:3.6.0:report (report) on project jacoco-maven-plugin: Execution report of goal org.apache.maven.plugins:maven-plugin-plugin:3.6.0:report failed: A required class was missing while executing org.apache.maven.plugins:maven-plugin-plugin:3.6.0:report: org/jacoco/report/check/IViolationsOutput

Therefore the Azure CI breaks, see for example PR #1217.

Godin:
Maven 3.8.1 disappeared from https://downloads.apache.org/maven/maven-3/ which causes our builds in Azure Pipelines to fail - see for example [build](https://dev.azure.com/jacoco-org/JaCoCo/_build/results?buildId=531&view=results) after merge of #1226

Instead we should be downloading from https://archive.apache.org/dist/maven/maven-3/ that contains all released versions.

Also seems that recently released Maven 3.8.3 solves the issue that we encountered with 3.8.2

Unfortunately Azure Pipelines images was not yet updated and still come with preinstalled Maven 3.8.2

Fixes #1218

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -62,7 +62,7 @@ jobs:
   - bash: |
       set -e
       mkdir .maven
-      curl -L "https://downloads.apache.org/maven/maven-3/3.8.1/binaries/apache-maven-3.8.1-bin.tar.gz" -o .maven/maven.tar.gz
+      curl -L "https://archive.apache.org/dist/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.tar.gz" -o .maven/maven.tar.gz
       tar -xzf .maven/maven.tar.gz -C .maven --strip-components 1
     displayName: Setup Maven
   - bash: |
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -551,7 +551,8 @@
                   <version>1.8</version>
                 </requireJavaVersion>
                 <requireMavenVersion>
-                  <version>3.3.9</version>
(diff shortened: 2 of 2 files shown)


.azure-pipelines/azure-pipelines.yml
```diff
@@ -62,7 +62,7 @@ jobs:
   - bash: |
       set -e
       mkdir .maven
-      curl -L "https://downloads.apache.org/maven/maven-3/3.8.1/binaries/apache-maven-3.8.1-bin.tar.gz" -o .maven/maven.tar.gz
+      curl -L "https://archive.apache.org/dist/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.tar.gz" -o .maven/maven.tar.gz
```
marchof, line 65: Any reason why we you switch to archive? This seems to work too:

https://downloads.apache.org/maven/maven-3/3.8.3/binaries/apache-maven-3.8.3-bin.tar.gz

Godin: @marchof see description of this PR 😉 that describes why I think that https://archive.apache.org/dist/maven/maven-3/ is more future-proof - this location contains versions that disappear from https://downloads.apache.org/maven/maven-3/

marchof: 🤦‍♂️

marchof (approved):

Issue #1243, opened 2021-11-03
Socket is not finally closed in org.jacoco.examples.ExecutionDataServer


ganxiyun:
### Steps to reproduce

* JaCoCo version:  0.8.8
* Operating system: `5.11.0-38-generic #42~20.04.1-Ubuntu SMP`
* Tool integration: JaCoCo API Usage Examples
* Complete executable reproducer: GitHub Repo
* Steps: 
1. Run `ExecutionDataServer` to receive execution data from TcpClient. 
```
java -cp org.jacoco.core/target/org.jacoco.core-0.8.8-SNAPSHOT.jar:org.jacoco.examples/target/org.jacoco.examples-0.8.8-SNAPSHOT.jar  org.jacoco.examples.ExecutionDataServer
```

2. After a relatively long-term period, we may see following Exceptions (reason unknown)

```
java.net.SocketException: Connection reset
        at java.net.SocketInputStream.read(SocketInputStream.java:210)
        at java.net.SocketInputStream.read(SocketInputStream.java:141)
        at java.net.SocketInputStream.read(SocketInputStream.java:224)
        at java.io.FilterInputStream.read(FilterInputStream.java:83)
        at org.jacoco.core.data.ExecutionDataReader.read(ExecutionDataReader.java:84)
        at org.jacoco.examples.ExecutionDataServer$Handler.run(ExecutionDataServer.java:83)
```

3. Check the connections 
```
netstat -aon | grep 6300
```

### Expected behaviour

No `CLOSE_WAIT` status connection

### Actual behaviour

We could see 

```
tcp6      10      0 127.0.0.1:6300          127.0.0.1:36616         CLOSE_WAIT  off (0.00/0/0)
```

The Socket in `ExecutionDataServer` hasn't been closed.

marchof:
As this example code we want to keep it simple and readable.

PR #1244, opened 2021-11-03
close socket finally in ExecutionDataServer example

ganxiyun:
Please see #1243

### DIFF
--- a/org.jacoco.examples/src/org/jacoco/examples/ExecutionDataServer.java
+++ b/org.jacoco.examples/src/org/jacoco/examples/ExecutionDataServer.java
@@ -82,12 +82,17 @@ public final class ExecutionDataServer {
 			try {
 				while (reader.read()) {
 				}
-				socket.close();
 				synchronized (fileWriter) {
 					fileWriter.flush();
 				}
 			} catch (final IOException e) {
 				e.printStackTrace();
+			} finally {
+				try {
+					socket.close();
+				} catch (IOException ex) {
+					ex.printStackTrace();
+				}
 			}
 		}
 


org.jacoco.examples/src/org/jacoco/examples/ExecutionDataServer.java
```diff
@@ -82,12 +82,17 @@ public void run() {
 			try {
 				while (reader.read()) {
 				}
-				socket.close();
 				synchronized (fileWriter) {
 					fileWriter.flush();
 				}
 			} catch (final IOException e) {
 				e.printStackTrace();
+			} finally {
```
marchof, line 90: If you nest a try/finally in the try/catch we we need only one catch block.

ganxiyun: `reader.read()` may throw `IOException`, if it happens, socket.close() will not be executed, leaving the socket connection in `CLOSED_WAIT` status and thus result in resource leakage. 

The exception catch inside finally block is needed for `socket.close()`. 

Is there a way to use only one catch while keeping the above function?

marchof: Here is my proposal:


		public void run() {
			try {
				try {
					while (reader.read()) {
					}
				} finally {
					socket.close();
				}
				synchronized (fileWriter) {
					fileWriter.flush();
				}
			} catch (final IOException e) {
				e.printStackTrace();
			}
		}

ganxiyun: @marchof Thank you for your review. 

I think there might be a small problem in the new version. 

If both reader.read() and `socket.close()` throws `IOException` (I see such case in other Closeable implementations, I don't know if it will happen in `socket.close()` or not), the `IOException` from `reader.read()` will be suppressed. 

I feel better to use a similar example in apache IOUtils. How do you think?

```
434     * Closeable closeable = null;
435     * try {
436     *     closeable = new FileReader(&quot;foo.txt&quot;);
437     *     // process closeable
438     *     closeable.close();
439     * } catch (Exception e) {
440     *     // error handling
441     * } finally {
442     *     IOUtils.closeQuietly(closeable);
443     * }
```

BTW, I think also better to move fileWriter.flush() out of the socket try block to make sure flush is executed always.

marchof: This is didactic example code. So we want it to be as simple and readable as possible.

ganxiyun: I see. I think the current implementation example is enough. I will close the PR. Thank you.

Godin: Should https://github.com/jacoco/jacoco/issues/1243 then also be closed?

ganxiyun: OK. I will close the issue.

PR #1246, opened 2021-11-07
Fix range for debug symbols of method parameters

marchof:
By inserting the probe array access code at the beginning of the method
we accidently moved the range of the debug symbols of the method
parameters after the probe array access code. Actually the method
parameters are still valid from the very beginning of the method.

This offset in the method parameter's debug symbols may confuse tools
like Jandex.

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeInserterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/instr/ProbeInserterTest.java
@@ -59,7 +59,7 @@ public class ProbeInserterTest {
 	}
 
 	@Test
-	public void testVariableStatic() {
+	public void probevar_should_be_at_position_0_for_static_method_without_parameters() {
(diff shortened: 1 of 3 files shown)


aloubyansky:
I confirm this change fixes the issue I originally reported. Great work @marchof!

aloubyansky:
I verified it by running two tests:
* https://github.com/aloubyansky/playground/tree/jacoco-jandex which loads a class (compiled w/o the `-parameters` arg) before JaCoCo instrumentation and reads method parameter names using Jandex API and then applies JaCoCo instrumentation to it and reads method parameter names again in the same way. This test fails with the 0.8.7 release and passes with this change;
* I also upgraded Quarkus locally to use the 0.8.8-SNAPSHOT version built from this branch and ran the original reproducers provided by the Quarkus user. They worked as well.

Godin (approved):

Godin:
@marchof nice fix 👍 and impressive that this was unnoticed for many years 😄 

@aloubyansky thank you for reporting this and testing! ❤️

PR #1259, opened 2021-11-25
Add filter for Jetpack Compose generated bytecode

### LINKED ISSUE #1208: Jacoco reports low coverage with Compose tests
### Steps to reproduce

* JaCoCo version:  0.8.7
* Operating system: macOS Big Sur 11.4
* Tool integration: Gradle
* Complete executable reproducer: https://github.com/arcadefire/JacocoComposeSample
* Steps: 
  - Checkout to the repo above and execute the command: `./gradlew clean JacocoTestReport`

### Expected behaviour
Creating a simple composable function displaying a text and a test covering it that checks that the text was displayed should produce 100% test coverage.

### Actual behaviour
The coverage for such a simple case is particularly low, especially the _conditional_ coverage, even though the function contains no conditional logic.

Screenshot 2021-07-04 at 21 00 02

Also, looking at the function's coverage details, the yellow diamonds report a number of untested conditional branches.

Screenshot 2021-07-04 at 21 00 55

lsuski:
This PR fixes #1208.
I'm aware that adding such thing to general purpose jvm coverage tool seems strange.  Ideally there should be some extension mechanism in JaCoCo which allows adding `IFilter` in similar way as `excludes` are added. This could be utilized by android gradle plugin. However as there is no such feature and there is high need from android developers using Jetpack Compose (at least from those who use some coverage quality gates like my company)  I decided to submit this PR.


Avenire:
@lsuski Tried your fork in my Compose project and it works great with instrumentation tests. 
Unfortunately Roboelectric unit tests report now 0% coverage, where with 0.8.7 Jacoco, they at least report somewhat correct instruction coverage. Would it be possible to support both?

Pushed an example project showing the problem in case you weren't aware of the issue:
https://github.com/Avenire/jacoco-compose-test

lsuski:
I'll check, in my projects I have some roboletric tests and coverage is quite fine (there are some uncovered lines even though they should be but the coberage is about 70%)

sarn0ld:
Hey @lsuski ! Awesome work! 
What's the easiest way to use your fork in our projects?

lsuski:
@sarn0ld you have to build my fork on your own. I didn't publish it publicly. I use my company internal maven repo

G00fY2:
Any chance this will get merged soon? For me it is not clear whats blocking this PR?

lsuski:
I think that the main problem is that JaCoCo is meant to be Jvm general coverage tool and should not have any specific frameworks filters and JetPack Compose is a niche from JaCoCo perspective. IMHO JaCoCo should have some extension mechanism which will allow registering such filters on demand and then there will be no need to add Compose filter here, it will be added by Android Gradle Plugin or any other third party Plugin

PR #1271, opened 2022-01-05
Add missing dependency (fix #1269)

### LINKED ISSUE #1269: Cannot run a concurrent build with maven
On a clean master (no previous build in the local repository):
```
mvn -T 4 verify
```
leads to
```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.2:copy (default) on project org.jacoco.agent: Unable to find artifact.: Could not find artifact org.jacoco:org.jacoco.agent.rt:jar:all:0.8.8-SNAPSHOT
```

I suspect a missing dependency which is hidden when using a single threaded build because of the module ordering.

See https://github.com/apache/maven-mvnd/issues/558

gnodet:


gnodet:
This is clearly not sufficient to allow parallel builds.  There are additional problems in test dependencies, as the `jacoco-maven-plugin` is used in various places during tests but not referenced as a dependency.

marchof:
Also some of the (IT) tests allocate ports. Probably this will result in random failure with parallel builds.

Out of curiosity: What are you working on to require parallel JaCoCo builds?

gnodet:
> Also some of the (IT) tests allocate ports. Probably this will result in random failure with parallel builds.
> 
> Out of curiosity: What are you working on to require parallel JaCoCo builds?

I'm just working on `mvnd` and had the report the jacoco was failing with it (see https://github.com/apache/maven-mvnd/issues/558), so I investigated a bit.

marchof (dismissed):

PR #1275, opened 2022-01-19
faq: added a Q&A each for different class with same name, and using the right runtime classifier jar for -javaagent

alkuma:
Based on the discussion with @marchof on https://groups.google.com/u/1/g/jacoco/c/NZdGZ9eO_CY


alkuma:
The jdk13 build failure 

https://dev.azure.com/jacoco-org/50fdda27-7f92-436e-9e67-3f48d7c2b6db/_apis/build/builds/718/logs/36

Seems to be a job setup issue where the VM isn't able to locate java. Anybody know what needs to be done to make it work?

PR #1277, opened 2022-01-24
Agent support report coverage continuously

lydiazhao0712:
In some scenarios, such as using coverage for traffic screening, it is necessary to continuously collect coverage data without stopping the service. Therefore, the above changes are made to support the agent to continuously report coverage data


marchof:
Hi, thanks for the pull request! Due to our limited resources we will not add more configuration options that add additional complexity. Your requirement can be solved outside JaCoCo with a TCP server/client implementation which periodically requests dumps.

To give you an idea what maintaining new features means here is what is missing in your pull request:

* Unit and integration tests for the new feature
* New attributes in Ant Task incl. Tests
* New properties in Maven Task incl Tests.
* Documentation of new properties (java agent, Ant task, Maven goal)

PR #1300, opened 2022-04-05
Use ASM Bill Of Materials (BOM)

andrioli:
Since version 9.3 ASM provide a Bill Of Materials (BOM). So we can remove a few lines from our dependency management.


marchof (approved):

Godin:
@andrioli thanks!

Issue #1100, opened 2020-10-06
Using --xml option of jacococli report fails if the folder does not exist


wlad:
### Steps to reproduce

* JaCoCo version: 0.8.6.202009150832
* Operating system: Ubuntu 19.10 eoan
* Tool integration: CLI 
* Complete executable reproducer: see failing step "Generate SDK IT Coverage Report" in CircleCI build:  https://app.circleci.com/pipelines/github/ehrbase/ehrbase/1273/workflows/6689b8e5-f276-4c14-ab24-c8676dfab5b0/jobs/15005

* Steps: 

  - Call `jacococli report` to generate a report with `--xml` option (below example is shortened for readability)
```
java -jar ~/jacoco/lib/jacococli.jar report test-coverage/target/jacoco-sdk-it-coverage.exec \
  --classfiles api/target/classes/ \
  --sourcefiles api/src/main/java/ \
  --html test-coverage/sdk-it-coverage-report \
  --xml test-coverage/sdk-it-coverage-report/jacoco.xml \
  --name "EHRbase Code Coverage w/ SDK Integration Tests"
```

### Expected behaviour
```
[INFO] Loading execution data file /home/wlad/projects/ec/ehrbase/test-coverage/target/jacoco-sdk-it-coverage.exec.
[INFO] Analyzing 519 classes.
```
No exception.
And **jacoco.xml** file created inside **test-coverage/sdk-it-coverage-report/**

> NOTE:
> when using `--xml jacoco.xml`  without a path, the file is created outside the html folder, which is not desired behaviour
> So intuitively I'd expect that either --xml can take a filename with relative or absolut path or place the file inside html folder defined by --html if filename is provided without a path

### Actual behaviour
```
[INFO] Loading execution data file /home/circleci/projects/test-coverage/target/jacoco-sdk-it-coverage.exec.
[INFO] Analyzing 519 classes.
Exception in thread "main" java.io.FileNotFoundException: test-coverage/sdk-it-coverage-report/jacoco.xml
(No such file or directory)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:187)
	at org.jacoco.cli.internal.commands.Report.createReportVisitor(Report.java:148)
[...]

PR #1303, opened 2022-04-23
A little enhancement of report

the-end-sea-of-the-world:
Deal with issue [#1100](https://github.com/jacoco/jacoco/issues/1100). Check the parent file of csv and xml, create if it does not exist.

### DIFF
--- a/org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
+++ b/org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
@@ -144,11 +144,21 @@ public class Report extends Command {
 		final List<IReportVisitor> visitors = new ArrayList<IReportVisitor>();
 
 		if (xml != null) {
+			String path = xml.getPath().substring(0, xml.getPath().lastIndexOf("\\"));
+			File file = new File(path);
+			if (!file.exists()){
+				boolean dr = file.mkdirs();
+			}
 			final XMLFormatter formatter = new XMLFormatter();
 			visitors.add(formatter.createVisitor(new FileOutputStream(xml)));
 		}
 
 		if (csv != null) {
+			String path = csv.getPath().substring(0, csv.getPath().lastIndexOf("\\"));
+			File file = new File(path);
+			if (!file.exists()){
+				boolean dr = file.mkdirs();
+			}
 			final CSVFormatter formatter = new CSVFormatter();
 			visitors.add(formatter.createVisitor(new FileOutputStream(csv)));
 		}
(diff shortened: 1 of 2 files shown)


marchof:
Hi @the-end-sea-of-the-world, welcome to the JaCoCo project and thanks for your contribution!

We try to provide a regression test for every feature. What about adding (or extending) the [existing tests](https://github.com/jacoco/jacoco/blob/master/org.jacoco.cli.test/src/org/jacoco/cli/internal/commands/ReportTest.java) to validate your code?

Also your implementation will not work on Unix/Linux system where the path separator is a forward slash /.

Instead of parsing Strings the `java.io.File` API allows to determine the parent folder. We already use this in our code base, for example [here](https://github.com/jacoco/jacoco/blob/master/org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/output/FileOutput.java#L52).

marchof (changes requested):
* Add unit test
* Use `java.io.File` API to also support Unix path separators

the-end-sea-of-the-world:
Thanks for your suggestion, I will try to modify it.

the-end-sea-of-the-world:
Hello, I have changed the code to use `java.io.File` API and added JUnit tests and a Maven project demo used in the test to  generate a report.

marchof:
@the-end-sea-of-the-world thanks for the update!

To be honest I don't think we will maintain the demo code in the JaCoCo project. May I ask you to remove it?

org.jacoco.cli.test/src/org/jacoco/cli/internal/MainTest.java
```diff
@@ -78,4 +78,88 @@ public void should_not_print_any_output_when_quiet_option_is_given()
 		assertNoOutput(err);
 	}
 
+	@Test
+	public void should_not_print_error_message_when_parent_file_of_xml_does_not_exist() throws Exception {
```
marchof, line 82: The exitsting `ReportTest` is the better place to test the `Report` class.

org.jacoco.cli/src/org/jacoco/cli/internal/commands/Report.java
```diff
@@ -140,15 +140,28 @@ private void writeReports(final IBundleCoverage bundle,
...
+
+	/**
+	 * Added blocks:
+	 * Check the parent files of xml and csv, create them if they do not exist.
+	 */
 	private IReportVisitor createReportVisitor() throws IOException {
 		final List<IReportVisitor> visitors = new ArrayList<IReportVisitor>();
 
 		if (xml != null) {
+			final File folder = xml.getParentFile();
```
marchof, line 152: Please extract a private method `createParent()` to avoid code duplication.

marchof:
Also as you can see [here](https://github.com/jacoco/jacoco/pull/1303/checks) the build fails as the source code has some formatting issues. Our development setup with Eclipse auto-formatting is described [here](https://www.jacoco.org/jacoco/trunk/doc/environment.html).

the-end-sea-of-the-world:
Thanks for your advice. There are a lot of things in school these days. I will modify the code on the weekend.

the-end-sea-of-the-world:
@marchof all the problems have been solved. Is there any other one?

Issue #993, opened 2020-01-11
Allow filtering of specific methods in Java


yuvalprtn:
### Scenario
* JaCoCo version: 0.8.5
* Operating system: Mac OS
* Tool integration: Gradle
* Description of your use case: Java code coverage 

### Current Behavior
Only class level exclusion is available 
### Wanted Behavior
Ability to filter out methods by name pattern. In this case we will be able to remove toString, hashCode, equals as well as specific methods if we choose not to pass through coverage for these.
Specifically toString is quite redundant in our POV.
Using annotations or comments will require us to go class by class for exclusion while a rule of some sort is what we are looking for
### Possible Workarounds
None that we know of

nikolim:
Is this topic still relevant? I think this is a good idea and would like to work with colleagues on this issue

better-712:
I have made a pull request for this issue, it can pass through coverage for toString ,if needed, hashCode, equals can also be added , they can be implemented similarly。

better-712:
I have made a pull request for this issue, it can pass through coverage for toString ,if needed, hashCode, equals can also be added , they can be implemented similarly。Thanks for DanielKraft

better-712:
By name pattern, we can also filter out other methods.

marchof:
To manage expectations: We will not include configurable filtering to JaCoCo as this makes testing and support even harder due to another dimension of complexity. Please understand that this is a pure side project and we need to keep the scope manageable for our limited time budget.

PR #1320, opened 2022-05-29
Allow filtering of specific methods in Java #993

better-712:
it can pass through coverage for toString method for need


marchof:
Filtering all methods called "toString" is not a feature we consider to include.

better-712:
Thanks for your reply, Should the issue #993 also be closed?



---Original---
From: "Marc R. ***@***.***&gt;
Date: Mon, May 30, 2022 00:57 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [jacoco/jacoco] Allow filtering of specific methods in Java #993 (PR #1320)




 
Filtering all methods called "toString" is not a feature we consider to include.

better-712:
Thanks for your reply, should the issue #993 also be closed?



---Original---
From: "Marc R. ***@***.***&gt;
Date: Mon, May 30, 2022 00:57 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [jacoco/jacoco] Allow filtering of specific methods in Java #993 (PR #1320)




 
Filtering all methods called "toString" is not a feature we consider to include.

marchof:
@better-712 I assume that #993 would be a config option (not hard coded like in this PR). But as we decided not to make filtering configurable the issue should be closed. Thanks for the hint!

PR #1334, opened 2022-06-17
Agent should not open `java.base` module to unnamed module of the application class loader

### LINKED ISSUE #1328: Agent opens access to java.base to unnamed module
We're migrating to JDK 17 and found that we were unable to run tests with JaCoCo to verify code was not accessing encapsulated Java internals in the `java.base` package.

### Steps to reproduce

* JaCoCo version: 0.8.8
* Operating system: macOS 10.4
* Tool integration: Command-line Agent

Given class `IllegalAccess.java` and JDK 17:
```
import java.lang.Throwable;

public class IllegalAccess {
	public static void main(String[] args) throws Exception {
		Throwable.class.getDeclaredMethod("setCause", Throwable.class).setAccessible(true);
	}
}
```

Run without Jacoco and note the exception:
```
$ java IllegalAccess
Exception in thread "main" java.lang.reflect.InaccessibleObjectException: Unable to make final void java.lang.Throwable.setCause(java.lang.Throwable) accessible: module java.base does not "opens java.lang" to unnamed module @a803f94
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354)
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
	at java.base/java.lang.reflect.Method.checkCanSetAccessible(Method.java:199)
	at java.base/java.lang.reflect.Method.setAccessible(Method.java:193)
	at IllegalAccess.main(IllegalAccess.java:5)
```

Run with Jacoco and no exception is thrown:
```
$ java -javaagent:/Users/dannyt/Downloads/jacoco-0.8.8/lib/jacocoagent.jar IllegalAccess
```

### Expected behaviour

The agent should not cause runtime side effects to code under test. This in particular makes it impossible to verify production code doesn't perform illegal accesses at test time causing this code to fail when deployed.

### Actual behaviour

The agent adds `java.base` to the exported modules for the unnamed module causing the entire classpath to be granted access.

Godin:
Fixes #1328


Godin:
@marchof foreseeing your question about why this is still a draft 😉  because the code requires polishing and the addition of in-code comments, I plan to work on this a bit later. But of course, feel free to review to provide your suggestions/comments - as usual, they are more than welcome 🤗

marchof:
First of all, thanks for digging deeper into this! I was too fast in closing this issue.

Just some high level remarks:

* Can we add a (integration) test case which demonstrates the original problem reported in #1328?
* Maybe add your comments and spec references from #1328.
* As the initialization is specific to the `InjectedClassRuntime`I wonder whether the code should move there for better encapsulation. Then the `PreMain`would look like this:

```
private static IRuntime createRuntime(final Instrumentation inst) throws Exception {
	if (InjectedClassRuntime.isSupported()) {
		return ModifiedSystemClassRuntime.createFor(inst);
	}
	return ModifiedSystemClassRuntime.createFor(inst, "java/lang/UnknownError");
}
```

org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
```diff
@@ -58,31 +61,46 @@ public static void premain(final String options, final Instrumentation inst)
...
+			}
+		}.loadClass(InjectedClassRuntime.class.getName())
+				.getConstructor(Class.class, String.class)
+				.newInstance(Object.class, "$JaCoCo"));
+		final Object module = getModule(runtime.getClass());
+		if (module.equals(getModule(PreMain.class))) {
+			throw new IllegalStateException();
 		}
-
 		Instrumentation.class.getMethod("redefineModule", //
```
marchof, line 104: Maybe for consistency and better readability this code can be extracted to a private method like `getModule()`

Godin:
> Can we add a (integration) test case

Sure.

First of all for such a test case we need a "stable" private JDK API 😆 e.g. `Throwable.setCause` mentioned in #1328 is not presented in JDK 11 - was added in JDK 12, so instead I propose to use

```
Class.forName("java.lang.Module").getDeclaredConstructors()[0].setAccessible(true);
```

because IMO less likely that any of the constructors of `java.lang.Module` will become public than the removal or change of signature of `java.lang.Throwable.setCause`

Secondly on 9 <= JDK < 16 we either need `--illegal-access=deny` or

```
class Example {
    public static void main(String[] args) throws Exception {
        System.out.println(Object.class.getModule().isOpen("java.lang", Example.class.getModule()));

        Class.forName("java.lang.Module").getDeclaredConstructors()[0].setAccessible(true);
    }
}
```

leads to

```
true
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by Main (file:/private/tmp/j/) to constructor java.lang.Module(java.lang.ModuleLayer,java.lang.ClassLoader,java.lang.module.ModuleDescriptor,java.net.URI)
WARNING: Please consider reporting this to the maintainers of Main
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

instead of

```
false
Exception in thread "main" java.lang.reflect.InaccessibleObjectException: Unable to make java.lang.Module(java.lang.ModuleLayer,java.lang.ClassLoader,java.lang.module.ModuleDescriptor,java.net.URI) accessible: module java.base does not "opens java.lang" to unnamed module @11739fa6
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:357)
        at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297)
        at java.base/java.lang.reflect.Constructor.checkCanSetAccessible(Constructor.java:188)
        at java.base/java.lang.reflect.Constructor.setAccessible(Constructor.java:181)
```

`--illegal-access=deny` is not needed on JDK 16 because of https://openjdk.org/jeps/396, and not needed on JDK 17 because of https://openjdk.org/jeps/403, which unfortunately also states

> We expect to remove the --illegal-access option entirely in a future release.

on JDK < 9 `--illegal-access=deny` leads to

```
Unrecognized option: --illegal-access=deny
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

so for simplicity I propose to test only JDK >= 16.

Finally to test the agent we need to
* either introduce a new type of tests
* or add it to `jacoco-maven-plugin.test` which are already slow
* or add it to `org.jacoco.ant.test` which for simplicity I propose to do

----

And I'll work on addressing other comments a bit later.

marchof:
> So for simplicity I propose to test only JDK >= 16

👍

org.jacoco.core/src/org/jacoco/core/runtime/InjectedClassRuntime.java
```diff
@@ -43,19 +54,76 @@ public class InjectedClassRuntime extends AbstractRuntime {
...
 	public InjectedClassRuntime(final Class<?> locator,
 			final String simpleClassName) {
+		this(null, locator, simpleClassName);
+	}
+
+	private InjectedClassRuntime(final ClassLoader classLoader,
+			final Class<?> locator, final String simpleClassName) {
+		this.classLoader = classLoader == null
+				? this.getClass().getClassLoader()
+				: classLoader;
```
Godin, line 64:
```suggestion
		this(this.getClass().getClassLoader(), locator, simpleClassName);
	}

	private InjectedClassRuntime(final ClassLoader classLoader,
			final Class<?> locator, final String simpleClassName) {
		this.classLoader = classLoader;
```

marchof:
Hi @Godin, now as it properly encapsulated in `InjectedClassRuntime` I think it is ok to create the classloader in `createFor()` only (so leave out the second commit).

But in `createFor()` I would add some implementation comments why the class loader is created.

org.jacoco.agent.rt/src/org/jacoco/agent/rt/internal/PreMain.java
```diff
@@ -58,58 +55,23 @@ public static void premain(final String options, final Instrumentation inst)
 	private static IRuntime createRuntime(final Instrumentation inst)
```
marchof, line 55: To further encapsulate the implementation details of `InjectedClassRuntime` I would add a static method `isSupported()` to this class. The code here in PreMain then reduces to:

	private static IRuntime createRuntime(final Instrumentation inst)
			throws Exception {
		if (InjectedClassRuntime.isSupported()) {
			return InjectedClassRuntime.createFor(inst);
		} else {
			return ModifiedSystemClassRuntime.createFor(inst,
					"java/lang/UnknownError");
		}
	}

org.jacoco.core/src/org/jacoco/core/runtime/InjectedClassRuntime.java
```diff
@@ -43,19 +54,76 @@ public class InjectedClassRuntime extends AbstractRuntime {
...
 
+	/**
+	 * Creates a new {@link InjectedClassRuntime}.
+	 *
+	 * @param instrumentation
+	 *            instrumentation interface
+	 * @return new runtime instance
+	 */
+	public static IRuntime createFor(final Instrumentation instrumentation)
+			throws Exception {
```
marchof, line 78: Eclipse shows me a warning for undocumented `Exception`

Godin:
@marchof I believe that all your previous comments were addressed, so could you please review?

org.jacoco.ant.test/src/org/jacoco/ant/IllegalReflectiveAccessTarget.java
```diff
@@ -0,0 +1,46 @@
...
+		try {
+			Class.forName("java.lang.invoke.MethodHandles$Lookup")
+					.getField("ORIGINAL");
+		} catch (ClassNotFoundException e) {
+			// Java < 7
+			return;
+		} catch (NoSuchFieldException e) {
+			// Java < 16
+			return;
+		}
```
marchof, line 30:
```suggestion
		try {
			Class.forName("java.net.UnixDomainSocketAddress");
		} catch (ClassNotFoundException e) {
			// Java < 16
			return;
		}
```

Java 16 detection can be simplified with checking for `UnixDomainSocketAddress` which was introduced with this release.

org.jacoco.core/src/org/jacoco/core/runtime/InjectedClassRuntime.java
```diff
@@ -48,6 +56,71 @@ public InjectedClassRuntime(final Class<?> locator,
...
+				if (!name.equals(InjectedClassRuntime.class.getName())
+						&& !name.equals(Lookup.class.getName())) {
+					return super.loadClass(name, resolve);
+				}
+				final InputStream resourceAsStream = getResourceAsStream(
+						name.replace('.', '/') + ".class");
+				final byte[] bytes;
+				try {
+					bytes = InputStreams.readFully(resourceAsStream);
+				} catch (IOException e) {
```
marchof, line 94:
```suggestion
				} catch (final IOException e) {
```
We currently make all locals final, Eclipse will add this if I save the file.

marchof (approved):
Wow very impressed by all the technical details of this fix! I only have two minor cleanup proposals.

Godin:
>  I only have two minor cleanup proposals.

@marchof actually maybe I myself have one more 😆 I have a feeling that maybe better to move `createFor` method into `PreMain` class: strictly speaking `InjectedClassRuntime` is not in the `internal` package and thus part of our public API, whereas `createFor` is just a particular way to instantiate `InjectedClassRuntime` specific to our agent with hard-coded values such as `java.lang` and `$JaCoCo`. WDYT?

marchof:
> > I only have two minor cleanup proposals.
> 
> @marchof actually maybe I myself have one more 😆 I have a feeling that maybe better to move `createFor` method into `PreMain` class: strictly speaking `InjectedClassRuntime` is not in the `internal` package and thus part of our public API, whereas `createFor` is just a particular way to instantiate `InjectedClassRuntime` specific to our agent with hard-coded values such as `java.lang` and `$JaCoCo`. WDYT?

I fully agree, except that `PreMain` isn't really a better place neither. What about a new class in package `org.jacoco.agent.rt.internal` which handles the aspect of class loader creation, module opening and loading specific classes? This may open opportunities for unit testing too. But I need to find a proper name and API.

Godin:
> But I need to find a proper name and API

Here are some ideas about name:
* `InjectedClassRuntimeInstantiator`
* we already have some classes with prefix `Support` - `InjectedClassRuntimeSupport`
* well maybe strictly speaking these are not really suitable names, but anyway - `InjectedClassRuntimeFactory` or `InjectedClassRuntimeBuilder`

Also what do you think about just moving the method into `PreMain` and merging now to unlock this bug fix, and then working on the improvement of `PreMain` separately?

marchof:
I wonder whether this `*Support` tool should be specific about the `InjectedClassRuntime` at all. I would encapsulate the aspect of loading/instantiating a arbitrary class in its own class loader. This way it may become unit testable.

pzygielo:
Seems that the change b865890d566c55832def986bb800b0b2f98200de causes failure https://ci.eclipse.org/glassfish/job/hk2/job/PR-819/3/console.

I've not investigated fully, it's possible that upgrade proposed in
- https://github.com/eclipse-ee4j/glassfish-hk2/pull/819

needs some additional modifications.

Reporting just for cross-linking, while it's relatively fresh.

Godin:
@pzygielo Thank you for letting us know! ❤️ I created #1425

pzygielo:
Evgeny, thanks for checking.

PR #1335, opened 2022-06-18
Add experimental support for Java 20 class files

Godin:


marchof (approved):
Same changes as in #1264, thanks!

Looks like we should create a scheduled action for byte code upgrades 😅

dougxc:
Is there a rough ETA as to when the 0.8.9 release will be made? It will be necessary for use of JaCoCo on JDK 20 without forcing people to rely on the 0.8.9.202211161124 snapshot build.

PR #1348, opened 2022-08-04
Add `skipPoms` configuration parameter

hgschmie:
Setting `<skipPoms>true</skipPoms>` will skip the `prepare-agent` and
`prepare-agent-it` execution in POM (non-code) projects. This is
desirable, however the default is `false` to be backwards compatible.


marchof:
Hi @hgschmie thanks for your contribution! Can you please elaborate what problem in what scenario is solved with this?

hgschmie:
I am running this plugin as part of the build process and it keeps executing the agent goal (which is of no use) on my pom modules. Lots of other plugins either ignore or can be configured to ignore pom projects.

PR #1352, opened 2022-08-14
Exclude internal packages from published JavaDoc

marchof:
While we already exclude internal packages for our all-in-one Javadoc we still publish them for the individual Maven modules. See for example: https://javadoc.io/doc/org.jacoco/org.jacoco.core/0.8.8/index.html

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -360,6 +360,7 @@
           <configuration>
             <quiet>true</quiet>
             <detectOfflineLinks>false</detectOfflineLinks>
+            <excludePackageNames>*.internal</excludePackageNames>
           </configuration>
         </plugin>
         <plugin>


Godin (changes requested):
@marchof I'm wondering how you discovered this - did you encounter some users who were misguided by published javadoc for internal packages? and believed that they are part of our public API, while this is not the case.

I agree that this might be confusing. And even after this change users still can discover internal packages in JARs with sources. Should we tweak this too? And even if so, then internal packages will anyway show up in IDEs during auto-completions/searches? I'm wondering can the addition of `module-info.class` help?

Also at the same time I like that we write documentation/comments even for internal packages. And seems that this change also disables doclint for internal packages:

using JDK 21 execution of `mvn clean verify` before this change produces

```
[INFO] --- javadoc:3.0.1:jar (attach-javadocs) @ org.jacoco.cli ---
[WARNING] Javadoc Warnings
[WARNING] /Users/evgeny.mandrikov/projects/jacoco/jacoco/org.jacoco.cli/src/org/jacoco/cli/internal/CommandHandler.java:35: warning: no description for @param
[WARNING] * @param parser
[WARNING] ^
[WARNING] /Users/evgeny.mandrikov/projects/jacoco/jacoco/org.jacoco.cli/src/org/jacoco/cli/internal/CommandHandler.java:36: warning: no description for @param
[WARNING] * @param option
[WARNING] ^
[WARNING] /Users/evgeny.mandrikov/projects/jacoco/jacoco/org.jacoco.cli/src/org/jacoco/cli/internal/CommandHandler.java:37: warning: no description for @param
[WARNING] * @param setter
[WARNING] ^
[WARNING] 3 warnings
[INFO] Building jar: /Users/evgeny.mandrikov/projects/jacoco/jacoco/org.jacoco.cli/target/org.jacoco.cli-0.8.9-SNAPSHOT-javadoc.jar
[INFO]
[INFO] --- source:2.1.2:jar-no-fork (attach-sources) @ org.jacoco.cli ---
```

after

```
[INFO] --- javadoc:3.0.1:jar (attach-javadocs) @ org.jacoco.cli ---
[INFO]
[INFO] --- source:2.1.2:jar-no-fork (attach-sources) @ org.jacoco.cli ---
```

Thus I'm wondering can we somehow get them both? or this doesn't add much value, wdyt? Maybe by grouping packages - see https://maven.apache.org/plugins/maven-javadoc-plugin/examples/group-configuration.html and https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#group ?

But most importantly this change disables the generation of javadoc jar for artifacts such as `org.jacoco.cli` that contain only internal packages, and so I'm almost certain that then we will not be able to publish to Maven Central - see https://central.sonatype.org/publish/requirements/#supply-javadoc-and-sources To overcome this we probably should supply javadoc jar containing at least something - see https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-compiler-embeddable/1.9.10/kotlin-compiler-embeddable-1.9.10-javadoc.jar that contains only `META-INF/MANIFEST.MF`.

marchof:
@Godin I didn't had a specific use case for this. Thanks for your analysis. It looks like it causes more trouble than it solves. So I propose to not follow this up.

Issue #1378, opened 2022-10-21
AssertFilter does not ignore the assert statement


morganga:
### Steps to reproduce

* JaCoCo version:  0.8.8.20220405071
* Operating system: Mac OS
* Tool integration: Maven
* Complete executable reproducer: https://github.com/morganga/jacoco-bug
* Steps: 
mvn clean install
the build fails (100% coverage is required) and
the target/site/jacoco/net.gcvs.jacoco/App.java.html report does not indicate full coverage of the assert statements.

### Expected behaviour
coverage of assert statement is considered complete, line in coverage report is green.

### Actual behaviour
assert statement contains two branches and "1 of 2 branches missed", line in coverage report is yellow.

(In case of exceptions provide FULL STACKTRACE)

morganga:
I've implemented a solution in a pull request: https://github.com/jacoco/jacoco/pull/1379
Please take a look at the pull request, which contains screenshots of the implemented solution and a very similar alternative solution.

ice1000:
Cool!!!!!!!!!!!!!!!!!!!!!!

ice1000:
Hi, I believe that you need to write some tests :) The maintainers told me they only accept fully tested code.

morganga:
I'm not sure what else they want.  I've modified the existing test expectations.
In the comment of the AssertTarget.java assert line, and in the unit test AssertFilterTest.java

oers:
I can confirm the same problems with 0.8.8 and a reduced code coverage.

barakugav:
would love to see progress on this issue

ice1000:
> would love to see progress on this issue

Me too!

Issue #324, opened 2015-06-12
Accounting of assert statements


Dridi:
I'm using jacoco with @SonarSource's sonarqube and one thing that bothers me is the actual accounting of `assert` statements in coverage reports.

They basically get a 100% line coverage, and 50% branch coverage.

Both metrics don't make sense since assert are not supposed to have any side effect, and thus do anything useful. You also don't want a 100% branch coverage, since it means corrupted state. And of course you can disable them at run time.

They should not be treated as code instructions.

Other than that, jacoco rocks!

marchof:
Hi, thanks for your feedback! This is a known issue and needs to be solved with filters one day (Issue #15). I added it to the list of use cases here: https://github.com/jacoco/jacoco/wiki/FilteringOptions

Dridi:
Sorry for the duplicate, it didn't show up when I searched open issues for "assert".

I will subscribe to #15 :)

marchof:
No problem -- it actually wasn't on the use case list for filtering till your your report.

PR #1379, opened 2022-10-21
ignore the assert statement's conditional branch.  (#1378)

morganga:
assert statement coverage should be considered complete if the assert statement was visited. There's no need to track multiple branches at the assert statement.

This is in agreement with the previously reported issue #324, which suggested "They should not be treated as code instructions."

I've extended the ignored instructions to include the assert statement check.  This will make coverage complete with or without assertions enabled.


morganga:
before fix:
image
after fix:
image

morganga:
An alternative solution to AssertFilter: 62

`				output.ignore(start, ((JumpInsnNode) cursor).label);
`
results in this report:
image

ice1000:
I love that

Cybermite:
Going to bump this and see if we can get it looked at and merged?

This has been the main reason why our team hasn't migrated off of clover.

gredler:
Does anyone know why the JDK 21 and 22 builds failed for this PR? Can somebody trigger the builds again? The failed builds no longer exist on Azure, so there's no telling what happened, AFAICT.

ice1000:
This would be really nice to have!

ice1000:
I opened #1816. All tests pass.

cstmgl:
Was there any explanation why this was never merged? I think after 4 years it's obvious it will not be merged but I'm just wondering if this is really something the community doesn't want to have.

I'm not an expert in Java so I wonder if it's really wrong that we should ignore the assert negative branches

PR #1400, opened 2023-02-13
Upgrade Groovy to 3.0.15

Godin:


marchof (changes requested):
In `org.jacoco.core.test.validation/pom.xml` line 354 there is another reference to "Groovy 3.0.13" in a comment.

Godin:
@marchof nice catch 👍 fixed

marchof (approved):

PR #1409, opened 2023-03-05
Separate class loading from runtime

marchof:
@Godin This is roughly what I had in mind when I was proposing to separate class loading from the runtime implementation. If you agree with the idea I would do the remaining cleanup tasks.

### DIFF
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2023 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.agent.rt.internal;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+
+import org.jacoco.core.test.validation.JavaVersion;
+import org.junit.AssumptionViolatedException;
+import org.junit.Test;
+
+/**
+ * Unit tests for {@link AgentModule}.
+ */
+public class AgentModulTest {
+
+	@Test
+	public void isSupported_should_return_false_before_Java9() {
+		if (JavaVersion.current().isAtLeast("9")) {
+			throw new AssumptionViolatedException(
+					"Modules available after Java 9");
+		}
+		assertFalse(AgentModule.isSupported());
+	}
+
+	@Test
+	public void isSupported_should_return_true_for_Java9_or_greater() {
+		if (JavaVersion.current().isBefore("9")) {
+			throw new AssumptionViolatedException(
+					"Modules only available after Java 9");
+		}
+		assertTrue(AgentModule.isSupported());
+	}
+
+	@Test
+	public void should_only_load_classes_in_scope() throws Exception {
+		if (JavaVersion.current().isBefore("9")) {
+			throw new AssumptionViolatedException(
+					"Modules only available after Java 9");
+		}
+
+		AgentModule am = new AgentModule();
+		Class<? extends Target> targetclass = am
+				.loadClassInModule(TargetImpl.class);
+		Target t = targetclass.getDeclaredConstructor().newInstance();
+
+		assertNotSame(this.getClass().getClassLoader(),
+				t.getClass().getClassLoader());
+		assertNotSame(this.getClass().getClassLoader(),
+				t.getInnerClassInstance().getClass().getClassLoader());
+		assertSame(this.getClass().getClassLoader(),
+				t.getOtherClassInstance().getClass().getClassLoader());
+	}
+
+	public interface Target {
+
+		Object getInnerClassInstance();
+
+		Object getOtherClassInstance();
+
+	}
+
+	public static class TargetImpl implements Target {
+
+		static class Inner {
+		}
+
+		public Object getInnerClassInstance() {
+			return new Inner();
+		}
+
+		public Object getOtherClassInstance() {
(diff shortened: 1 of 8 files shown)


org.jacoco.core/src/org/jacoco/core/runtime/AgentModule.java
```diff
@@ -0,0 +1,151 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.jacoco.core.internal.InputStreams;
+
+/**
+ * An isolated class loader and distinct module to encapsulate JaCoCo runtime
+ * classes. This isolated environments allows to specifically open JDK packages
```
Godin, line 27:
```suggestion
 * classes. This isolated environment allows to specifically open JDK packages
```

org.jacoco.core/src/org/jacoco/core/runtime/AgentModule.java
```diff
@@ -0,0 +1,151 @@
...
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.runtime;
```
Godin, line 13: Should be in package `org.jacoco.agent.rt.internal`

org.jacoco.core/src/org/jacoco/core/runtime/AgentModule.java
```diff
@@ -0,0 +1,151 @@
...
+	/**
+	 *
+	 * @param instrumentation
+	 *            service provided to the agent by the Java runtime
+	 * @param classInPackage
+	 *            example class of the package to open
+	 * @throws Exception
+	 *             if package cannot be opened
+	 */
+	public void openPackage(final Instrumentation instrumentation,
```
Godin, line 91: Please add a description for this method.

Godin (changes requested):
@marchof actually modulo names and comments we came up to the same implementation - I just didn't had time to push mine :wink: so yes - please go ahead with the remaining cleanup 👍

org.jacoco.core/src/org/jacoco/core/runtime/AgentModule.java
marchof: Done. But now we have a dependency on internal core class `org.jacoco.core.internal.InputStreams`. Technically this is no problem as the agent is shaded anyways. Or should I better duplicate the code and avoid the dependency?

Godin: IMO ok as it is.

marchof:
@Godin I addressed you comments and added a unit test

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.junit.Assert.assertTrue;
+
+import org.jacoco.core.test.validation.JavaVersion;
+import org.junit.AssumptionViolatedException;
+import org.junit.Test;
+
+/**
+ * Unit tests for {@link AgentModule}.
+ */
+public class AgentModulTest {
```
Godin, line 27: Typo

```suggestion
public class AgentModuleTest {
```

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
```diff
@@ -0,0 +1,92 @@
...
+		}
+		assertTrue(AgentModule.isSupported());
+	}
+
+	@Test
+	public void should_only_load_classes_in_scope() throws Exception {
+		if (JavaVersion.current().isBefore("9")) {
+			throw new AssumptionViolatedException(
+					"Modules only available after Java 9");
+		}
```
Godin, line 52: AFAICS this test passes even without this restriction of Java version, so I'm wondering why restrict?

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+		AgentModule am = new AgentModule();
+		Class<? extends Target> targetclass = am
+				.loadClassInModule(TargetImpl.class);
+		Target t = targetclass.getDeclaredConstructor().newInstance();
+
+		assertNotSame(this.getClass().getClassLoader(),
+				t.getClass().getClassLoader());
+		assertNotSame(this.getClass().getClassLoader(),
+				t.getInnerClassInstance().getClass().getClassLoader());
```
Godin, line 62:
```suggestion
		assertSame(t.getClass().getClassLoader(),
				t.getInnerClassInstance().getClass().getClassLoader());
		assertNotSame(this.getClass().getClassLoader(),
				t.getInnerClassInstance().getClass().getClassLoader());
```

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	@Test
+	public void isSupported_should_return_true_for_Java9_or_greater() {
+		if (JavaVersion.current().isBefore("9")) {
+			throw new AssumptionViolatedException(
+					"Modules only available after Java 9");
+		}
+		assertTrue(AgentModule.isSupported());
+	}
```
Godin, line 45: Maybe simply

```
	@Test
	public void isSupported_should_return_false_before_Java9() {
		assertEquals(!JavaVersion.current().isBefore("9"),
				AgentModule.isSupported());
	}
```

to avoid marking tests as unsupported?

Godin (changes requested):

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
marchof: Removed.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/AgentModulTest.java
marchof: 👍

Godin (approved):
Thanks! ❤️

PR #1404, opened 2023-02-20
Upgrade ECJ from 3.12.1 to 3.32.0

Godin:
Version `4.6.1` of previously used artifact `org.eclipse.jdt:ecj` was actually version of Eclipse IDE.

This version supports source/target from Java 1.1 up to 19.

Also note that starting from version `3.27.0` ECJ requires Java 11 for execution.

PR #1413, opened 2023-03-15
Require at least JDK 11 for the build

Godin:


Godin:
@marchof I wanted to add a validation test for JEP 432 Record Patterns, but to format them
* either need to update Eclipse Formatter to the latest version, which like Eclipse Compiler requires JDK 11 (see https://github.com/jacoco/jacoco/pull/1404)
* or use a different version of formatter in different modules
* or disable automatic formatting in the corresponding module

I propose simply go with option 1 and require JDK 11 to start Maven. After this is done, I'll do a separate PR for the update of Eclipse Formatter.

marchof (approved):

PR #1414, opened 2023-03-17
Upgrade spotless-maven-plugin to 2.35.0 and Eclipse JDT Formatter to 4.27

Godin:


marchof:
I'm fine with the implementation. I just have doubts whether we really want use the formatter for latest preview features in the long run. But we can revert this once it is blocking us.

marchof:
Did you report `IndexOutOfBoundsException` to JDT project?

PR #1415, opened 2023-03-22
Add validation test for JEP 432: Record Patterns

Godin:


marchof:
Hi @Godin , a couple of lines are `assertPartlyCovered()`, which I would not expect from the execution paths. Any idea why?

Godin:
@marchof not just an idea - I had a look into the current compiler implementation and bytecode that it currently produces 😉

Partial coverage and missed branches are due to how Record Patterns are compiled and seems that we will need to implement filter(s) for a few things:

To support split-compilation (Record can be changed and recompiled after Pattern) deconstruction of each Record component is compiled into a call of component accessor inside of try-catch block
```
Point p = (Point) o;
try { x = p.x(); } catch (Throwable t) { throw new MatchException(t.toString(), t); }
```

Also enhanced-for-statement under the hood compiles into `for` with `switch`, and `switch` is exhaustive so in its turn contains an additional branch that also throws `MatchException`.

marchof (approved):

PR #1422, opened 2023-03-31
Fix validation test for Java 21

Godin:
OpenJDK 21 EA b15 and b16 produce different class files for `RecordPatternsTarget.java` because of
https://github.com/openjdk/jdk/commit/138cdc9283ae8f3367e51f0fe7e27833118dd7cb

Here is diff between outputs of `javap -v -p RecordPatternsTarget.class`:

```diff
          line 34: 57
          line 35: 72
+         line 34: 78
          line 37: 92
```

Bytecode instructions at offset 78 correspond to the following exception handler generated for the `case` at line 34:

```
        78: astore_1
        79: new           #24                 // class java/lang/MatchException
        82: dup
        83: aload_1
        84: invokevirtual #26                 // Method java/lang/Throwable.toString:()Ljava/lang/String;
        87: aload_1
        88: invokespecial #30                 // Method java/lang/MatchException."<init>":(Ljava/lang/String;Ljava/lang/Throwable;)V
        91: athrow
```

### DIFF
--- a/org.jacoco.core.test.validation.java20/src/org/jacoco/core/test/validation/java20/RecordPatternsTest.java
+++ b/org.jacoco.core.test.validation.java20/src/org/jacoco/core/test/validation/java20/RecordPatternsTest.java
@@ -12,6 +12,8 @@
  *******************************************************************************/
 package org.jacoco.core.test.validation.java20;
 
+import org.jacoco.core.test.validation.JavaVersion;
+import org.jacoco.core.test.validation.Source.Line;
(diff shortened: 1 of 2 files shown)


Godin:
@marchof note that I updated `JDK21_URL` in AzurePipelines and it now points to OpenJDK EA b16 - that's why this PR is green, and so without change-back or merge of this PR other further builds will be red.

marchof (approved):

PR #1438, opened 2023-05-15
Maven 3.9.2 should not produce warnings for jacoco-maven-plugin

Godin:


Godin:
This is superseded by #1466 and #1468

PR #1449, opened 2023-05-15
Refactor re-calculation of stackmap frames

marchof:
* Try to make code more readable by adding a utility class to track types and variable positions on locals
* Re-use locals array to avoid object creations

----

While trying to understand the new code in #893 I was thinking how to refactor the code to make it more comprehensible (at least for me).

@Godin Please feel free to reject this proposal if you don't think it helps.

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
@@ -171,30 +171,45 @@ class ProbeInserter extends MethodVisitor implements IProbeInserter {
 					"ClassReader.accept() should be called with EXPAND_FRAMES flag");
 		}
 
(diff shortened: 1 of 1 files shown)


org.jacoco.core/src/org/jacoco/core/internal/instr/ProbeInserter.java
```diff
@@ -171,30 +171,45 @@ public final void visitFrame(final int type, final int nLocal,
...
-		while (idx < nLocal) {
-			newLocal[newIdx++] = local[idx++];
+
+		mv.visitFrame(type, newLocals.typeCount, newLocals.types, nStack,
+				stack);
+	}
+
+	private final Locals newLocals = new Locals();
+
+	private static class Locals {
```
marchof, line 200: Make this class package private and add unit tests

PR #1466, opened 2023-05-23
Require Java 8 for jacoco-maven-plugin

### LINKED ISSUE #1435: Maven 4 compatibility / Maven 3.9 warnings
Maven 3.9.2 reports the following issues:

```
[WARNING]  * org.jacoco:jacoco-maven-plugin:0.8.10
[WARNING]   Plugin issue(s):
[WARNING]    * Plugin is a Maven 2.x plugin, which will be not supported in Maven 4.x
[WARNING]    * Plugin mixes multiple Maven versions: [3.0, 2.0.2]
[WARNING]    * Plugin should declare these Maven artifacts in `provided` scope: [org.apache.maven:maven-repository-metadata:3.0, org.apache.maven:maven-artifact:3.0]
[WARNING]    * Plugin depends on plexus-container-default, which is EOL
```

Godin:
This change will
* unblock #1462 
* allow to solve #1435 differently than proposed in #1438

### DIFF
--- a/jacoco-maven-plugin/.classpath
+++ b/jacoco-maven-plugin/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/J2SE-1.5">
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8">
 		<attributes>
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
--- a/jacoco-maven-plugin/pom.xml
+++ b/jacoco-maven-plugin/pom.xml
@@ -32,6 +32,12 @@
     <maven>3.0</maven>
   </prerequisites>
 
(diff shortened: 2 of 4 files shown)


Godin:
@marchof if you're agree with this change, then I'll finalize this PR by updating changelog

marchof:
@Godin I agree with this change!

marchof (approved):

irbian:
Looks like this change has broken some of our oldest builds based on java 7

Normally called as org.jacoco:jacoco-maven-plugin:prepare-agent , without specifying version, now uses 0.8.11 instead of 0.8.10

Would this have happened anyway if the versioning used were semver? We are looking where exactly is decided what jacoco version is used

Godin:
> Would this have happened anyway if the versioning used were semver? 

According to https://semver.org/ 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. 

😉 

----

Your build got broken because

> Normally called as org.jacoco:jacoco-maven-plugin:prepare-agent , without specifying version

this is not normal - see https://maven.apache.org/guides/mini/guide-configuring-plugins.html

> Important Note: Always define the version of each plugin used to guarantee build reproducibility.

irbian:
> initial development
image

I see XD

> this is not normal - see https://maven.apache.org/guides/mini/guide-configuring-plugins.html

I agree, it surprise me too. I guess that using semver wouldn't had made a difference? without specifying it would have started using 1.0.0 anyway? I have to check where the version to use was decided

PR #1468, opened 2023-05-24
Maven 3.9.2 should not produce warnings for jacoco-maven-plugin

Godin:


org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -22,8 +22,11 @@ <h2>Snapshot Build @qualified.bundle.version@ (@build.date@)</h2>
 
 <h3>New Features</h3>
 <ul>
+  <li>Maven 3.9.2 should not produce warnings for jacoco-maven-plugin
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/1468">#1468</a>).</li>
   <li>jacoco-maven-plugin now requires at least Java 8
-      (GitHub <a href="https://github.com/jacoco/jacoco/issues/1466">#1466</a>).</li>
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/1466">#1466</a>,
+      <a href="https://github.com/jacoco/jacoco/issues/1468">#1468</a>).</li>
```
marchof, line 29: Is it intentional that the PR is listed twice? If so maybe the commit message should be extended to explain the dependency on Java 8.

Godin: @marchof Yes - it is intentional in order to express the link between these changes in the changelog.
I updated the commit message.

marchof (approved):

PR #1472, opened 2023-06-05
Add filter for exhaustive switch expression

### LINKED ISSUE #771: Add filter for exhaustive switch-expression

Godin:
Before this change for the following `Example.java`

```
enum Example {
    A, B;

    static int example(Example e) {
        return switch (e) {
        case A -> 1;
        case B -> 2;
        };
    }

        example(A);
        example(B);
    }
}
```

using

```
openjdk version "17.0.2" 2022-01-18 LTS
OpenJDK Runtime Environment Zulu17.32+13-CA (build 17.0.2+8-LTS)
OpenJDK 64-Bit Server VM Zulu17.32+13-CA (build 17.0.2+8-LTS, mixed mode, sharing)
```

execution of

```
javac Example.java -d classes
java -javaagent:jacocoagent.jar -cp classes Example
java -jar jacococli.jar report jacoco.exec --classfiles classes --sourcefiles . --html report
```

produces

before

before

and after this change

after

after

---

Fixes #771


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/ExhaustiveSwitchFilter.java
```diff
@@ -0,0 +1,110 @@
...
+				dflt = ((TableSwitchInsnNode) start).dflt;
+				labels = ((TableSwitchInsnNode) start).labels;
+			} else {
+				return;
+			}
+
+			cursor = skipToLineNumberOrInstruction(dflt);
+			if (cursor == null) {
+				return;
+			} else if (cursor.getType() == AbstractInsnNode.LINE) {
```
marchof, line 63: In consistency with other early returns I would use else here.
```suggestion
			}
			if (cursor.getType() == AbstractInsnNode.LINE) {
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/ExhaustiveSwitchFilter.java
```diff
@@ -0,0 +1,110 @@
...
+				if (line != ((LineNumberNode) cursor).line) {
+					return;
+				}
+				cursor = skipNonOpcodes(cursor);
+			}
+			if (cursor == null || cursor.getOpcode() != Opcodes.NEW) {
+				return;
+			}
+			if ("java/lang/MatchException"
```
marchof, line 72: What about some inline documentation or extraction into two method with speaking names.

    String type = ((TypeInsnNode) cursor).desc;
    if (!isJava5Exception(desc) && !isJava21Exception(desc) {
      return;
    }

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/ExhaustiveSwitchFilter.java
```diff
@@ -0,0 +1,110 @@
...
+						"()V");
+			} else {
+				return;
+			}
+			nextIs(Opcodes.ATHROW);
+			if (cursor == null) {
+				return;
+			}
+			output.ignore(dflt, cursor);
+			final HashSet<AbstractInsnNode> replacements = new HashSet<AbstractInsnNode>();
```
marchof, line 93: Use more general type `Set` here?

marchof (approved):
Great work again! I left some formal comments you might want to consider.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/ExhaustiveSwitchFilter.java
marchof: That's what I wanted to say. The truth is in the code 😅

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/ExhaustiveSwitchFilter.java
marchof: Ok, one day in future we will replace it with `var` anyways 😄

nioncode:
@Godin should this already handle multi-line switch expressions? I still get missed branch warnings in my code, where I have something like:

```java

enum Example {
    A, B;

    static int example(Example e) {
        return switch (e) {
        case A -> {
            if (false) {
              throw new InvalidStateException();
            }
            yield 1
        };
        case B -> {
            if (false) {
              throw new InvalidStateException();
            }
            yield 2;
        };
    }

        example(A);
        example(B);
    }
}
```

Godin:
@nioncode for the following `Example.java` which is the same example as yours after fixing compilation errors

```
enum Example {
    A, B;

    static int example(Example e) {
        return switch (e) {
        case A -> {
            if (false) {
              throw new IllegalStateException();
            }
            yield 1;
        }
        case B -> {
            if (false) {
              throw new IllegalStateException();
            }
            yield 2;
        }
        };
    }

        example(A);
        example(B);
    }
}
```

using

```
openjdk version "21" 2023-09-19
OpenJDK Runtime Environment (build 21+35-2513)
OpenJDK 64-Bit Server VM (build 21+35-2513, mixed mode, sharing)
```

execution of

```
javac Example.java -d classes
java -javaagent:jacoco-0.8.12/lib/jacocoagent.jar -cp classes Example
java -jar jacoco-0.8.12/lib/jacococli.jar report jacoco.exec --classfiles classes --sourcefiles . --html report
```

produces the following report

Screenshot 2024-04-29 at 11 38 05

Screenshot 2024-04-29 at 11 37 46

nioncode:
This is true. Sorry, I ran this through gradle and it seems it does not pick up the latest version by default.

Just for others, if you use the jacoco plugin in gradle, you can force a specific version via:
```
plugins {
  id "jacoco" // note: don't specify a version here, specify it in the jacoco block below
}

jacoco {
  toolVersion = "0.8.12"
}
```

Everything works fine as is, sorry for the noise.

PR #1473, opened 2023-06-05
Add filter for Record patterns

Godin:
Before this change for the following `Example.java`

```
record Example(int x, int y) {

  static void example_instanceof(Object o) {
    if (o instanceof Example(int x, int y)) {
      System.out.println(x + ", " + y);
    }
  }

  static void example_switch(Object o) {
    switch (o) {
      case Example(int x, int y) -> System.out.println(x + ", " + y);
      default -> System.out.println();
    }
  }

    example_instanceof(new Object());
    example_instanceof(new Example(1, 2));
    example_switch(new Object());
    example_switch(new Example(1, 2));
  }
}
```

using

```
openjdk version "21-ea" 2023-09-19
OpenJDK Runtime Environment (build 21-ea+25-2212)
OpenJDK 64-Bit Server VM (build 21-ea+25-2212, mixed mode, sharing)
```

execution of

```
javac Example.java -d classes
java -javaagent:jacocoagent.jar -cp classes Example
java -jar jacococli.jar report jacoco.exec --classfiles classes --sourcefiles . --html report
```

produces

before

before

and after this change

after

after


marchof (approved):

marchof:
@Godin While reviewing this PR I noticed that record pattern matching and switch pattern matching has been finalized in Java 21. Shouldn't we move it to a new project `org.jacoco.text.validation.java21`?

Issue #973, opened 2019-11-12
Skip coverage for methods generated for inline classes


mkrussel77:
THIS IS A BUG TRACKER ONLY. FOR QUESTIONS PLEASE CHECK FAQ OR USE FORUM:

http://www.jacoco.org/jacoco/trunk/doc/faq.html

https://groups.google.com/forum/?fromgroups=#!forum/jacoco

Please understand that
ISSUES WITHOUT FOLLOWING INFORMATION WILL BE CLOSED WITHOUT COMMENTS!
Thank you for filling feature request!

### Scenario

* JaCoCo version: (from right bottom corner of JaCoCo report) 0.8.5.201910111838
* Operating system: Mac OS
* Tool integration: Maven / Ant / CLI / API (for others please report to respective project) Gradle
* Description of your use case: (detailed description or executable reproducer, e.g. GitHub repo)

Kotlin generates several methods for inline classes similar to how it does for data classes. 

- equals()
- hashcode()
- toString()
- getter for wrapped value

Would like inline classes to behave like data classes.  I know inline classes are experimental so adding it at this point might cause breakage in the future.

### Current Behaviour
All of these methods are reported as unused

### Wanted Behaviour
The methods to not be reported as unused

### Possible Workarounds
Use data classes instead of inline classes

Godin:
For methods generated in data classes Kotlin compiler does not generate line numbers - see #689

Whereas seems that for inline classes Kotlin compiler `1.3.60` generates methods that always have line number `1` - for example compilation of following `Example.kt` (`kotlinc Example.kt`)

```
data class DataClassExample( // line 1
  val p: String
)

inline class InlineClassExample( // line 5
  val p: String
)
```

results in following methods (`javap -v -p InlineClassExample.class`)

```
  public java.lang.String toString();
    descriptor: ()Ljava/lang/String;
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: getfield      #11                 // Field p:Ljava/lang/String;
         4: invokestatic  #72                 // Method "toString-impl":(Ljava/lang/String;)Ljava/lang/String;
         7: areturn
      LineNumberTable:
        line 1: 0

  public int hashCode();
    descriptor: ()I
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=1, locals=1, args_size=1
         0: aload_0
         1: getfield      #11                 // Field p:Ljava/lang/String;
[...]

PR #1475, opened 2023-06-07
Add filter for Kotlin inline value classes

Godin:
Fixes https://github.com/jacoco/jacoco/issues/973


marchof (approved):

marchof:
@Godin I just approved this. Looks like there are now 2 reviewers required?

PR #1476, opened 2023-06-07
Add test to catch cases when compilers generate wrong line numbers

Godin:
Initially in #1475 I wanted to add a validation test that verifies resolution of https://youtrack.jetbrains.com/issue/KT-35553 , but turns out that javac version `20` violates it on `RecordPatternsTarget` 😆 So decided to extract the addition of such test into dedicated PR.


marchof:
As [JDK-8307814](https://bugs.openjdk.org/browse/JDK-8307814) has just been fixed I propose to wait some more days until this can be merged without the special handling in `RecordPatternsTest`.

PR #1483, opened 2023-06-16
Remove unused parts of configuration for AppVeyor

Godin:
* Nowadays AppVeyor sets `M2_HOME` pointing on preinstalled Maven, so the execution of downloaded `apache-maven-3.3.9\bin\mvn` actually leads to the execution of preinstalled.

* Also `mvn` uses `JAVA_HOME`, so its addition to `PATH` is not needed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,17 +1,11 @@
 cache:
-  - C:\Users\appveyor\apache-maven-3.3.9 -> appveyor.yml
   - C:\Users\appveyor\.m2 -> **\pom.xml
 
 install:
-  - if not exist C:\Users\appveyor\apache-maven-3.3.9 (
-      curl -LsS "http://www.apache.org/dyn/closer.cgi?action=download&filename=maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip" > apache-maven-3.3.9-bin.zip &&
-      unzip apache-maven-3.3.9-bin.zip -d C:\Users\appveyor
-    )
   - SET JAVA_HOME=C:\Program Files\Java\jdk11
-  - SET PATH=C:\Users\appveyor\apache-maven-3.3.9\bin;%JAVA_HOME%;%PATH%
 
 build_script:
-  # Maven 3.3.9 requires Java >= 7, but generation of Javadocs requires Java <= 6 (https://github.com/jacoco/jacoco/issues/110)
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -1,17 +1,12 @@
...
   - C:\Users\appveyor\.m2 -> **\pom.xml
 
 install:
-  - if not exist C:\Users\appveyor\apache-maven-3.3.9 (
-      curl -LsS "http://www.apache.org/dyn/closer.cgi?action=download&filename=maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip" > apache-maven-3.3.9-bin.zip &&
-      unzip apache-maven-3.3.9-bin.zip -d C:\Users\appveyor
-    )
   - SET JAVA_HOME=C:\Program Files\Java\jdk11
-  - SET PATH=C:\Users\appveyor\apache-maven-3.3.9\bin;%JAVA_HOME%;%PATH%
+  - SET PATH=%JAVA_HOME%;%PATH%
```
marchof, line 6: What is the usage of `%JAVA_HOME%` on `PATH`? Shouldn't that be `%JAVA_HOME%\bin` or what packaging do we have here?

Godin: @marchof nice finding! 👍  locally to control Java version I do both - set `JAVA_HOME` and prepend `$JAVA_HOME/bin` to `PATH`, so tried to do the same in https://github.com/jacoco/jacoco/commit/4fca868d0c1ddd36457af8420136c1f11e56a1ba but apparently did a typo 😅 which went thru review 😆 😉 and CI run because `mvn` relies on `JAVA_HOME`, so actually this modification of `PATH` is unused.

Given the above I'm wondering what is better

```suggestion
```

or

```suggestion
  - SET PATH=%JAVA_HOME%\bin;%PATH%
```

and whether one of those changes should be in this PR or in a separate one since it is a bit unrelated to the subject of this PR?

Godin: The first option will be consistent with what we do for Azure Pipelines where we only set `JAVA_HOME`
https://github.com/jacoco/jacoco/blob/08f4f83e5df8b7a996aa83249f5377310e493494/.azure-pipelines/azure-pipelines.yml#L84

marchof: Not setting the path at all is what I prefer. Just add a comment about it in the commit message.

Godin: @marchof 👍 done

PR #1512, opened 2023-09-27
`Opcodes.RET` should be processed by `visitVarInsn` instead of `visitInsn`

Godin:
See
* https://docs.oracle.com/javase/specs/jvms/se17/html/jvms-6.html#jvms-6.5.ret
* https://asm.ow2.io/javadoc/org/objectweb/asm/MethodVisitor.html#visitVarInsn(int,int)
* https://gitlab.ow2.org/asm/asm/-/blob/ASM_9_5/asm/src/main/java/org/objectweb/asm/ClassReader.java?ref_type=tags#L2408-2409


marchof:
Great finding! But how did you even discover this… ?

marchof (approved):

Godin:
> But how did you even discover this… ?

@marchof I mention JSR and RET instruction in one of my talks, and wanted to refresh my knowledge/memory about them before presentation by checking specification and JaCoCo code 😉

PR #1532, opened 2023-11-01
Fix misleading outdated javadoc

Godin:
Outdated since created in commit 88802a1de70be3fbf7675316f0f66b10f5ff34b4 in JaCoCo version 0.3.3


marchof (approved):

marchof:
Sorry, thanks for fixing! I can actually not remember why I overlooked thin back in 2010... 😅

Godin:
> I can actually not remember why I overlooked thin back in 2010... 😅

@marchof 🤣 I'd be very surprised if you can 🤣

PR #1556, opened 2023-12-13
Fix KotlinDefaultArgumentsFilter for the case of more than 32 parameters

### LINKED ISSUE #1546: Data class with >32 properties shows missing branch coverage
### Steps to reproduce

* JaCoCo version: 0.8.11
* Create a Kotlin data class with >32 properties

### Expected behavior

Tests invoking the constructor should show full coverage.

### Observed behavior

Jacoco reports a branch miss on the constructor line.

This may (?) be due to [KotlinDefaultArgumentsFilter](https://github.com/jacoco/jacoco/blob/master/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java) presuming that there will be 2 extra arguments at the end of the synthetic constructor used by Kotlin. But when there are >32 properties there will be additional bitmask integers, leading `penultimateArgument = argumentTypes.length - 2` to yield an incorrect value.

Godin:
Fixes #1546

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilterTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilterTest.java
@@ -12,6 +12,8 @@
  *******************************************************************************/
 package org.jacoco.core.internal.analysis.filter;
 
+import static org.junit.Assert.assertEquals;
+
 import org.jacoco.core.internal.instr.InstrSupport;
 import org.junit.Test;
 import org.objectweb.asm.Label;
@@ -247,7 +249,7 @@ public class KotlinDefaultArgumentsFilterTest extends FilterTestBase {
 				.add(KotlinGeneratedFilter.KOTLIN_METADATA_DESC);
 
 		m.visitVarInsn(Opcodes.ILOAD, 34);
-		m.visitLdcInsn(1 << 31);
+		m.visitLdcInsn(Integer.valueOf(1 << 31));
 		m.visitInsn(Opcodes.IAND);
 		final Label label = new Label();
 		m.visitJumpInsn(Opcodes.IFEQ, label);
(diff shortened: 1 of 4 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilterTest.java
```diff
@@ -219,4 +219,113 @@ public void should_filter_methods_with_parameters_that_consume_two_slots() {
...
+		final MethodNode m = new MethodNode(InstrSupport.ASM_API_VERSION,
+				Opcodes.ACC_SYNTHETIC, "<init>",
+				"(" + paramTypes
+						+ "IILkotlin/jvm/internal/DefaultConstructorMarker;)V",
+				null, null);
+		context.classAnnotations
+				.add(KotlinGeneratedFilter.KOTLIN_METADATA_DESC);
+
+		m.visitVarInsn(Opcodes.ILOAD, 34);
+		m.visitLdcInsn(1 << 31);
```
marchof, line 250:
```suggestion
		m.visitLdcInsn(Integer.valueOf(1 << 31));
```

Avoid autoboxing warning.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java
```diff
@@ -132,15 +138,21 @@ public void match(final MethodNode methodNode,
...
 		private static int maskVar(final String desc,
 				final boolean constructor) {
+			final Type[] argumentTypes = Type.getMethodType(desc)
+					.getArgumentTypes();
+			int masks;
+			for (masks = 1; masks <= 8; masks++) {
+				if (argumentTypes.length <= masks * 32 + masks + 1) {
+					break;
+				}
+			}
```
marchof, line 148: Mathematically equivalent to:

```suggestion
			int masks = (argumentTypes.length - 2) / 33 + 1;
```

Extract this to its own method with documentation and unit test?

Godin: oh no, increase of method visibility for testing 🙀 😉

marchof (approved):

PR #1581, opened 2024-02-20
Add exclusions support to cli

Stark-X:


Stark-X:
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.3.0:enforce (enforce-distribution-size) on project jacoco: 
[ERROR] Rule 0: org.apache.maven.enforcer.rules.files.RequireFilesSize failed with message:
[ERROR] /home/vsts/work/1/s/jacoco/target/jacoco-0.8.12.202402201145.zip size (4608852) too large. Max. is 4600000/home/vsts/work/1/s/jacoco/target/jacoco-0.8.12.202402201145.zip
[ERROR] -> [Help 1]

Above is the error from JDK9 pipeline, how could I sovle ?

marchof:
@Stark-X We have a sanity check in our build for the size of the zip with all artifacts:

https://github.com/jacoco/jacoco/blob/master/jacoco/pom.xml#L114

You may increase the upper limit.

org.jacoco.core/pom.xml
```diff
@@ -38,6 +38,11 @@
       <groupId>org.ow2.asm</groupId>
       <artifactId>asm-tree</artifactId>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
Stark-X: Could you suggest which package i should place the `FileFilter` which using the `org.codehaus.plexus.util.FileUtils` and `org.codehaus.plexus.util.StringUtils`

org.jacoco.report/pom.xml
```diff
@@ -30,6 +30,11 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>org.jacoco.core</artifactId>
     </dependency>
+    <dependency>
```
marchof, line 33: Really, we don't want to have that dependency outside the Maven plugin. Why should any user of our library get an dependeny on the Maven framework?

marchof: For the exiting filtering in core we use our own class [WildcardMatcher](https://github.com/jacoco/jacoco/blob/master/org.jacoco.core/src/org/jacoco/core/runtime/WildcardMatcher.java), maybe this works here.

Stark-X:
before jdk 7, the `java.nio` does not exists, any suggestion?

barnett-yuxiang (commented):
test code？

Stark-X:
Hi @marchof  ，as the release [0.8.11](https://github.com/jacoco/jacoco/releases/tag/v0.8.11) said, 
> - jacoco-maven-plugin now requires at least Java 8 
> - JaCoCo build now requires JDK 17 

is it still necessary to check JDK 5 compatiblity in the GitHub action?

marchof:
> is it still necessary to check JDK 5 compatiblity in the GitHub action?

It is. The JaCoCo agent runs with in the target application. And here we support Java 5 as a runtime (still many application in production out there using it). As the JaCoCo agent uses large parts of the JaCoCo core we maintain Java 5 compatibility for the overall code base.

PR #1600, opened 2024-03-24
Upgrade ASM to 9.7

### LINKED ISSUE #1599: Update to ASM 9.7
### Scenario

This is currently the most recent version of jacoco available at Maven Central:

https://repo1.maven.org/maven2/org/jacoco/org.jacoco.core/0.8.11

Its OSGi bundle

https://repo1.maven.org/maven2/org/jacoco/org.jacoco.core/0.8.11/org.jacoco.core-0.8.11.jar

has quite restrictive version range dependencies on org.objectweb.asm:

```
org.objectweb.asm.commons;version="[9.6.0,9.7)",org.objectweb.asm.tree;version="[9.6.0,9.7)"
```

There is now a new ASM version 9.7 available:

https://repo1.maven.org/maven2/org/ow2/asm/asm/9.7

It would be nice to have  org.jacoco artifacts available at Maven central compatible with the latest ASM version.

From https://github.com/jacoco/jacoco/issues/1517 I understand that the restrictive range is intentional to avoid breakage that may happen even with new minor versions.

### Current Behaviour

Can't work with latest version of ASM.

### Wanted Behaviour

Would like to work in combination with latest version of ASM.

### Possible Workarounds

Use older version of ASM.

merks:
Fixes #1599


merks:
@marchof @Godin 

I tried to reproduce the corresponding changes for this PR that I saw for the following PR's changes for upgrading from 9.5 to 9.6:

https://github.com/jacoco/jacoco/pull/1518

org.jacoco.doc/docroot/doc/changes.html
```diff
@@ -76,8 +76,8 @@ <h3>Non-functional Changes</h3>
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/1468">#1468</a>).</li>
   <li>JaCoCo build now requires JDK 17
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/1482">#1482</a>).</li>
-  <li>JaCoCo now depends on ASM 9.6
-      (GitHub <a href="https://github.com/jacoco/jacoco/issues/1518">#1518</a>).</li>
+  <li>JaCoCo now depends on ASM 9.7
```
marchof, line 79: We can't modify the change log of released JaCoCo version 0.8.11. Please add on top under "Snapshot...". Thanks!

marchof (changes requested):
Welcome @merks here at the JaCoCo project! Looks good to me except the change log entry.

org.jacoco.doc/docroot/doc/changes.html
merks: I should have looked at the context to see that one needs to add new entries, not modify historical entries.  😱

marchof (approved):

merks:
Thanks guys! I'll try to get the PR fully correct the next time...

merks:
If the link is to the PR rather than the issue, it's not actually possible to get it right until one first creates the PR.  Of course typically it's the issue +1, but I'm not sure that this is always the case...

org.jacoco.build/licenses/asm-9.7.html
```diff
@@ -1,7 +1,7 @@
 <h4>ASM</h4>
 
 <p>
-  <a href="https://asm.ow2.io/">ASM 9.6</a> is subject to the terms and
+  <a href="https://asm.ow2.io/">ASM 9.7</a> is subject to the terms and
```
Godin, line 4: I did check that their license stayed unchanged.

Godin:
> I'll try to get the PR fully correct the next time

Thanks for this PR ❤️ And no worries - slight polishing by us is not a big deal 👐 

Previously we've been quite proactive with ASM updates thanks to a subscription to their mailing list, but seems that they haven't done announcements for the past two versions 😞 

> it's not actually possible to get it right until one first creates the PR

Usually changelog modification is the very last commit and PR is first opened without this modification 😉 

> but I'm not sure that this is always the case

You have to believe me that this is true for the existing entries or at least for the majority.

Godin (approved):
For the record
here is a link to their changelog - https://asm.ow2.io/versions.html
and actual changes between versions - https://gitlab.ow2.org/asm/asm/-/compare/ASM_9_6...ASM_9_7

From this to me seems that this time there were no changes that could negatively impact us.

PR #1625, opened 2024-05-25
Add filter for Kotlin enum classes

### LINKED ISSUE #1557: Investigate why Kotlin K2 compiler produces line numbers in the enums `getEntries` method
Currently execution of

```
mvn \
  -Dkotlin.version=2.0.0-Beta2 \
  -Dbytecode.version=8 \
  clean package
```

leads to

```
Failed tests:
  KotlinWhenExpressionTest>ValidationTestBase.all_missed_instructions_should_have_line_number:142 sum of missed instructions of all lines should be equal to missed instructions of file expected:<4> but was:<3>
```

which is because for

```
enum class Example
```

Kotlin `2.0.0-Beta2` compiler produces

```
  public static kotlin.enums.EnumEntries<Example> getEntries();
    descriptor: ()Lkotlin/enums/EnumEntries;
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
         0: getstatic     #43                 // Field $ENTRIES:Lkotlin/enums/EnumEntries;
         3: areturn
      LineNumberTable:
        line 1: 3
    Signature: #38                          // ()Lkotlin/enums/EnumEntries<LExample;>;
    RuntimeInvisibleAnnotations:
      0: #39()
        org.jetbrains.annotations.NotNull
```

whereas Kotlin `1.9.22` compiler produces

```
  public static kotlin.enums.EnumEntries<Example> getEntries();
    descriptor: ()Lkotlin/enums/EnumEntries;
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
         0: getstatic     #43                 // Field $ENTRIES:Lkotlin/enums/EnumEntries;
         3: areturn
    Signature: #38                          // ()Lkotlin/enums/EnumEntries<LExample;>;
    RuntimeInvisibleAnnotations:
      0: #39()
        org.jetbrains.annotations.NotNull
```

Bisection of changes in Kotlin compiler pointed me to https://github.com/jetBrains/kotlin/commit/5fb38008b7a4548eda14d031abe37ee9fd45c086 which merely changes the language version, so either I made a mistake during bisection, or bytecode generation depends on the language version. I suspect the second because execution of Kotlin `1.9.22` compiler with option `-language-version 2.0` produces the same result as the execution of Kotlin `2.0.0-Beta2` compiler.

@qwwdfsad could you please have a look and advise whether this change is intentional and we need to filter out `getEntries` methods in enums similarly to filtering `valueOf` and `values` methods, or this is an unintentionally overlooked change in the compiler?

Godin:
Fixes #1557


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinEnumFilter.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Filters method <code>getEntries</code> that Kotlin compiler creates for
+ * enums.
+ */
+final class KotlinEnumFilter implements IFilter {
+
+	public void filter(final MethodNode methodNode,
+			final IFilterContext context, final IFilterOutput output) {
+		if (!"java/lang/Enum".equals(context.getSuperClassName())) {
```
leveretka, line 25: I don't know if this is the convention here, but as a code reader, I'd prefer to see a single `if` without negations in the conditions to reduce a cognitive load. Since the body of the if is just one line, this is an acceptable level of nesting.

org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/KotlinWhenExpressionTest.java
```diff
@@ -24,9 +24,4 @@ public KotlinWhenExpressionTest() {
 		super(KotlinWhenExpressionTarget.class);
 	}
 
-	@Override
-	public void all_missed_instructions_should_have_line_number() {
-		// https://github.com/jacoco/jacoco/issues/1557
```
leveretka, line 29: Just to clarify, will the current PR also close this ticket?

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
```diff
@@ -43,6 +43,7 @@ public static IFilter all() {
 				new ExhaustiveSwitchFilter(), //
 				new RecordPatternFilter(), //
 				new AnnotationGeneratedFilter(), new KotlinGeneratedFilter(),
+				new KotlinEnumFilter(), //
```
leveretka, line 46: Why do we need a `//` here?

leveretka (commented):
Overall looks good, thanks for fixing it!

I just have a small suggestion to improve readability.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
Godin: Otherwise Eclipse JDT Formatter will join lines and the change will look like

```diff
                                new AnnotationGeneratedFilter(), new KotlinGeneratedFilter(),
-                               new KotlinLateinitFilter(), new KotlinWhenFilter(),
-                               new KotlinWhenStringFilter(),
+                               new KotlinEnumFilter(), new KotlinLateinitFilter(),
+                               new KotlinWhenFilter(), new KotlinWhenStringFilter(),
                                new KotlinUnsafeCastOperatorFilter(),
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinEnumFilter.java
Godin: We don't have a strict convention about this and both styles can be found, but seems that the early-returns style is prevalent:

single `if` in

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/EnumEmptyConstructorFilter.java#L39-L48

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/PrivateEmptyNoArgConstructorFilter.java#L26-L35

early returns in

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/BridgeFilter.java#L23-L30

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinInlineClassFilter.java#L74-L87

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinGeneratedFilter.java#L31-L50

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultMethodsFilter.java#L25-L31

https://github.com/jacoco/jacoco/blob/5aabb2eb60bbcd05df968005f1746ba19dcd5361/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java#L72-L86

leveretka (approved):

PR #1633, opened 2024-06-08
Specify the version and location of sonar-maven-plugin

leveretka:
Currently, we're executing Sonar analysis by using the command `mvn sonar:sonar`. This approach has 2 problems:

1. No version is defined, so the latest will always be picked up, and this might unexpectedly change the behavior of the analysis. The better approach here is to specify the version and update it explicitly once the new version is out.
2. The second problem is more important. Since the plugin is not mentioned by its groupId and artifactId, it defaults to `org.codehaus.mojo:sonar-maven-plugin`. This is legacy and, eventually Sonar will stop publishing there. Currently, we can also see the warning in the logs:

> [WARNING] The artifact org.codehaus.mojo:sonar-maven-plugin:jar:4.0.0.4121 has been relocated to org.sonarsource.scanner.maven:sonar-maven-plugin:jar:4.0.0.4121: SonarQube plugin was moved to SonarSource organisation

So I suggest this small change to use the right artifact and get rid of the warning.


pom.xml
```diff
@@ -10,4 +10,16 @@
...
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.sonarsource.scanner.maven</groupId>
+          <artifactId>sonar-maven-plugin</artifactId>
+          <version>4.0.0.4121</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
```
Godin, line 23: Should be in `org.jacoco.build/pom.xml` because CI executes `mvn -f org.jacoco.build` which doesn't inherit from this `pom.xml`.

.azure-pipelines/azure-pipelines.yml
```diff
@@ -95,7 +95,7 @@ jobs:
       elif [[ "$BUILD_SOURCEBRANCH" == "refs/heads/master" && "$JDK_VERSION" == "11" ]]; then
         .maven/bin/mvn -V -B -e --no-transfer-progress -f org.jacoco.build \
           verify -Djdk.version=$JDK_VERSION -Dbytecode.version=$JDK_VERSION \
-          sonar:sonar \
+          org.sonarsource.scanner.maven:sonar-maven-plugin:sonar \
```
Godin, line 98: AFAICS this is not needed

Godin (changes requested):

Godin (approved):

Issue #1639, opened 2024-06-13
Support excluding classes from reporting on CLI


ttho:
### Scenario

* JaCoCo version: 0.8.12
* Operating system: all
* Tool integration: CLI
* Description of your use case: Support excluding some classes from report generation on CLI as it is possible for ant (maybe also maven - did not check). 

Implementation is ready. See PR.

### Current Behaviour
With CLI it is not possible to exclude some classes from the report.

### Wanted Behaviour
CLI should allow pattern-based black-listing classes.

### Possible Workarounds
List classes to include one-by-one instead of providing directories.

marchof:
See discussion at #1640

PR #1640, opened 2024-06-13
support for --excludeclassfiles on CLI

ttho:
Support for --excludeclassfiles on CLI. See issue #1639


marchof:
Hi @ttho thanks for the contribution! The tool does not just work for class files, it also accepts ZIP/JAR files where this filtering will not work. 

Also up to now we declined this feature we want to keep the CLI as minimal as possible and don't replicate features which are  provided by the shell.

ttho:
Hi @marchof, feature creep, I see. Should have asked earlier. Implemented my use case by giving jacoco a filtered list which is also fine.

marchof:
@ttho Thanks for your understanding!

PR #1655, opened 2024-07-26
Fix KotlinDefaultArgumentsFilter for case when argument number 33 or higher has default value

### LINKED ISSUE #1750: Function with params count > 32 is covered partially
For the following `src/Example.kt`
```
fun functionWithManyParams(
    x00: Int, x01: Int, x02: Int, x03: Int, x04: Int, x05: Int, x06: Int, x07: Int, x08: Int, x09: Int,
    x10: Int, x11: Int, x12: Int, x13: Int, x14: Int, x15: Int, x16: Int, x17: Int, x18: Int, x19: Int,
    x20: Int, x21: Int, x22: Int, x23: Int, x24: Int, x25: Int, x26: Int, x27: Int, x28: Int, x29: Int,
    x30: Int, x31: Int,
    a: Long = 1
) {
}

fun functionWithTwoParams(
    x00: Int,
    a: Long = 1
) {

}

fun main() {
    functionWithManyParams(
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        0, 0
    )
    functionWithTwoParams(
        0
    )
}


```

execution of

```
kotlin-2.0.21/bin/kotlinc src -d classes
java -javaagent:jacoco-0.8.13-SNAPSHOT/lib/jacocoagent.jar -cp classes:kotlin-2.0.21/lib/kotlin-stdlib.jar ExampleKt
java -jar jacoco-0.8.13-SNAPSHOT/lib/jacococli.jar report jacoco.exec --classfiles classes --sourcefiles src --html report
```
produces report
Screenshot 2024-10-16 at 14 19 48
Screenshot 2024-10-16 at 14 20 00

### Expected behaviour
Function `functionWithManyParams` should be covered same as `functionWithTwoParams`

### Environment
* JaCoCo version:  actual `master` branch built locally
* Kotlin 2.0.21
* Operating system: MacOS 15
* Tool integration: CLI

Godin:
Unfortunately modification in https://github.com/jacoco/jacoco/pull/1556 of KotlinDefaultArgumentsFilter for the case of more than 32 parameters was incomplete:

Screenshot 2024-07-27 at 00 37 54

branches are shown for class "B" because condition

https://github.com/jacoco/jacoco/blob/4f851cae6822289c55aff539786405d2bf473584/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinDefaultArgumentsFilter.java#L120

takes into account only the first mask argument.

Fixes #1750


leveretka (approved):
Good catch! Looks good!

Let's maybe update the title mentioning the fact that it's when the default argument is at the 33+ position. Took me some time to understand what changed and I still had to ask.

PR #1660, opened 2024-07-30
NormalizedFileNames should add prefix instead of suffix

### LINKED ISSUE #1656: Unabled to read report because tilde is attached to the extension
* JaCoCo version:  0.8.12
* Operating system: MacOS
* Tool integration: Maven

I generated HTML reports using the jacoco plugin. However, since the strings that make up the java file name are not included in the [`legal`](https://github.com/jacoco/jacoco/blob/master/org.jacoco.report/src/org/jacoco/report/internal/NormalizedFileNames.java#L42), it is generated with the name "_________Service.html". To make matters worse, even though it's actually a different java file name, since it gets replaced to `_`, so the HTML file name is duplicated and  and generated as "_________Service.html~1". Eventually, it will be downloaded by clicking the link connected to that file.

### Expected behaviour

- expect a unique string(`~idx`) to be present before the extension to ensure uniqueness

Godin:
Otherwise it changes file extensions.

Fixes #1656

Alternatives are
* change it to handle extensions separately/explicitly
* or change its consumers to append extension after it

### DIFF
--- a/org.jacoco.doc/docroot/doc/changes.html
+++ b/org.jacoco.doc/docroot/doc/changes.html
@@ -42,6 +42,9 @@
 <ul>
   <li>Fixed interpretation of Kotlin SMAP
       (GitHub <a href="https://github.com/jacoco/jacoco/issues/1525">#1525</a>).</li>
+  <li>File extensions are preserved in HTML report in case of clashes of normalized
+      file names
+      (GitHub <a href="https://github.com/jacoco/jacoco/issues/1660">#1660</a>).</li>
 </ul>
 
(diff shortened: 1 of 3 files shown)


Godin:
@marchof could you please review? imo the only missing thing here is an update of changelog, so maybe you have an idea of how this fix can be concisely described in it?

marchof:
@Godin What about this:

> Preserve file extensions in HTML report in case of file name clashes.

Godin:
@marchof Thank you for the idea! ❤️ I added slightly modified version of it into changelog.

marchof (approved):

PR #1680, opened 2024-08-09
Preserve information about covered branches for lines of methods

Godin:
This is needed for at least the following two use cases.

----

First - in combination with #1670 this allows to calculate branch coverage for Kotlin inline functions.

----

Second - this information can be shown to users.
This feature exists in IntelliJ IDEA when using https://github.com/JetBrains/intellij-coverage
Based on these changes @zuevmaxim from JetBrains kindly prepared PoC that enables this feature when using JaCoCo in IntelliJ IDEA - https://github.com/Godin/intellij-community/tree/jacoco_branches
Here is how it looks:

Screenshot 2024-08-10 at 00 37 38

And maybe in the future somebody will use this to implement similar feature in https://github.com/eclipse-eclemma/eclemma for Eclipse IDE.


marchof:
As discussed at BaselOne:

* We should add the new capabilities to the API
* This only makes sense for methods. Therefore a new `IMethodLine extends ILine` type should be designed
* To make memory optimization transparent to the callers the `append()` methods should return the new instance to work with.
* There should be only one `append()` method for `LineImpl` and `MethodLineImpl` which needs to be called on the respective use case.

PR #1689, opened 2024-08-19
jacoco-maven-plugin should analyze test classes

Godin:
In combination with #1670 this enables analysis of Kotlin inline functions call sites in tests, i.e. calculation of line coverage for Kotlin inline functions that are invoked only from tests.

For the same example from https://github.com/jacoco/jacoco/pull/1670#issue-2448677183 execution of `mvn verify` after this change produces

Screenshot 2024-08-19 at 13 39 33

Screenshot 2024-08-19 at 13 39 49


ddello32:
@Godin are there plans of advancing with this MR? I have a project that would benefit from this changes.

PR #1675, opened 2024-09-05
Add validation test for synthetic accessor methods in Kotlin

Godin:
We have such [test for Java](https://github.com/jacoco/jacoco/blob/v0.8.12/org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/targets/SyntheticTarget.java), so I think we should add similar for Kotlin too.

### DIFF
--- a/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinSyntheticAccessorsTarget.kt
+++ b/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinSyntheticAccessorsTarget.kt
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2024 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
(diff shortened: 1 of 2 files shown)


org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinSyntheticAccessorsTarget.kt
```diff
@@ -0,0 +1,45 @@
...
+                x += 1 // assertNotCovered()
+                getX()
+            }
+        }
+    }
+
+    @JvmStatic
+    fun main(args: Array<String>) {
+        Outer().Inner()
+    }
```
leveretka, line 43: Not related to this change, just thoughts: It seems to me you need this method with argument and can't make it top-level because of `ValidationTestBase .run` that looks for such signature `targetClass.getMethod("main", String[].class) `.

Since you're adding more Kotlin tests, do you think it's worth creating a `KotlinValidationTestBase` and allowing top-level `main` with no args? Not that it's very important, but it will look more like idiomatic Kotlin.

leveretka (approved):
Looks good!
Thanks for adding it.

PR #1701, opened 2024-09-09
Add configuration for Dependabot to simplify updates of Maven plugins

Godin:


leveretka (approved):
A very nice idea!

.github/dependabot.yml
```diff
@@ -8,7 +8,26 @@ updates:
       - "dependencies"
       - "component: build"
   - package-ecosystem: "maven"
-    directory: "/"
+    directory: "/org.jacoco.build"
```
marchof, line 11: I wonder whether we want to specify rules for every POM separately. There is also a [`directories`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories) option to specify multiple locations. Then we could merge the rules into one block.

Godin: @marchof this separation allows to have different settings (labels and schedule) for different dependencies

marchof (approved):

Issue #1239, opened 2021-11-01
.exec file is created in wrong directory on Windows (wrong encoding)


SimSonic:
### Steps to reproduce

JaCoCo 0.8.7 (maven plugin).
Windows 10, JDK 17 (Liberica), Maven 3.8.3.
Project is located in `D:\Воспроизведение-issue\`
Globally set environment variable JAVA_TOOL_OPTIONS: `-Dfile.encoding=UTF8 -Djansi.passthrough=true -XX:+UnlockExperimentalVMOptions`.
Output of `chcp` in cmd is 866.

Attaching file with reproduce steps: [Воспроизведение-issue.zip](https://github.com/jacoco/jacoco/files/7450750/-issue.zip)

### Expected behaviour
jacoco-unit.exec is created in project's dir.
mvn verify shows in command line current code coverage stats:
```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.683 s - in example.JaCoCoIssueTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.7:report (jacoco-report) @ jacoco-issue ---
[INFO] Loading execution data file D:\Воспроизведение-issue\jacoco-unit.exec
[INFO] Analyzed bundle 'jacoco-issue' with 1 classes
[INFO] 
[INFO] --- maven-jar-plugin:2.4:jar (default-jar) @ jacoco-issue ---
[INFO] Building jar: D:\Воспроизведение-issue\target\jacoco-issue-0.1-SNAPSHOT.jar
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.7:check (jacoco-check) @ jacoco-issue ---
[INFO] Loading execution data file D:\Воспроизведение-issue\jacoco-unit.exec
[INFO] Analyzed bundle 'jacoco-issue' with 1 classes
[INFO] All coverage checks have been met.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  18.800 s
[INFO] Finished at: 2021-11-01T11:24:15+07:00
[INFO] ------------------------------------------------------------------------
```
(this output was manually edited to match exactly expected).

### Actual behaviour
[...]

PR #1737, opened 2024-10-05
#1239 Fix Agent properties passing at Windows filesystem directory with diacritics characters

bedla:
Hi,
for issue https://github.com/jacoco/jacoco/issues/1239 I have created this pull request.
When parameters are encoded (here I have chosen url-encoding) I am able to use Jacoco with diacritics in path.
What do you think?
Thx
Ivos


bedla:
It is failing with JDK5, I assume that there has to be JDK 5 used, right?

marchof:
@bedla Thanks for this contribution! Yes, JaCoCo minimum runtime requirement is Java 5.

For me it looks like breaking change. I don't think we can afford to change the JaCoCo agent API in an incompatible way.

bedla:
I see.
What about adding new parameters for files (mind `destfileEnc` parameter) to make it look like this, and in case that they do not match use encoded ones?

`-javaagent:C:\\Users\\bedla.czech\\IdeaProjects\\jacoco\\jacoco-maven-plugin.test\\target\\local-repo\\org\\jacoco\\org.jacoco.agent\\0.8.13-SNAPSHOT\\org.jacoco.agent-0.8.13-SNAPSHOT-runtime.jar=destfile=C%3A%5CUsers%5Cbedla.czech%5CIdeaProjects%5Cjacoco%5Cjacoco-maven-plugin.test%5Ctarget%5Cit%5Cit-dump%5Ctarget%5Cjacoco.exec,output=tcpserver`
vs
`-javaagent:C:\\Users\\bedla.czech\\IdeaProjects\\jacoco\\jacoco-maven-plugin.test\\target\\local-repo\\org\\jacoco\\org.jacoco.agent\\0.8.13-SNAPSHOT\\org.jacoco.agent-0.8.13-SNAPSHOT-runtime.jar=destfile=C:\Users\bedla.czech\IdeaProjects\jacoco\jacoco-maven-plugin.test\target\it\it-dump\target\jacoco.exec,destfileEnc=C%3A%5CUsers%5Cbedla.czech%5CIdeaProjects%5Cjacoco%5Cjacoco-maven-plugin.test%5Ctarget%5Cit%5Cit-dump%5Ctarget%5Cjacoco.exec,output=tcpserver`

I mean, without having that bug fixed in the JDK, Jacoco (and other Agents) is not usable on Windows with diacritics in filepaths (eg. in corporate laptops where you do not have power to change some parent directories :( ).

marchof:
@bedla I'm a bit reluctant to add extra code for this situation, for the following reasons:

- JaCoCo is widely used and we only got two reports over all the years
- I've never seen an organization that uses written user names as Windows profile names (especially with non-ASCII characters in the profile name)

I would rather prefer to:

- find a way to set environment variables/system properties in a way that the current JVM behaves correctly
- fix the root cause upstream in the JDK

bedla:
I understand and agree in general.
BUT
- I saw it at almost every corporate, I worked for, to put those characters into names :( and because corporate organization you have zero power to change it (I agree that putting strange characters into directories is total non-sense, I also found in the wild that some Mac guys putting colon `:` into file name because they can and making git clones broken on Windows fs :D). And in my country it is very common to have diacritics in surnames.
- I understand this as workaround until JDK fix found.
- Find environment variable) from discussion above I thought that there is no other way except to fix it in JDK - I can give a try to it to find needle in haystack :) 
- Could you share link to JDK reported issue? I want to follow and see progress.

marchof:
@bedla 

Maybe you misunderstood me: Of course the organizations put your real name into the directory entry, but the Windows profile id typically is a handle like `u123456`. This is also done as names can change over time.

bedla:
I am talking about directory name like `c:\Users\IvoŠmíd\` I currently have. At least operations are able not to put spaces into directory names :) - like `c:\Users\Ivo Šmíd\` what I saw in the past.

marchof:
Exactly. In this case "IvoŠmíd" is the profile name. Which is a interesting choice to actually use names for this in an organization. I would guess JaCoCo is their least problem...

PR #1765, opened 2024-11-04
add gradle build

### LINKED ISSUE #1764: jacoco should itself build with gradle so new java versions are immediate supported
java-23 oes not work with jacoco. the gradle plugin is outdated, setting the tool.version to 0.8.13.202410300952 fails as well. to avoid this brawl with every new java version, jacoco should build itself with gradle.

soloturn:
fix #1764.


marchof:
As explained in #1764: We cannot maintain the Gradle plugin here and will also not maintain another parallel build. Please contribute to the Gradle project.

PR #1766, opened 2024-11-04
no-fonts not exists in jdk23, bug is fixed

soloturn:


Godin:
https://bugs.openjdk.org/browse/JDK-8324774 is about addition of `--no-fonts` option into JDK 23 - see corresponding change https://github.com/openjdk/jdk/commit/d0a265039a36292d87b249af0e8977982e5acc7b

Also as you can see our builds with JDK 23 pass with this option, whereas fail fail on JDK 24 when you remove it.

PR #1792, opened 2024-11-24
Add unit test for LocalProbeArrayStrategy

Godin:
We already have dedicated unit tests for other `IProbeArrayStrategy` implementations - see [`NoneProbeArrayStrategyTest`](https://github.com/jacoco/jacoco/blob/v0.8.12/org.jacoco.core.test/src/org/jacoco/core/internal/instr/NoneProbeArrayStrategyTest.java) and [`CondyProbeArrayStrategyTest`](https://github.com/jacoco/jacoco/blob/v0.8.12/org.jacoco.core.test/src/org/jacoco/core/internal/instr/CondyProbeArrayStrategyTest.java).

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/instr/LocalProbeArrayStrategyTest.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/instr/LocalProbeArrayStrategyTest.java
@@ -1,0 +1,73 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2024 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.internal.instr;
+
+import static org.junit.Assert.assertEquals;
+
+import org.jacoco.core.instr.MethodRecorder;
+import org.jacoco.core.runtime.IExecutionDataAccessorGenerator;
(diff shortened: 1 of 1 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/instr/LocalProbeArrayStrategyTest.java
```diff
@@ -0,0 +1,75 @@
...
+		assertEquals(42, maxStack);
+
+		final MethodRecorder expectedMethod = new MethodRecorder();
+		final MethodVisitor expected = expectedMethod.getVisitor();
+		expected.visitMethodInsn(Opcodes.INVOKESTATIC, "Runtime", "getProbes",
+				"()[Z", false);
+		expected.visitVarInsn(Opcodes.ASTORE, 13);
+		final MethodRecorder actualMethod = new MethodRecorder();
+		m.instructions.accept(actualMethod.getVisitor());
+		assertEquals(expectedMethod, actualMethod);
```
marchof, line 63: Tmp copy in MethodNode not required, MethodRecorder can be used directly:

```suggestion
		final MethodRecorder actualMethod = new MethodRecorder();
		final int maxStack = strategy.storeInstance(actualMethod.getVisitor(),
				false, 13);
		assertEquals(42, maxStack);

		final MethodRecorder expectedMethod = new MethodRecorder();
		final MethodVisitor expected = expectedMethod.getVisitor();
		expected.visitMethodInsn(Opcodes.INVOKESTATIC, "Runtime", "getProbes",
				"()[Z", false);
		expected.visitVarInsn(Opcodes.ASTORE, 13);
		assertEquals(expectedMethod, actualMethod);
```

Godin: @marchof nice catch! 👍 fixed

marchof (approved):

PR #1797, opened 2024-11-25
Add convenience method for asserting MethodNodes

marchof:
As a follow-up from the previously added new tests I propose to add a new convenience method to the test utils to directly obtain a `MethodRecoder` instance from an existing `MethodNode`.


org.jacoco.core.test/src/org/jacoco/core/instr/MethodRecorder.java
```diff
@@ -33,6 +34,12 @@ public MethodRecorder() {
 		visitor = new TraceMethodVisitor(printer);
 	}
 
+	public static MethodRecorder from(MethodNode method) {
```
Godin, line 37:
```suggestion
	public static MethodRecorder from(final MethodNode method) {
```

Godin (approved):

PR #1799, opened 2024-11-26
Fix typos

Godin:
Replace
* "a enum" by "an enum"
* "a optional" by "an optional"
* "a input file" by "an input file"
* "a explicit" by "an explicit"
* "a implicit" by "an implicit"
* "a extra" by "an extra"
* "a event" by "an event"
* "a execution" by "an execution"
* "a immutable" by "an immutable"
* "a instance" by "an instance"
* "a unconditional" by "an unconditional"
* "a internal" by "an internal"
* "a entry" by "an entry"
* "a identifier" by "an identifier"
* "a item" by "an item"
* "a index" by "an index"
* "a in-memory" by "an in-memory"
* "a element" by "an element"

"an" should be used instead of "a" when the
following word starts with a vowel sound.


Godin:
@marchof To be honest, IMO better to fix such little things as they are found, rather than remembering that there is a branch with similar unfinished work. So to finish this PR I went through all code comments and hopefully fixed all of this kind, but still can't guarantee and still doubt that fixed all the typos.

marchof (approved):

marchof (approved):

Issue #1684, opened 2024-08-13
Add validation test for finally in Kotlin


Godin:
We have such [test for Java](https://github.com/jacoco/jacoco/blob/v0.8.12/org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/targets/FinallyTarget.java), so I think we should add similar for Kotlin too.

PR #1806, opened 2024-12-06
IFilter implementations should be package-private and final

Godin:
Currently visibility of IFilter implementations is inconsistent:

package-private `final`
* `AssertFilter`
* `BridgeFilter`
* `ExhaustiveSwitchFilter`
* `KotlinComposeFilter`
* `KotlinDefaultMethodsFilter`
* `KotlinEnumFilter`
* `KotlinInlineClassFilter`
* `RecordPatternFilter`

`public` non-`final`
* `KotlinGeneratedFilter`
* `KotlinLateinitFilter`

all others are `public final`.

I propose to make them all, except `Filters`, consistently package-private and `final`.
IMO design-wise they should have been such since the beginning.

This will also fix a good amount of javadoc doclint warnings 
```
use of default constructor, which does not provide a comment
```
which appear when building with JDK versions starting from 18.


marchof:
👍 For cleaning-up package visibility!

A remark regarding method `isKotlinClass()`:

* The new location introduces a cyclic dependency between `Filters` and the Kotlin filter implementations (which we also have at other places)
* The following filters miss this check: `KotlinLateinitFilter`, `KotlinWhenFilter`, `KotlinWhenStringFilter`, `KotlinNotNullOperatorFilter`

Both could be addressed by creating a separate Group for the Kotlin filters. It also eliminates the repeated `isKotlinClass()` checks in every Filter:

	private static IFilter allKotlinFilters() {
		return new Filters( //
				new KotlinGeneratedFilter(), //
				new KotlinEnumFilter(), //
				new KotlinLateinitFilter(), //
				new KotlinWhenFilter(), //
				new KotlinWhenStringFilter(), //
				new KotlinUnsafeCastOperatorFilter(), //
				new KotlinNotNullOperatorFilter(), //
				new KotlinInlineClassFilter(), //
				new KotlinDefaultArgumentsFilter(), //
				new KotlinInlineFilter(), new KotlinCoroutineFilter(), //
				new KotlinDefaultMethodsFilter(), //
				new KotlinComposeFilter()) {
			@Override
			public void filter(final MethodNode methodNode,
					final IFilterContext context, final IFilterOutput output) {
				if (isKotlinClass(context)) {
					super.filter(methodNode, context, output);
				}
			}
		};
	}

Godin:
> A remark regarding method `isKotlinClass()`:
> The following filters miss this check: `KotlinLateinitFilter`, `KotlinWhenFilter`, `KotlinWhenStringFilter`, `KotlinNotNullOperatorFilter`

* `KotlinWhenStringFilter` - recognizes pattern where switch by `hashCode` of `String` is followed by `String` comparisons, which is IMO unlikely can be written by a human, and even if it will filter non-Kotlin classes, then this most likely will mean that non-Kotlin compiler implemented the same strategy to implement switch by `String` and it should be filtered anyway - we already have filters that are applicable for different languages and compilers - just a few examples among others are [`StringSwitchFilter`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/StringSwitchFilter.java#L27-L32) and [`FinallyFilter`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/FinallyFilter.java)

* `KotlinLateinitFilter` - unlikely that [`INVOKESTATIC kotlin/jvm/internal/Intrinsics.throwUninitializedPropertyAccessException`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinLateinitFilter.java#L81-L83) appears outside Kotlin classes

* `KotlinNotNullOperatorFilter` unlikely that [`INVOKESTATIC kotlin/jvm/internal/Intrinsics.throwNpe`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinNotNullOperatorFilter.java#L40-L41) appears outside Kotlin classes

* `KotlinWhenFilter` - unlikely that [`INVOKESPECIAL kotlin/NoWhenBranchMatchedException.<init>`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinWhenFilter.java#L55) and [`GETSTATIC` for a field with a name starting with `$EnumSwitchMapping$` from class with a name ending `$WhenMappings`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinWhenFilter.java#L87-L94) appear outside Kotlin classes

So IMO the absence of this check in them is not a problem. We can add it, but IMO such a change should not be made within this PR.

> The new location introduces a cyclic dependency between `Filters` and the Kotlin filter implementations

I agree that factoring out repeated `isKotlinClass` is indeed a good thing to do, even on its own, and actually I have been planning to do so for some time already.

But unfortunately this will not remove all cycles - it is also used in [`SyntheticFilter`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java)
It seems that `SyntheticFilter` should be split into two or more to remove all cycles - for Kotlin and for the rest. IMO such change should not be made within this PR.

There are also filters which are likely not applicable for Kotlin - for example [`TryWithResourcesJavac11Filter`](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/TryWithResourcesJavac11Filter.java). IMO such classification should not be made within this PR and instead made carefully with the addition of more tests to not regress which I already started to add - see https://github.com/jacoco/jacoco/pull/1703 and https://github.com/jacoco/jacoco/issues/1684

@marchof All in all, I think that the scope of this PR should not be extended above its current title - change of visibility of classes (see "Clear scope" in [org.jacoco.doc/docroot/doc/conventions.html](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.doc/docroot/doc/conventions.html#L161-L163)), and I think that it should not be blocked and should not wait for all the above, ie IMO factorization of `isKotlinClass` and removal of cycles can be done later. Also it seems that to avoid cycles `isKotlinClass` can be moved to `KotlinSMAP` - I just updated this PR to do so. WDYT about this?

marchof:
@Godin Ok, let's focus on the class visibility in this PR and discuss `isKotlinClass()` in a separate PR, ok?

Godin:
@marchof ok and so where this PR should place `isKotlinClass`? 😅 😉 
1. in `KotlinSMAP` class as it is currently proposed in this PR - I like this variant because they are related since the use of `KotlinSMAP` must be guarded by `isKotlinClass` as it can't be used for non-Kotlin classes, and seems that this avoids cycles
2. in `Filters` class as was initially proposed in this PR - this variant introduced cycles
4. in `KotlinGeneratedFilter` as was initially - in this case `KotlinGeneratedFilter` will remain `public` and we'll need to add a comment to its constructor to get rid of the warning
5. and one more idea just came to my mind - as non-`static` in interface `IFilterContext` implemented by `ClassAnalyzer`, but IMO this is overkill and again IMO requires separate PR

marchof:
@Godin Ah I see, due to visibility issues we need to move it 😅

So let's go with your initial proposal and keep it in `Filters` for now.

marchof (approved):

marchof:
@Godin Thanks! Let me propose a follow-up for the filter code organization.

PR #1809, opened 2024-12-09
Restructure filter grouping

marchof:
Hi @Godin, this is a proposal how we could group the filters. We may move one step further and put filters in separate sub-packages (Java/Kotlin).

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
@@ -12,8 +12,10 @@
  *******************************************************************************/
 package org.jacoco.core.internal.analysis.filter;
 
+import org.objectweb.asm.tree.MethodNode;
+
 /**
- * Filter that combines other filters.
+ * Factory for all JaCoCo filters.
  */
 public final class Filters {
 
@@ -23,11 +25,17 @@ public final class Filters {
 	public static final IFilter NONE = new FilterSet();
 
 	/**
-	 * Creates filter that combines all other filters.
+	 * Creates a filter that combines all filters.
 	 *
-	 * @return filter that combines all other filters
+	 * @return filter that combines all filters
 	 */
 	public static IFilter all() {
+		return new FilterSet( //
+				allCommonFilters(), //
+				allKotlinFilters());
+	}
+
+	private static IFilter allCommonFilters() {
 		return new FilterSet( //
 				new EnumFilter(), //
 				new SyntheticFilter(), //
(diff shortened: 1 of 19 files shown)


Godin:
@marchof I understood what you meant in https://github.com/jacoco/jacoco/pull/1806#issuecomment-2525536887 since the beginning, and actually I did a very similar change locally 😉 

As I said in https://github.com/jacoco/jacoco/pull/1806#issuecomment-2527732972

> I agree that factoring out repeated isKotlinClass is indeed a good thing to do, even on its own, and actually I have been planning to do so for some time already.

But after doing it, I realized that

> unfortunately this will not remove all cycles - it is also used in [SyntheticFilter](https://github.com/jacoco/jacoco/blob/e571765a0ec4fc6d6ce31b708e13016904bc3fbc/org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/SyntheticFilter.java)
It seems that `SyntheticFilter` should be split into two or more to remove all cycles - for Kotlin and for the rest.

So to break the cycle `Filters` → `SyntheticFilter` → `Filters.isKotlinClass` we need https://github.com/jacoco/jacoco/pull/1811, after which `Filter.isKotlinClass` can be removed from `SyntheticFilter` and placed in

```
	private static IFilter allNonKotlinFilters() {
		return new FilterSet( //
				new SyntheticFilter()) {
			@Override
			public void filter(final MethodNode methodNode,
					final IFilterContext context, final IFilterOutput output) {
				if (!isKotlinClass(context)) {
					super.filter(methodNode, context, output);
				}
			}
		};
	}
```

finally breaking the cycle.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
```diff
@@ -15,54 +15,70 @@
...
-				new KotlinCoroutineFilter(), new KotlinDefaultMethodsFilter(),
-				new KotlinComposeFilter());
+		return new FilterSet( //
+				allJavaFilters(), //
+				allKotlinFilters());
 	}
 
-	private Filters(final IFilter... filters) {
-		this.filters = filters;
+	private static IFilter allJavaFilters() {
```
Godin, line 38: There are applied even if source language is not Java, so I prefer

```suggestion
	private static IFilter allCommonFilters() {
```

Godin (changes requested):

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Filters.java
```diff
@@ -15,56 +15,72 @@
 import org.objectweb.asm.tree.MethodNode;
 
 /**
- * Filter that combines other filters.
+ * Factory for all JaCoCo filters.
  */
-public final class Filters implements IFilter {
+public final class Filters {
```
Godin, line 20:
```suggestion
public final class Filters {

	private Filters() {
		// no instances
	}
```

Godin (approved):
Looks good.

Just can be noted that there are no direct unit tests for the behavior of `FilterSet` and `isKotlinClass` in `Filters.allKotlinFilters` - observed coverage is indirect from `AnalyzerTest` and `ClassAnalyzerTest`, but we have validation tests, so this can be improved later.

PR #1810, opened 2024-12-12
Add filter for Kotlin safe call operators chains

Godin:


leveretka (approved):
LGTM!

Discussed on-site about the case where `c` is actually `Nullable` VS its "nullability" is propagated due to nullable `b`. These cases are indistinguishable.

Agreed that it is better to filter this out since in the described case, when nullability is propagated, it's impossible to test.

This will be a huge improvement for many Kotlin users. Thanks!

Issue #1773, opened 2024-11-12
Fix handling of implicit `else` clause of `when` by String in Kotlin


Godin:
<img width="505" alt="Screenshot 2024-11-12 at 01 24 01" src="https://github.com/user-attachments/assets/13319241-9c27-459f-aa01-d9ce08d3a3fa">

whereas for ordinary `when`:

<img width="505" alt="Screenshot 2024-11-12 at 01 24 45" src="https://github.com/user-attachments/assets/c27445d1-9502-4fdd-a6b4-1a4a1afd477f">

see also https://github.com/jacoco/jacoco/issues/1772

----

For the following `Example.kt`

```
fun example(input: String) {
    when (input) {
         "a" -> println("case a")
         "b" -> println("case b")
         "c" -> println("case c")
    }
}
```

Kotlin compiler generates

```
  public static final void example(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: (0x0019) ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=2, locals=2, args_size=1
         0: aload_0
         1: ldc           #9                  // String input
         3: invokestatic  #15                 // Method kotlin/jvm/internal/Intrinsics.checkNotNullParameter:(Ljava/lang/Object;Ljava/lang/String;)V
         6: aload_0
         7: astore_1
         8: aload_1
         9: invokevirtual #21                 // Method java/lang/String.hashCode:()I
        12: tableswitch   { // 97 to 99
                      97: 40
                      98: 52
                      99: 64
                 default: 109
            }
        40: aload_1
        41: ldc           #23                 // String a
        43: invokevirtual #27                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        46: ifne          76
        49: goto          109
        52: aload_1
        53: ldc           #29                 // String b
        55: invokevirtual #27                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        58: ifne          88
        61: goto          109
        64: aload_1
        65: ldc           #31                 // String c
        67: invokevirtual #27                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        70: ifne          100
        73: goto          109
        76: ldc           #33                 // String case a
        78: getstatic     #39                 // Field java/lang/System.out:Ljava/io/PrintStream;
        81: swap
        82: invokevirtual #45                 // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
        85: goto          109
[...]

Issue #1772, opened 2024-11-12
Fix handling of implicit `default` clause of `switch` by String when compiled by ECJ


Godin:
when compiled by javac:

<img width="505" alt="Screenshot 2024-11-12 at 01 16 24" src="https://github.com/user-attachments/assets/4954c198-784a-42b6-b4eb-42872c4db8a3">

whereas when compiled by ECJ:

<img width="505" alt="Screenshot 2024-11-12 at 01 18 10" src="https://github.com/user-attachments/assets/764c0bcd-d86a-4632-8f9c-4cd9bda4f11c">

in case of javac result is consistent with ordinary switch:

<img width="505" alt="Screenshot 2024-11-12 at 01 19 51" src="https://github.com/user-attachments/assets/6bf8706d-71f0-4721-b7d4-a229ad67875a">

----

For the following `Example.java`

```java
class Example {
    static void example(String input) {
        switch (input) {
        case "a":
            System.out.println("case a");
            break;
        case "b":
            System.out.println("case b");
            break;
        case "c":
            System.out.println("case c");
            break;
        }
    }
}
```

ECJ generates

```
  static void example(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: (0x0008) ACC_STATIC
    Code:
      stack=2, locals=2, args_size=1
         0: aload_0
         1: dup
         2: astore_1
         3: invokevirtual #13                 // Method java/lang/String.hashCode:()I
         6: lookupswitch  { // 3
                      97: 40
                      98: 52
                      99: 64
                 default: 106
            }
        40: aload_1
        41: ldc           #19                 // String a
        43: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        46: ifne          76
        49: goto          106
        52: aload_1
        53: ldc           #25                 // String b
        55: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        58: ifne          87
        61: goto          106
        64: aload_1
        65: ldc           #27                 // String c
        67: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        70: ifne          98
        73: goto          106
        76: getstatic     #29                 // Field java/lang/System.out:Ljava/io/PrintStream;
        79: ldc           #35                 // String case a
        81: invokevirtual #37                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        84: goto          106
[...]

PR #1813, opened 2024-12-12
`IFilterOutput.replaceBranches` should allow the use of individual branches of instructions

Godin:
This is required for the resolution of 
* https://github.com/jacoco/jacoco/issues/1772
* https://github.com/jacoco/jacoco/issues/1773

For example for the following `Example.java`

```java
class Example {
    static void example(String input) {
        switch (input) {
        case "a":
            System.out.println("case a");
            break;
        case "b":
            System.out.println("case b");
            break;
        case "c":
            System.out.println("case c");
            break;
        }
    }
}
```

ECJ generates

```
  static void example(java.lang.String);
    descriptor: (Ljava/lang/String;)V
    flags: (0x0008) ACC_STATIC
    Code:
         0: aload_0
         1: dup
         2: astore_1
         3: invokevirtual #13                 // Method java/lang/String.hashCode:()I
         6: lookupswitch  { // 3
                      97: 40
                      98: 52
                      99: 64
                 default: 106
            }
        40: aload_1
        41: ldc           #19                 // String a
        43: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        46: ifne          76
        49: goto          106
        52: aload_1
        53: ldc           #25                 // String b
        55: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        58: ifne          87
        61: goto          106
        64: aload_1
        65: ldc           #27                 // String c
        67: invokevirtual #21                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        70: ifne          98
        73: goto          106
        76: getstatic     #29                 // Field java/lang/System.out:Ljava/io/PrintStream;
        79: ldc           #35                 // String case a
        81: invokevirtual #37                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        84: goto          106
        87: getstatic     #29                 // Field java/lang/System.out:Ljava/io/PrintStream;
        90: ldc           #42                 // String case b
        92: invokevirtual #37                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        95: goto          106
        98: getstatic     #29                 // Field java/lang/System.out:Ljava/io/PrintStream;
       101: ldc           #44                 // String case c
       103: invokevirtual #37                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
       106: return
```

`default` branch of `switch` should be considered executed only if
* or one of `goto 106` instructions (at offsets `46`, `61`, `73` or `84`) was executed (`input` does not match any `case`, but has the same `hasCode` as one of them)
* `default` (branch `0`) of `lookupswitch` instruction was executed (`hashCode` of input does not match `hashCode`s of `case`s)

and not when instruction at offset `106` is reached as it is now.


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/MethodAnalyzerTest.java
```diff
@@ -514,6 +519,128 @@ public void jump_to_first_instruction_should_show_full_branch_coverage_when_prob
...
+		// probe[1] when jump to l2
+		method.visitJumpInsn(Opcodes.IFNONNULL, l2);
+		// probe[2] is unreachable because `b.c` can not be null
+		method.visitLabel(l1);
+		method.visitInsn(Opcodes.POP);
+		method.visitLdcInsn("");
+		// probe[3]
+		method.visitLabel(l2);
+		// probe[4]
+		method.visitInsn(Opcodes.ARETURN);
```
Godin, line 553: As a side note:

Every time I touch `MethodAnalyzerTest` I struggle to recall/realize how/where probes are inserted. Maybe it becomes easier after every touch, but still. That's why I started to add comments in the form `// probe[id]` - they at least help to recall, but unfortunately not all tests have them, and unfortunately they do not appear magically for new tests 😅 

IMO it would be nice to have a way to automatically show the places of probes or even better - assert them. Moreover `MethodAnalyzerTest` is already indirectly testing `LabelFlowAnalyzer` and `MethodProbesAdapter` and IMO contains more interesting (and far more) scenarios than their tests.

org.jacoco.core/src/org/jacoco/core/internal/analysis/Instruction.java
```diff
@@ -160,22 +159,27 @@ public Instruction merge(final Instruction other) {
...
-		result.branches = newBranches.size();
-		int idx = 0;
-		for (final Instruction b : newBranches) {
-			if (!b.coveredBranches.isEmpty()) {
-				result.coveredBranches.set(idx++);
+		for (int i = 0; i < branches.length; i++) {
+			if (fromInstructions[i].coveredBranches.get(fromBranches[i])) {
+				result.coveredBranches.set(branches[i]);
 			}
+			result.branches = Math.max(result.branches, branches[i] + 1);
```
Godin, line 182: As a side note:

`Math.max` is used here because some unit tests violate contract of `Instruction` that states that after CFG construction [each instruction has at least one branch](https://github.com/jacoco/jacoco/blob/e6f29c902242c24c20f7740e27006a83330b8e88/org.jacoco.core/src/org/jacoco/core/internal/analysis/Instruction.java#L23).

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -105,18 +105,36 @@ private void applyMerges() {
...
+			int b = 0;
+			for (final Collection<InstructionBranch> list : targets.values()) {
+				for (final InstructionBranch ib : list) {
+					branches[i] = b;
+					fromInstructions[i] = instructions.get(ib.instruction);
+					fromBranches[i] = ib.branch;
+					i++;
+				}
+				b++;
+			}
```
Godin, line 131: I already did not like that the old code was doing allocations merely to pass the data between different interfaces, but the code was small enough, and now it become bigger 😞 

IMO allocations can be avoided and code can be simplified by introduction of "mapper" interface similar to `java.util.function.Function<AbstractInstructionNode, Instruction>`

```
mapper = (AbstractInsnNode node) -> instructions.get(node);
mapper.apply(node).replaceBranches(mapper, targets);
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/Instruction.java
```diff
@@ -160,22 +159,27 @@ public Instruction merge(final Instruction other) {
 
 	/**
 	 * Creates a copy of this instruction where all outgoing branches are
-	 * replaced with the given instructions. The coverage status of the new
-	 * instruction is derived from the status of the given instructions.
+	 * replaced. The coverage status of the branches of the new instruction is
+	 * derived from the status of the given branches of the given instructions.
```
Godin, line 163:
```suggestion
	 * replaced. The coverage statuses of the branches of the new instruction are
	 * derived from the statuses of the given branches of the given instructions.
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -53,7 +55,42 @@ public interface IFilterOutput {
...
+	 * @param newTargets
+	 *            new targets of branches
+	 * @deprecated use {@link #replaceBranches(AbstractInsnNode, Map)} instead
+	 */
+	@Deprecated
 	void replaceBranches(AbstractInsnNode source,
 			Set<AbstractInsnNode> newTargets);
 
+	/**
+	 * {@link #instruction} and index of its {@link #branch}.
```
Godin, line 73: It will be nice to add a comment about branch indexes for `JumpInsnNode`, `TableSwitchInsnNode`, `LookupSwitchInsnNode` and others.

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -130,9 +131,10 @@ private void applyReplacements() {
 			}
 
 			final AbstractInsnNode node = entry.getKey();
-			instructions.put(node, instructions.get(node)
+			replacedInstructions.put(node, instructions.get(node)
 					.replaceBranches(branches, fromInstructions, fromBranches));
```
Godin: Execution of new tests of `MethodAnalyzerTest` in CI with different JDK versions caught the need to store replaced instructions separately from originals: in this test computation of new statuses for two different instructions uses their original statuses - without separation result depends on the order of entries in `replacements` `Map`.

Food for thought: is it possible to demonstrate similar problem on old code? ie that order of replacements matters.

Also it would be nice to add a check at least into the validation tests that no two filters replace the same instruction.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -53,7 +55,42 @@ public interface IFilterOutput {
...
+	 */
+	class InstructionBranch {
+		/** Instruction. */
+		public final AbstractInsnNode instruction;
+		/** Branch. */
```
Godin, line 78:
```suggestion
		/** Branch index. */
```

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/InstructionTest.java
```diff
@@ -164,11 +162,29 @@ public void replaceBranches_should_calculate_coverage_on_new_branches() {
 		Instruction i1 = new Instruction(1);
 		Instruction i2 = new Instruction(2);
 		Instruction i3 = new Instruction(3);
```
Godin, line 164: At least this violates `Instruction` contract - see https://github.com/jacoco/jacoco/pull/1813/files#r1883024911

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -170,8 +188,28 @@ public void merge(AbstractInsnNode i1, AbstractInsnNode i2) {
...
+	/**
+	 * @deprecated use {@link #replaceBranches(AbstractInsnNode, Map)} instead
+	 */
+	@Deprecated
 	public void replaceBranches(final AbstractInsnNode source,
 			final Set<AbstractInsnNode> newTargets) {
+		final HashMap<AbstractInsnNode, Collection<InstructionBranch>> replacements = new HashMap<AbstractInsnNode, Collection<InstructionBranch>>();
+		for (final AbstractInsnNode target : newTargets) {
+			final ArrayList<InstructionBranch> list = new ArrayList<InstructionBranch>();
+			final int branches = Math.max(
```
Godin, line 200: `Math.max` is used here for the same reason as in https://github.com/jacoco/jacoco/pull/1813/files#r1883024911

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
```diff
@@ -105,17 +105,33 @@ private void applyMerges() {
...
+				b++;
+			}
+
 			final AbstractInsnNode node = entry.getKey();
-			instructions.put(node,
-					instructions.get(node).replaceBranches(newBranches));
+			instructions.put(node, instructions.get(node)
+					.replaceBranches(branches, fromInstructions, fromBranches));
 		}
 	}
```
Godin, line 136: I already did not like that the old code was doing allocations merely to pass the data between different interfaces, but the code was small enough, and now it become bigger 😞 

IMO allocations can be avoided and code can be simplified by introduction of "mapper" interface similar to `java.util.function.Function<AbstractInstructionNode, Instruction>`

```
mapper = (AbstractInsnNode node) -> instructions.get(node);
mapper.apply(node).replaceBranches(mapper, targets);
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
Godin: Done in https://github.com/jacoco/jacoco/pull/1813/commits/348bfc0e0240fdd02c62497f4d41ad3f26224835

leveretka (commented):
Very impressive work that will unblock many Kotlin features. Thanks for working on that.

So far the core change looks very straightforward (thanks for splitting it into commits was much easier to review).

I will probably need a second look on the updated filters. But so far the change looks nice.

I would probably recommend splitting the PR into several commits. In the first you add a new API `for` the `replaceBranches`, and in the other ones you migrate filters. You already have it in several commits, maybe we can also merge it separately (Just an idea)

I still would like to hear @marchof opinion, as you change some internals.

leveretka (approved):
Since I don't see any concerns mentioned here, I think we can merge it

org.jacoco.core/src/org/jacoco/core/internal/analysis/MethodCoverageCalculator.java
Godin: Done in 12778c3093972a27b5bfd0234197ffd3e91318a2

jacoco/pom.xml
```diff
@@ -111,8 +111,8 @@
             <configuration>
               <rules>
                 <requireFilesSize>
-                  <maxsize>4600000</maxsize>
-                  <minsize>4000000</minsize>
+                  <maxsize>4700000</maxsize>
+                  <minsize>4100000</minsize>
```
Godin, line 115: even without this PR we got close to upper limit - see https://github.com/jacoco/jacoco/commit/5334d9fb6cbcd78e35217556a40703b927c2b80f

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/InstructionTest.java
```diff
@@ -161,14 +168,61 @@ public void merge_should_calculate_superset_of_covered_branches() {
...
+								new IFilterOutput.InstructionBranch(n2, 0)),
+						Collections.singleton(
+								new IFilterOutput.InstructionBranch(n3, 0))),
+				mapper);
+		assertEquals(CounterImpl.getInstance(3, 0),
+				instruction.getBranchCounter());
 
-		instruction = instruction.replaceBranches(Arrays.asList(i1, i2, i3));
+		instruction = instruction.replaceBranches(
+				Arrays.<Collection<IFilterOutput.InstructionBranch>> asList(
```
marchof, line 203: This expression results in the following warning for me:

> Type safety: A generic array of Collection<IFilterOutput.InstructionBranch> is created for a varargs 
parameter

Same problem in MethodAnalyzerTest and MethodCoverageCalculatorTest. Not a blocker for me though.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/IFilterOutput.java
```diff
@@ -50,10 +51,50 @@ public interface IFilterOutput {
...
 	 *            instruction which branches should be replaced
-	 * @param newTargets
-	 *            new targets of branches
+	 * @param newBranches
+	 *            new branches
+	 * @see Replacements
 	 */
 	void replaceBranches(AbstractInsnNode source,
-			Set<AbstractInsnNode> newTargets);
+			Iterable<Collection<InstructionBranch>> newBranches);
```
marchof, line 59: The parameter type `Iterable<Collection<InstructionBranch>>` looks strange without further explanation. A wonder whether you could use `Replacements` as a type here. In nearly all use case the method is called with:

    replaceBranches(..., replacements.values());
    
Then it would be 

    replaceBranches(..., replacements);

marchof (approved):
I have to admit that I wasn't able to wrap my mind around the functionality implemented here. So I left only minor comments regarding the code structure. @Godin please feel free to just ignore them if you don't agree.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/InstructionTest.java
Godin: This looks like a consequence of compilation for Java 5:

given

```java
import java.util.Collections;

public class Example {

    @SafeVarargs
    static <T> void example1(T... a) {
    }

    static <T> void example2(T... a) {
    }

        example1(Collections.singleton(""));
        example2(Collections.singleton(""));
    }

}
```

execution of

```
java -jar ecj-3.37.0.jar -source 5 -target 5 Example.java
```

produces

```
----------
1. WARNING in /private/tmp/e/Example.java (at line 13)
        example1(Collections.singleton(""));
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type safety: A generic array of Set<String> is created for a varargs parameter
----------
2. WARNING in /private/tmp/e/Example.java (at line 14)
        example2(Collections.singleton(""));
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type safety: A generic array of Set<String> is created for a varargs parameter
----------
2 problems (2 warnings)
```

whereas execution of

```
java -jar ecj-3.37.0.jar -source 7 -target 7 Example.java
```

produces

```
----------
1. WARNING in /private/tmp/e/Example.java (at line 9)
        static <T> void example2(T... a) {
                                      ^
Type safety: Potential heap pollution via varargs parameter a
----------
2. WARNING in /private/tmp/e/Example.java (at line 14)
        example2(Collections.singleton(""));
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type safety: A generic array of Set<String> is created for a varargs parameter
----------
2 problems (2 warnings)
```

in Java 7 `Arrays.asList` was marked with annotation `@SafeVarargs` like `example1` above

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Replacements.java
```diff
@@ -0,0 +1,141 @@
...
+	 * {@link AbstractInsnNode#getNext()}</li>
+	 * </ul>
+	 * </li>
+	 *
+	 * <li>for instructions with {@link org.objectweb.asm.Opcodes#RETURN} and
+	 * {@link org.objectweb.asm.Opcodes#ATHROW} there is only branch 0 that
+	 * corresponds to exit from the method</li>
+	 *
+	 * <li>there are no branches for instructions whose
+	 * {@link AbstractInsnNode#opcode} is -1</li>
```
Godin:
```suggestion
	 * {@link AbstractInsnNode#getOpcode()} is -1</li>
```

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterTestBase.java
```diff
@@ -68,17 +69,77 @@ final void assertMethodIgnored(final MethodNode m) {
...
+		assertEquals(expectedStringBuilder.toString(),
+				actualStringBuilder.toString());
+	}
+
+	static class Replacement {
+		final int newBranch;
+		final AbstractInsnNode instruction;
+		final int branch;
+
+		Replacement(int newBranch, AbstractInsnNode instruction, int branch) {
```
Godin:
```suggestion
		Replacement(final int newBranch, final AbstractInsnNode instruction, final int branch) {
```

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/InstructionTest.java
Godin: Anyway got rid of warnings in 02908fa4513f109d347dbad8dba0ba8acad297af

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/Replacements.java
```diff
@@ -0,0 +1,178 @@
...
+		 *            branch index
+		 */
+		public InstructionBranch(final AbstractInsnNode instruction,
+				final int branch) {
+			this.instruction = instruction;
+			this.branch = branch;
+		}
+
+		@Override
+		public boolean equals(Object o) {
```
marchof, line 163:
```suggestion
		public boolean equals(final Object o) {
```

PR #1819, opened 2024-12-16
Improve KotlinSafeCallOperatorFilter for the cases when chain of safe call operators throws exception

Godin:
Prior to #1810

Screenshot 2024-12-16 at 23 15 52

However after

Screenshot 2024-12-16 at 23 15 17

----

```
  public static final java.lang.String example(org.example.A);
    descriptor: (Lorg/example/A;)Ljava/lang/String;
    Code:
      stack=2, locals=4, args_size=1
         0: aload_0
         1: dup
         2: ifnull        27
         5: astore_1
         6: aload_1
         7: astore_2
         8: iconst_0
         9: istore_3
        10: invokestatic  #16                 // Method exception:()V
        13: aload_1
        14: invokevirtual #22                 // Method org/example/A.getB:()Lorg/example/B;
        17: dup
        18: ifnull        27
        21: invokevirtual #28                 // Method org/example/B.getC:()Ljava/lang/String;
        24: goto          29
        27: pop
        28: aconst_null
        29: areturn
```

branch `0` of `ifnull` instructions of a chain should not be replaced.


leveretka (approved):
Looks good

PR #1822, opened 2024-12-17
Get rid of cycle Filters.allKotlinFilters → SyntheticFilter → Filters.isKotlinClass

Godin:

PR #1823, opened 2024-12-21
Get rid of cycle Offline → ConfigLoader.load → Offline.class

Godin:


Godin:
@marchof I had a look at this cycle as a follow-up of https://github.com/jacoco/jacoco/pull/1822#issuecomment-2556884638 and AFAICS this cycle has been there since the introduction of `ConfigLoader` by commit 47990a4a02bb738eedd7dce6a2f2dd10d5f28cc4 : `RT` → `ConfigLoader.load` → `RT.class`, then `RT` was renamed to `Offline` by commit 2a8b5339c422de268cae3a862a85666a9b0ac176 What I don't see - is any reasons for `ConfigLoader` to use `Offline.class.getResourceAsStream` instead of `ConfigLoader.class.getResourceAsStream`, tests also pass after this change. But maybe I and tests miss something, could you please double-check?

Godin:

marchof (approved):

Issue #1862, opened 2025-03-30
Update to ASM 9.8


merks:
### Scenario

There is now a new ASM version 9.8 available:

https://repo1.maven.org/maven2/org/ow2/asm/asm/9.8

It would be nice to have  org.jacoco artifacts available at Maven central compatible with the latest ASM version.

### Current Behaviour

Can't work with latest version of ASM.

### Wanted Behaviour

Would like to work in combination with latest version of ASM.

### Possible Workarounds

Use older version of ASM.

PR #1863, opened 2025-03-30
Update to ASM 9.8

merks:
https://github.com/jacoco/jacoco/issues/1862


iloveeclipse:
@Godin : could we have an approval?

marchof (approved):

PR #1875, opened 2025-04-08
Improve filtering of Kotlin suspending functions with tail call optimization

### LINKED ISSUE #1036: Closing branch of suspend unit functions marked as not covered
### Steps to reproduce

* JaCoCo version:  0.8.5 / 0.8.6.202003291235 (snapshot)
* Operating system: macOS 10.14.5
* Tool integration: Gradle
* Complete executable reproducer: https://github.com/fondesa/jacoco-suspend-bug-poc
* Steps: described in the readme of the repository

### Expected behaviour
The closing branch of a `Unit` suspend function which invokes another `Unit` suspend function as the last instruction, should be marked as covered if the last instruction was executed.

### Actual behaviour
The closing branch of a `Unit` suspend function which invokes another `Unit` suspend function as the last instruction, is always marked as not covered.

Godin:
Fixes #1036
Closes #1019

### DIFF
--- a/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinCoroutineTailCallOptimizationTarget.kt
+++ b/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinCoroutineTailCallOptimizationTarget.kt
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2025 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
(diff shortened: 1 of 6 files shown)


org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/targets/KotlinCoroutineTailCallOptimizationTarget.kt
```diff
@@ -0,0 +1,48 @@
...
+ * Test target with suspending function with tail call optimization.
+ */
+object KotlinCoroutineTailCallOptimizationTarget {
+
+    suspend fun unitReturning(b: Boolean) { // assertEmpty()
+        if (b) // assertFullyCovered(0, 2)
+            suspension() // assertFullyCovered()
+        else // assertEmpty()
+            noSuspension() // assertFullyCovered()
+    } // assertEmpty()
```
leveretka, line 29: I was looking at the test and the implementation.
It looks like with this implementation we'll ignore the return from the function. So we won't show it as red if, for example `suspension()` function throws an exception. Am I right or I am missing anything here?

Nevertheless, I don't think I have a better idea how to workaround this, so this solution is still better than the current behavior(at the moment it's always red if the last suspending function call contains suspending function invocations).

PR #1877, opened 2025-04-09
Improve `FilterTestBase.assertIgnored`

Godin:
Similarly to changes of `FilterTestBase.assertReplacedBranches` in https://github.com/jacoco/jacoco/pull/1813

### before

Screenshot 2025-04-09 at 15 52 57

### after

Screenshot 2025-04-09 at 15 51 33

### before

Screenshot 2025-04-09 at 14 41 48

### after

Screenshot 2025-04-09 at 14 47 20

### before

Screenshot 2025-04-09 at 16 08 12

### after

Screenshot 2025-04-09 at 16 07 05

### DIFF
--- a/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterTestBase.java
+++ b/org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterTestBase.java
@@ -12,12 +12,12 @@
  *******************************************************************************/
 package org.jacoco.core.internal.analysis.filter;
 
-import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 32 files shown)


org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterTestBase.java
```diff
@@ -58,12 +58,30 @@ public void replaceBranches(final AbstractInsnNode source,
...
+				rangesToString(methodNode, Arrays.asList(expected)),
+				rangesToString(methodNode, ignoredRanges));
+	}
+
+	private String rangesToString(final MethodNode m,
+			final List<Range> ranges) {
+		final StringBuilder stringBuilder = new StringBuilder();
+		for (int i = 0; i < ranges.size(); i++) {
+			final Range range = ranges.get(i);
+			stringBuilder.append("range ").append(i)
```
marchof, line 73: Maybe add `Range.toString()`?

Godin: @marchof it can be `Range.toString(Method)`, but not just `Range.toString` because here we need index of instruction within method - is it what you mean?

marchof: Right... in this case I don't have a strong preference. With `Object.toString()` one could simply convert the array to a list and use its `toString()` method.

marchof (approved):

Issue #1896, opened 2025-06-06
Migrate to the Central Publisher Portal


Godin:
See https://central.sonatype.org/news/20250326_ossrh_sunset/

> The OSSRH service will reach end-of-life on June 30th, 2025.

PR #1897, opened 2025-06-06
Migrate publishing of snapshot to the Central Publisher Portal

Godin:
Part of #1896 

https://central.sonatype.org/publish/publish-portal-snapshots/#publishing-via-other-methods

----

https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/org/jacoco/jacoco/


marchof (approved):
Thanks for taking care of this!

PR #1906, opened 2025-06-29
Option for method level only probing 

AntiGameZ:


marchof:
@AntiGameZ Also to be honest I'm a bit reluctant to spend time on an AI generated PR which completely ignores our existing unit tests and user documentation. I don't feel like spending my personal free time for other's laziness.

PR #1915, opened 2025-07-13
Provide agent default values as constants for documentation

### LINKED ISSUE #1676: Feature Request: document that the prepare-agent append configuration is true by default

marchof:
Try to fix #1676 for all properties.


kwin (commented):
LGTM, that would help users a lot.

Issue #921, opened 2019-08-16
Kotlin: Incorrect branch coverage with nested elvis operator


nroi:
### Steps to reproduce

Test a method where the elvis operator is nested. Call this method both with a null input and a non-null input:

```Kotlin
@Test 
fun testElvis() {
    elvisNested("foo")
    elvisNested(null)
}

fun elvisNested(input: String?): String? {
    return input
        ?.trim()
        ?.capitalize()
}
```

### Expected behaviour

There are two possible branches for the line
```Kotlin
?.trim()
```
since this method either is called, or it isn't. Since the method receives both null and non-null as input, both cases are covered, so I would have expected both branches to be covered.

### Actual behaviour
It says  "1 out of 2 branches missed" for the line `?.trim()`.

JaCoCo version: 0.8.4
Operating system: Linux
Tool integration: Gradle
Kotlin version: 1.3.41

Godin:
`?:` is ["elvis operator"](https://kotlinlang.org/docs/reference/null-safety.html#elvis-operator)

`?.` is ["safe call operator"](https://kotlinlang.org/docs/reference/null-safety.html#safe-calls)

JaCoCo performs analysis of bytecode. Compilation of following `Example.kt`

```
object Example {
    fun example(input: String?): String? {
        return input
                ?.t() // line 4
                ?.t() // line 5
    }
}

fun String.a(): String = this
fun String.b(): String = this

fun main() {
    Example.example("foo")
    Example.example(null)
}
```

using Kotlin compiler versions [1.3.41](https://github.com/JetBrains/kotlin/releases/tag/v1.3.41) and [1.3.50-eap-86](https://github.com/JetBrains/kotlin/releases/tag/v1.3.50-eap-86) with and without new type inference (`kotlinc -d classes Example.kt` and `kotlinc -Xnew-inference -d classes Example.kt`) 
results in following bytecode (`javap -v -p classes/Example.class`)

```
  public final java.lang.String example(java.lang.String);
    descriptor: (Ljava/lang/String;)Ljava/lang/String;
    flags: (0x0011) ACC_PUBLIC, ACC_FINAL
    Code:
      stack=2, locals=2, args_size=2
         0: aload_1
         1: dup
         2: ifnull        18
         5: invokestatic  #12                 // Method ExampleKt.a:(Ljava/lang/String;)Ljava/lang/String;
         8: dup
         9: ifnull        18
[...]

PR #1954, opened 2025-09-23
KotlinSafeCallOperatorFilter should filter "unoptimized" safe call followed by elvis

Godin:
This was overlooked in https://github.com/jacoco/jacoco/pull/1814
See https://github.com/jacoco/jacoco/issues/921#issuecomment-3319067842

----

For https://github.com/SonarSource/sonar-kotlin/tree/2.22.0.5972
prior to this change we find 102 safe call operator chains
and after 114
ie 12 more

For http://github.com/JetBrains/kotlin/tree/v2.0.0
prior to this change 5042
and after 5323
ie 281 more

### DIFF
--- a/org.jacoco.build/pom.xml
+++ b/org.jacoco.build/pom.xml
@@ -450,7 +450,7 @@
         <plugin>
           <groupId>com.diffplug.spotless</groupId>
           <artifactId>spotless-maven-plugin</artifactId>
-          <version>2.46.1</version>
+          <version>3.0.0</version>
        </plugin>
        <plugin>
          <groupId>org.sonarsource.scanner.maven</groupId>


org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinSafeCallOperatorFilter.java
```diff
@@ -132,18 +187,23 @@ private static Collection<ArrayList<JumpInsnNode>> findChains(
 	}
 
 	/**
-	 * @return non pseudo-instruction preceding given
+	 * @return non pseudo-instruction preceding given iff it has given opcode,
```
leveretka, line 190: is `iff` necessary here? It seems to me just `if` + `otherwise` is good enough explanation.

org.jacoco.core/src/org/jacoco/core/internal/analysis/filter/KotlinSafeCallOperatorFilter.java
```diff
@@ -108,13 +133,43 @@ private static Collection<ArrayList<JumpInsnNode>> findChains(
...
 				if (chain != null) {
-					final AbstractInsnNode p2 = preceding(p1);
-					if (p2 == null || p2.getOpcode() != Opcodes.ASTORE
+					final AbstractInsnNode p2 = preceding(p1, Opcodes.ASTORE);
+					if (p2 == null
+							|| ((VarInsnNode) p1).var != ((VarInsnNode) p2).var) {
+						continue;
+					}
+				}
+			} else if (target.getType() != AbstractInsnNode.JUMP_INSN
```
leveretka, line 147: COmplexity of this `if` chain is too high. Maybe you can split the logic somehow? This will make the code more readable.

PR #1959, opened 2025-09-28
Kotlin since version 2.2.0 support compilation into Java 24 bytecode

Godin:


org.jacoco.core.test.validation/pom.xml
```diff
@@ -594,7 +593,7 @@
         </property>
       </activation>
       <properties>
-        <!-- Kotlin 2.1.0 doesn't support compilation into 26 -->
+        <!-- Kotlin 2.2.20 doesn't support compilation into 26 -->
         <kotlin.compiler.jvmTarget>23</kotlin.compiler.jvmTarget>
```
marchof, line 597: Can't we also compile to Java 24 bytecode here?

marchof (approved):

Issue #1964, opened 2025-10-05
Update to ASM 9.9


merks:
### Scenario

There is now a new ASM version 9.9 available:

https://repo1.maven.org/maven2/org/ow2/asm/asm/9.9

It would be nice to have  org.jacoco artifacts available at Maven central compatible with the latest ASM version.

### Current Behaviour

Can't work with latest version of ASM.

### Wanted Behaviour

Would like to work in combination with latest version of ASM.

### Possible Workarounds

Use older version of ASM.

PR #1965, opened 2025-10-05
Update to ASM 9.9

merks:
https://github.com/jacoco/jacoco/issues/1964


org.jacoco.build/licenses/asm-9.9.html
```diff
@@ -1,7 +1,7 @@
 <h4>ASM</h4>
 
 <p>
-  <a href="https://asm.ow2.io/">ASM 9.8</a> is subject to the terms and
+  <a href="https://asm.ow2.io/">ASM 9.9</a> is subject to the terms and
```
Godin, line 4: I did check that their license stayed unchanged.

Godin (approved):
For the record
here is a link to their changelog - https://asm.ow2.io/versions.html
and actual changes between versions - https://gitlab.ow2.org/asm/asm/-/compare/ASM_9_8...ASM_9_9

From this to me seems that this time there were no changes that could negatively impact us.

org.jacoco.build/licenses/asm-9.9.html
merks: Yes, I did check the license file too

https://asm.ow2.io/license.html

I was a bit surprised that `ASM is released under the following 3-Clause BSD License:` was left out of the copy:

image

iloveeclipse:
Apropos: will there be an Eclemma update soon?

merks:
I'm still awaiting the Maven Central update for Orbit.  I image the Maven Central publishing is a prerequisite for Eclemma as well.

I know that sometimes it can take a while from when it's published to when it's visible here:

https://repo.maven.apache.org/maven2/org/jacoco/org.jacoco.agent/

Godin:
@merks we do not publish releases for every merge of every PR 😉 but we do publish snapshots at https://central.sonatype.com/repository/maven-snapshots/ which unfortunately as of now doesn't allow to get list of files in browser, but they are downloadable by Maven.

It took me quite some time, much more than usually 😪 , because of EOL of OSSRH service ( oss.sonatype.org ) and migration to Central Publisher Portal (central.sonatype.com) - https://github.com/jacoco/jacoco/issues/1896, but release has been staged - https://groups.google.com/g/jacoco-dev/c/wdKGT0rprS0/m/q9MlqmKzAQAJ

So hope it will reach Maven Central tomorrow after successful validation by @marchof 

Then after Orbit update, publishing of EclEmma milestone should not take long for me.

merks:
Thank you for your effort and for the updates.  It's **all very much appreciated**. ❣️

PR #1944, opened 2025-10-24
Add filter for Kotlin `@JvmExposeBoxed`

Godin:
See
* https://kotlinlang.org/docs/whatsnew22.html#improved-java-interop-with-inline-value-classes
* https://github.com/Kotlin/KEEP/blob/jvm-expose-boxed/proposals/jvm-expose-boxed.md


leveretka (approved):
Looks good, I'd suggest renaming method "hasAnnotation" to "hasExposedBoxedAnnotation" to be more explicit. Will leave it up to you.

PR #1985, opened 2025-10-28
Kotlin SMAPs in synthetic classes should be processed

Godin:


marchof:
Many thanks for the detailed explanation!

marchof (approved):

marchof:
@Godin I don't see a use case of `package-info.class` files in our reports, especially as they never contain code. I would therefore hard-code the filtering of module-info and package-info in the analyzer.

PR #2004, opened 2025-11-26
Add experimental support for Java 27

Godin:
To be merged when JDK 27 development starts in December.


Godin:
Quoting https://gitlab.ow2.org/asm/asm/-/merge_requests/445#note_117501

> Eric Bruneton: Are we sure there won't be any class file format change for V27?
> Remi Forax: yes, value classes should land in Java 28

PR #2008, opened 2025-11-27
Add validation test for Java 22 JEP 456: unnamed patterns

Godin:


marchof:
@Godin I wonder whether we should create validation tests suites for Java versions which are

a) end of live
b) there is already a newer LTS version

So I propose to add a validation test suite for Java 25 which contains all language features from 22 to 25.

Of course this also raises the questions whether we also could drop builds for the intermediate EOL versions at all...

marchof (approved):

marchof (approved):

PR #1698, opened 2024-12-17
Add validation test for `Serializable` lambda in Java

Godin:

PR #2047, opened 2026-02-03
Add validation test for method references in Java

Godin:


ponyrramirez-hash (commented):
org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/targets/MethodReferencesTarget.java

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,105 @@
...
+			throws IOException {
+		final HashSet<String> names = new HashSet<>();
+		for (final Method method : MethodReferencesTarget.class
+				.getDeclaredMethods()) {
+			if (method.isSynthetic()) {
+				names.add(method.getName());
+			}
+		}
+
+		if (isJDKCompiler) {
```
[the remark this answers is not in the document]
Godin: @marchof for consistency with already existing similar checks introduced by me to check shape of bytecode that might be filtered out and hence won't appear in our IClassCoverage nodes - for example https://github.com/jacoco/jacoco/blob/v0.8.14/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/KotlinLambdaSerializableTest.java#L34-L45 introduced by https://github.com/jacoco/jacoco/pull/1698 reviewed by you 😉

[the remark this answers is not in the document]
Godin: @marchof also in this particular case I'm interested to check that compiler generates specifically **synthetic** methods and not just methods (afaik/afair this flag is not visible in our IClassCoverage nodes), so that this validation test can be easily discovered when one wants to modify filtering around synthetic methods - to me validation tests serve as a reference and reminder about various language constructs and their compiled variants/quirks.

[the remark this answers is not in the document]
Godin: @marchof speaking in general: change in compiler and more importantly change in validation test can accidentally make it irrelevant as test of corresponding filter. This actually happened to me few times in past during updates of some filters. So ideally I dream about more stronger connection between validation tests, filters and unit tests of filters than current connection of manual inspection of compiler output during work on some filter. I even have some idea and PoC about this, but it is not ready for prime/show time and more work/time is needed to polish it compared to simple use of reflection for now 😉

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,105 @@
...
+				assertEquals(2, names.size());
+			}
+		} else {
+			assertTrue(names.contains("lambda$2"));
+			assertTrue(names.contains("lambda$3"));
+			assertTrue(names.contains("lambda$8"));
+			assertEquals(3, names.size());
+		}
+		assertMethodCount(/* constructors */ 4 + /* methods */ 3
+				+ /* lambdas */ names.size());
```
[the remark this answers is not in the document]
Godin: @marchof Thank you for this good suggestion!

However seems to be one of the cases where one need to fight a bit with Eclipse Formatter for better readability 😅 because it reformats your suggestion as

```java
		assertMethodCount(4 // constructors
				+ 3 // methods
				+ names.size() // lambdas
		);
```

But maybe you know if possible to adjust formatter configuration so that it will do exactly as in your suggestion? 😉 

Another option with warning about simplification of expression 😞 

```java
		assertMethodCount(0 //
				+ 4 // constructors
				+ 3 // methods
				+ names.size() // lambdas
		);
```

Other warning-free options

```java
		assertMethodCount(4 // constructors
				+ 3 // methods
				+ names.size() // lambdas
		);
```

```java
		assertMethodCount( //
				4 // constructors
						+ 3 // methods
						+ names.size() // lambdas
		);
```

```java
		assertMethodCount( //
				names.size() // lambdas
						+ 4 // constructors
						+ 3 // methods
		);
```

If not possible to adjust formatter configuration, then personally I like the last option the most. WDYT?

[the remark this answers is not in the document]
marchof: I'm sorry, I didn't want to cause any trouble with formatting 🙈

Yes, I like the last version too.

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/targets/MethodReferencesTarget.java
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}
+
+	static class PrivateMethod {
+		private static void m() {
+		}
+	}
+
+	public static void main(String[] args) {
+		exec(PublicConstructor::new); // assertFullyCovered()
```
marchof, line 48: For completeness: Does it make sense to add tests for reference to bound and unbound instance methods? I wrote down the different method references some time ago: https://javaalmanac.io/features/method-references/

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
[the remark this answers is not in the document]
Godin: @marchof I applied last variant
and IMO this PR ready for approval 😉

marchof (approved):

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/targets/MethodReferencesTarget.java
Godin: @marchof done

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,141 @@
...
+			final Source.Line line) {
+		if (isJDKCompiler) {
+			assertPartlyCovered(line);
+		} else {
+			assertFullyCovered(line);
+		}
+	}
+
+	/**
+	 * TODO report to ECJ that it generates LINENUMBER 1
```
marchof, line 63: Report and add bug reference here?

Godin: @marchof I totally agree and since the beginning planned to do so when will have some spare time for this and that's why added todo-comment. BTW don't hesitate to do it yourself if you have time 😉 

At the same time similarly to https://github.com/jacoco/jacoco/pull/2099#issuecomment-4316507024 with some additions:

To be confirmed, but this behaviour in ECJ might exist for ages across many its versions, and reporting it in my opinion shouldn't be blocking merge of this PR.

Sometimes feel that even the most (if not all) of software development industry after switch from Subversion to Git and apperance of easy and cheap branching started to forget/forgot principles and benefits of early and often integration that increase velocity and not necessarily decrease quality. Even merge of a test that demonstrates wrong/unexpected behavior (and not just for external project such as ECJ, but even for our own code) is much better to be done immediately after discovery, what might be far before its actual fix. In our case much easier to track changes of external behavior(s) when test already landed into main branch rather than sitting in PR.

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,141 @@
...
+			}
+		}
+
+		if (isJDKCompiler) {
+			// array constructor method references
+			assertTrue(names.contains("lambda$main$0"));
+			assertTrue(names.contains("lambda$main$1"));
+			if (bytecodeVersion() < Opcodes.V15) {
+				// constructor, static, bound and unbound
+				// private method references
```
Godin:
```suggestion
				// references to private constructor,
				// private static, private bound and
				// private unbound methods
```

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,141 @@
...
+		final HashSet<String> accessors = new HashSet<>();
+		for (final Method method : MethodReferencesTarget.PrivateMethod.class
+				.getDeclaredMethods()) {
+			if (method.isSynthetic()) {
+				accessors.add(method.getName());
+			}
+		}
+
+		if (isJDKCompiler) {
+			// array constructor method references
```
Godin:
```suggestion
			// references to array constructor
```

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,141 @@
...
+					// JEP 181: Nest-Based Access Control
+					assertEquals(0, accessors.size());
+				}
+			} else {
+				assertEquals(2, names.size());
+				assertEquals(0, accessors.size());
+			}
+		} else {
+			// array constructor method reference
+			// private constructor method references
```
Godin:
```suggestion
			// references to array constructor,
			// and private constructor
```

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
Godin: BTW note that ECJ unlike javac produces smaller bytecode by deduplication of references to array constructor - 2 in source code, while 1 lambda method in bytecode. So in fact actually to me even unclear what should be the expected behavior in this case - same non-1 but not always related to line of reference number, no line number, zero? all of them seem equally not good for us.

org.jacoco.core.test.validation.java8/src/org/jacoco/core/test/validation/java8/MethodReferencesTest.java
```diff
@@ -0,0 +1,141 @@
...
+			assertTrue(names.contains("lambda$2"));
+			assertTrue(names.contains("lambda$3"));
+			assertTrue(names.contains("lambda$4"));
+			assertEquals(3, names.size());
```
Godin, line 129:
```suggestion
			// note that ECJ unlike javac performs
			// deduplication of references to array constructor
			assertEquals(3, names.size());
```

PR #2099, opened 2026-04-08
Use try-catch idiom for testing exceptions

Godin:
This was discussed multiple times in past:
* https://github.com/jacoco/jacoco/pull/1057#discussion_r664138632
* https://github.com/jacoco/jacoco/pull/618#discussion_r150205478

----

Quoting https://github.com/junit-team/junit4/wiki/exception-testing

> If you project is not yet using JUnit 4.13 or your code base does
> not support lambdas, you can use the try/catch idiom

> The `expected` parameter should be used with care. The test will
> pass if any code in the method throws expected Exception. Also
> cannot test the value of the message in the exception, or the state
> of a domain object after the exception has been thrown.

### DIFF
--- a/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/MockSocketConnectionTest.java
+++ b/org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/MockSocketConnectionTest.java
@@ -18,7 +18,6 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
-import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetAddress;
@@ -135,8 +134,8 @@ public class MockSocketConnectionTest extends ExecutorTestBase {
 		a.close();
 		try {
 			a.getOutputStream();
-			fail("IOException expected");
-		} catch (final IOException e) {
+			fail("SocketException expected");
+		} catch (final SocketException e) {
 			// expected
 		}
 	}
@@ -147,8 +146,8 @@ public class MockSocketConnectionTest extends ExecutorTestBase {
 		a.close();
 		try {
 			out.write(123);
-			fail("IOException expected");
-		} catch (final IOException e) {
+			fail("SocketException expected");
+		} catch (final SocketException e) {
 			// expected
 		}
 	}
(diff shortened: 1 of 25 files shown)


org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/MockSocketConnectionTest.java
```diff
@@ -108,23 +119,38 @@ public Void call() throws Exception {
...
-	@Test(expected = SocketException.class)
+	@Test
 	public void testWriteOnClosedSocket() throws Exception {
 		final OutputStream out = a.getOutputStream();
 		a.close();
-		out.write(123);
+		try {
+			out.write(123);
+			fail("IOException expected");
+		} catch (final IOException e) {
```
marchof: Here you generalize from `SocketException` to `IOException`. Is that on purpose?

org.jacoco.core.test/src/org/jacoco/core/data/ExecutionDataReaderWriterTest.java
```diff
@@ -146,31 +152,55 @@ public void testInvalidVersion() throws IOException {
 		final char version = (char) (ExecutionDataWriter.FORMAT_VERSION - 1);
 		buffer.write(version >> 8);
 		buffer.write(version & 0xFF);
-		createReader().read();
+		final ExecutionDataReader reader = createReader();
+		try {
+			reader.read();
+			fail("IncompatibleExecDataVersionException expected");
+		} catch (final IncompatibleExecDataVersionException e) {
```
marchof, line 159: Now that we have the exception at hand we could assert `actualVersion`.

org.jacoco.agent.rt.test/src/org/jacoco/agent/rt/internal/output/MockSocketConnectionTest.java
Godin: @marchof oups, definitely not intentional - probably copy-paste from another place 😞 will fix it. Thanks for catching!

marchof (approved):
I consider this as an intermediate step until we finally upgrade to Java 8 and can use JUnit 5 where we have a less verbose syntax.

I wonder whether we should now verify some of the exceptions (message, cause etc), especially when we throw the exception in out code.

Godin:
> I wonder whether we should now verify some of the exceptions (message, cause etc), especially when we throw the exception in out code.

Indeed, but I prefer small baby-step increments rather than never ending improvements within one PR,
and to be honest I quite often feel that we lack the first approach and fall into trap(s) of second, even if our own conventions https://github.com/jacoco/jacoco/blob/v0.8.14/org.jacoco.doc/docroot/doc/conventions.html#L161-L163 state

> Clear scope: We track and review every semantical change through a separate pull request. Pull requests handling various topics ("I fixed this and that") are typically difficult in handling and are therefore declined.

After this PR one can search for empty catch blocks

```java
catch (...) {
  // expected
}
```

and choose what to assert in each on a case by case basis.
Can even create ticket for this.

Even this pull request is already a follow-up on comments in very old previous ones (4 and 9 years) and even without ticket 😜 btw we managed to live quite successfully without such assertions in these cases for such a big amount of time 😉

PR #2104, opened 2026-04-26
Fix typo

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/LineImpl.java
```diff
@@ -45,7 +45,7 @@ public abstract class LineImpl implements ILine {
 	}
 
 	/**
-	 * Empty line without instructions or branches.
+	 * Empty line without instructions and branches.
```
marchof, line 48: Boolean logic :)

```suggestion
	 * Empty line without instructions and without branches.
```

marchof (approved):

org.jacoco.core/src/org/jacoco/core/internal/analysis/LineImpl.java
Godin: @marchof while I like your proposal, unfortunately I enabled auto-merge prior to request of review, so PR got merged at the moment of approval prior to the moment when I noticed suggestion 😞  WDYT about other one containing this suggestion?

marchof: Sorry, I didn't notice auto-merge was on and as I didn't had a strong feeling here I didn't want to block you. Sure, we can still fix it in a separate PR or sneak it in with your upcoming work on singletons.

PR #2109, opened 2026-05-05
Use StringPool for class names from Kotlin SMAPs

Godin:
This was overlooked in https://github.com/jacoco/jacoco/commit/831ecd609e17a3a5f8e114134fa0ce1d2071c6bd


org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -153,15 +153,18 @@ private void calculateFragments(
 			// that's why parsing is done here
 			smap = new KotlinSMAP(getSourceFileName(), sourceDebugExtension);
```
marchof: I wonder if we should use the `StringPool` already within `KotlinMap` so that the mappings already have normalized string instances.

Godin: @marchof to do so `KotlinSMAP` should not be also constructed in `KotlinInlineFilter`, which is anyway a good thing on its own - #2114

marchof: So now that `KotlinMap` is created within `ClassAnalyzer`: Do we want to use `StringPool` in it?

marchof (approved):

PR #2110, opened 2026-05-07
Add validation test for Kotlin boolean expressions

Godin:
We have such [test for Java](https://github.com/jacoco/jacoco/blob/v0.8.14/org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/targets/BooleanExpressionsTarget.java), so I think we should add similar for Kotlin too.

It will also be useful for https://github.com/jacoco/jacoco/pull/1680


leveretka (approved):
LGTM!

PR #2114, opened 2026-05-09
Parse Kotlin SMAP only once for each class

Godin:


org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -104,6 +104,13 @@ public MethodProbesVisitor visitMethod(final int access, final String name,
 			final String desc, final String signature,
 			final String[] exceptions) {
 
+		if (sourceDebugExtension != null && Filters.isKotlinClass(this)) {
+			// Note that visitSource is invoked before visitAnnotation,
```
marchof: Thanks for the clarifying comment. Indeed it looks a bit strange that the SMAP calculation is now in `visitMethod()`.

WDYT about calulating it lazily in `getKotlinSMAP()`? Then we still have the assumption that the method is called after `visitSource()` and `visitAnnotation()` but at least we have a more intuitive location to perform the calculation.

marchof: In this context but for a separate PR to improve readablity: Maybe our ASM `*Visitor` implementations should have the `visit*()` methods always in the source files ordered like the call sequence defined by ASM.

Godin: @marchof actually then might be better to do parsing in `visitAnnotation` when it encounters `Lkotlin/Metadata;` 😉

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -84,6 +84,12 @@ public void visit(final int version, final int access, final String name,
 	@Override
 	public AnnotationVisitor visitAnnotation(final String desc,
 			final boolean visible) {
+		if (desc.equals("Lkotlin/Metadata;") && sourceDebugExtension != null) {
```
marchof: We have a constant for this: `KotlinGeneratedFilter.KOTLIN_METADATA_DESC`

What about moving the constant to `KotlinSMAP` and also use it here?

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -84,6 +84,12 @@ public void visit(final int version, final int access, final String name,
...
+			// Note that visitSource is invoked before visitAnnotation,
+			// that's why parsing is done here
+			smap = new KotlinSMAP(getSourceFileName(), sourceDebugExtension);
+			sourceDebugExtension = null;
```
marchof, line 92: Do we still need to set this to `null`? I think we can safely assume that there is only one `Lkotlin/Metadata;` annotation as consider class files as valid.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
```diff
@@ -84,6 +84,12 @@ public void visit(final int version, final int access, final String name,
 	@Override
 	public AnnotationVisitor visitAnnotation(final String desc,
 			final boolean visible) {
+		if (desc.equals("Lkotlin/Metadata;") && sourceDebugExtension != null) {
+			// Note that visitSource is invoked before visitAnnotation,
```
marchof, line 89: Not sure whether the comment now still makes sense in this form as we always rely on the defined sequence.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
marchof: Is it a valid scenario that we have a `Lkotlin/Metadata;` annotation but no `sourceDebugExtension`? If not I think we can fail with an exception instead of producing unexpected results.

Godin: @marchof yes - it is possible to have `Lkotlin/Metadata;` annotation but no `SourceDebugExtension` attribute - in absence of Kotlin `inline` functions.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof IMO still makes sense since parsing should not be done in `visitSource` prior to `visitAnnotation` - there might be `SourceDebugExtension` in non-Kotlin class that should not be parsed

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/filter/FilterContextMock.java
```diff
@@ -27,6 +27,7 @@ public class FilterContextMock implements IFilterContext {
 	public Set<String> classAttributes = new HashSet<String>();
 	public String sourceFileName = "Foo.java";
 	public String sourceDebugExtension;
+	public KotlinSMAP kotlinSMAP;
```
Godin:
```suggestion
	public KotlinSMAP kotlinSMAP;
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof or it can be moved to and reused from `Filters` which is the only other place were it is used. In either case let's first remove its unnecessary uses - #2116

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof this `null`-ing allows GC to release memory allocated for `sourceDebugExtension` once it is parsed

marchof: Well, but the whole `ClassAnalyzer` is a short living object.

org.jacoco.core.test/src/org/jacoco/core/internal/analysis/ClassAnalyzerTest.java
```diff
@@ -198,15 +194,12 @@ public void should_not_parse_absent_SourceDebugExtension_attribute_when_kotlin()
...
 		analyzer.visitSource("Foo.kt", "SMAP\n");
-		analyzer.visitAnnotation("Lkotlin/Metadata;", false);
-		final MethodNode mn = new MethodNode(0, "foo", "()V", null, null);
-		final MethodProbesVisitor mv = analyzer.visitMethod(mn.access, mn.name,
-				mn.desc, mn.signature, mn.exceptions.toArray(new String[0]));
 		try {
-			mv.accept(mn, mv);
+			analyzer.visitAnnotation("Lkotlin/Metadata;", false);
 			fail("exception expected");
 		} catch (Exception e) {
```
marchof: Should we test for the specific exception type?

```suggestion
		} catch (IllegalStateException e) {
```

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof are you ready to bet that lifetime is short enough to not cross boundaries of multiple minor GCs so that this allocation has no chance to leave TLAB and Eden space? 😉 and while I don't have proof of opposite, if it happens IMO better to increase chances to reclaim this space prior to processing of method bodies which also allocates many short living objects. Also note that SMAP string is not the most compact representation of information that it contains.

And `null`-check is likely cheaper that `String`-comparison, so in fact I'd rather flip operands of `&&` in `if`.

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof I added comment explaining this to the corresponding test

org.jacoco.core/src/org/jacoco/core/internal/analysis/ClassAnalyzer.java
Godin: @marchof I added comment explaining this to the corresponding test

marchof (approved):

PR #2115, opened 2026-05-09
Do not execute `LineImplMemoryTest.currentVM` on 32-bit platforms such as AArch32

### LINKED ISSUE #2113: Build Failure on ARM32
Since 8a3d2c258792ba1d0f057780c432f2ccafa66cf1 builds on ARM 32 fail:

```
[ERROR] org.jacoco.core.internal.analysis.LineImplMemoryTest.currentVM -- Time elapsed: 0.011 s <<< FAILURE!
org.junit.ComparisonFailure: 
expected:<...        VALUE
  0   [8                                                 (object header: mark)     N/A
  8   4                                                 (object header: class)    N/A
 12   4   org.jacoco.core.internal.analysis.CounterImpl LineImpl.instructions     N/A
 16   4   org.jacoco.core.internal.analysis.CounterImpl LineImpl.branches         N/A
 20   4                                                 (object alignment gap)    
Instance size: 24 bytes
Space losses: 0 bytes internal + 4 bytes external = 4] bytes total
> but was:<...        VALUE
  0   [4                                                 (object header: mark)     N/A
  4   4                                                 (object header: class)    N/A
  8   4   org.jacoco.core.internal.analysis.CounterImpl LineImpl.instructions     N/A
 12   4   org.jacoco.core.internal.analysis.CounterImpl LineImpl.branches         N/A
Instance size: 16 bytes
Space losses: 0 bytes internal + 0 bytes external = 0] bytes total
```

@Godin I wonder whether this platform should be considered in our code base (for example by making assumptions in the `currentVM` test). Otherwise I can simply exclude the the benchmark module.

Godin:
Fixes #2113

### DIFF
--- a/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
+++ b/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
@@ -222,7 +222,7 @@ public class LineImplMemoryTest {
 		}
 		if ("32".equals(System.getProperty("sun.arch.data.model"))) {
 			throw new AssumptionViolatedException(
-					"this test requires 64-bit architecture");
+					"this test does not support 32-bit architecture");
 		}
 		return new CurrentLayouter();
 	}
--- a/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
+++ b/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
@@ -220,6 +220,10 @@ public class LineImplMemoryTest {
 			}
 			throw e;
 		}
+		if ("32".equals(System.getProperty("sun.arch.data.model"))) {
+			throw new AssumptionViolatedException(
+					"this test requires 64-bit architecture");
+		}
 		return new CurrentLayouter();
 	}
 


marchof (approved):

org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
```diff
@@ -220,6 +220,10 @@ private static CurrentLayouter currentLayouter() {
 			}
 			throw e;
 		}
+		if ("32".equals(System.getProperty("sun.arch.data.model"))) {
```
marchof, line 223: For conistency with the assumption message:

```suggestion
		if (!"64".equals(System.getProperty("sun.arch.data.model"))) {
```

Godin: @marchof current condition follows my initial logic (general/usual logic of reducing such exclusions to necessary required minimum) to exclude what we know should be excluded without excluding what might require exclusion if we will encounter non 32 and non 64 bit platforms, so should the message be changed instead or maybe I'm overthinking this?

marchof: Changing the message to be consistent with the current check is also ok. Or let JUnit do the job:

    Assume.assumeThat((System.getProperty("sun.arch.data.model"), CoreMatchers.equalTo("64"));

marchof (approved):

PR #2135, opened 2026-05-23
Add validation test for chained and nested method invocations in Java

Godin:


org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/MethodInvocationsTest.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Test of code coverage in {@link MethodInvocationsTarget}.
+ */
+public class MethodInvocationsTest extends ValidationTestBase {
+
+	public MethodInvocationsTest() {
+		super(MethodInvocationsTarget.class);
+	}
+
+	public void assertMethodInvocation(final Source.Line line) {
```
marchof, line 29: The test case `MethodInvocationsTest` sounds quite generic and may host more tests in future, which is good.

Therefore I propose to name the assert methods more specific, for example:

* `assertMethodInvocation` --> `assertChainedMethodInvocation`
* `assertNonMethodInvocation` --> `assertNonChainedMethodInvocation`

Godin: @marchof here is diff with proposed changes

```diff
index 345aadb59..6cc333330 100644
--- i/org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/targets/MethodInvocationsTarget.java
+++ w/org.jacoco.core.test.validation.java5/src/org/jacoco/core/test/validation/java5/targets/MethodInvocationsTarget.java
@@ -26,17 +26,17 @@ public class MethodInvocationsTarget {

        private static void chained() {
                new MethodInvocationsTarget() // assertFullyCovered()
-                               .chained("first") // assertMethodInvocation()
-                               .chained("second"); // assertMethodInvocation()
+                               .chained("first") // assertChainedMethodInvocation()
+                               .chained("second"); // assertChainedMethodInvocation()
        }

        private static void nested() {
                nested( // assertFullyCovered()
-                               nested("")); // assertMethodInvocation()
+                               nested("")); // assertChainedMethodInvocation()

                /* For comparison with above: */
                nested( // assertFullyCovered()
-                               ""); // assertNonMethodInvocation()
+                               ""); // assertNonChainedMethodInvocation()
        }

        private MethodInvocationsTarget chained(final String s) {
```

and sorry, but IMO original is much better because

* use of "chained" for "nested" seems weird/wrong
* `assertMethodInvocation` and `assertNonMethodInvocation` mean assert on line with and without method invocation respectively, no matter whether method invocation is chained or nested or other
* for lines with method invocation (no matter chained or not) javac generates line number starting only from version 8
* for lines without method invocation (no matter chained or not) javac not always generates line number - eg not in this case, but does in `ControlStructuresTarget.executedForBlock`

marchof: I see your points but I still wonder why the first line here is *not* a "method invocation":

```
	nested( // assertFullyCovered()
		nested("")); // assertMethodInvocation()
```

Godin: @marchof and I see your point 😅

quoting https://bugs.openjdk.org/browse/JDK-7024096

> linenumbertable in compiled code only has line numbers for starts of statements

so what about `assertMethodInvocation` -> `assertMethodInvocationNotOnStatementStartLine` ? 😉

marchof: Just keep its is is and we think about a different name once we have new test cases which conflict with the current naming!

marchof (approved):

PR #2146, opened 2026-06-01
Use cache in GitHub Actions

Godin:
Execution of

```
mvn verify -Dmaven.repo.local=$PWD/maven-local
du -hs maven-local
```

provides rough estimation (due to some differences between builds - eg with/without ECJ)
that execution of each build downloads about 300M from Maven Central,
each GitHub Actions CI run contains 27 builds, so about 8G in total.

Better to cache this in GitHub Actions for the following reasons even if size of matrix might be reduced:

To reduce downloads from Maven Central - see https://central.sonatype.org/faq/429-error/

To be less affected by incidents with access of Maven Central (amount of which might increase due to previous point) - one known to me is quite recent https://github.com/orgs/community/discussions/183607

Also on Linux single job takes on average about 4.3 min,
out of which about 30 sec (13%) is downloading dependencies,
while cache restoration takes just 3-5 sec.
Time benefits seem even bigger on Windows.

Total duration of all jobs in GitHub Actions prior to this change is about 11.5 min and after this change is about 8.6 min (about 25% faster).

### Possible future improvements

* single cache entry for both Windows and Linux
* reduce downloads even further - even in presence of cache populated by `dependency:go-offline` there are still few downloads due to slight differences in configurations between builds (eg with/without ECJ) and dynamic dependencies (such as providers downloaded by maven-surefire-plugin, formatter downloaded by spotless-maven-plugin, and scala compiler)
* at first glance saving of cache for PR builds seems unnecessary
* time based (always/weekly/monthly) cache invalidation for default branch to prevent infinite growth - update of dependency causes restore of previous and save of updated, currently size is 250 MB for Linux and 120 MB for Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,21 @@ jobs:
           java-version: |
             ${{ matrix.jdk }}
             21
+      - &cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.m2/wrapper
+          key: maven-${{ runner.os }}-${{ hashFiles('**/pom.xml', '.mvn/wrapper/maven-wrapper.properties') }}
+          restore-keys: maven-${{ runner.os }}
+      - # Cache will be saved by one job from matrix that completes first,
+        # to maximize number of dependencies in cache
+        # below command resolves statically declared dependencies
+        # (for example misses dynamically downloaded such as formatter for sptoless-maven-plugin)
+        # for modules in active by default profiles
+        # (for example includes dependencies of `org.jacoco.core.test.validation.kotlin`, but misses ecj)
+        run: ./mvnw --quiet dependency:go-offline
       - name: 'Generate toolchains.xml'
         env:
           JDK_VERSION: ${{ matrix.jdk }}
@@ -70,7 +85,7 @@ jobs:
           " > toolchains.xml
       - name: 'Build'
         run: |
-          ./mvnw -V -B -e --no-transfer-progress \
+          ./mvnw -V -B -e \
             verify -Djdk.version=${{ matrix.jdk }} -Dbytecode.version=${{ matrix.jdk }} \
             ${{ matrix.ecj && '-Decj' || ''}} \
             --toolchains=toolchains.xml
@@ -85,6 +100,7 @@ jobs:
           java-version: |
             6
             21
+      - *cache
       - name: 'Generate toolchains.xml'
         env:
           JDK_VERSION: 6
@@ -108,7 +124,7 @@ jobs:
       - name: 'Build'
         shell: bash
         run: |
-          ./mvnw -V -B -e --no-transfer-progress \
+          ./mvnw -V -B -e \
             verify -Djdk.version=6 -Dbytecode.version=5 \
             -f org.jacoco.build \
             ${{ secrets.SONARQUBE_TOKEN && 'sonar:sonar' || '' }} \


.github/workflows/ci.yml
```diff
@@ -49,6 +49,21 @@ jobs:
...
+            ~/.m2/wrapper
+          key: maven-${{ runner.os }}-${{ hashFiles('**/pom.xml', '.mvn/wrapper/maven-wrapper.properties') }}
+          restore-keys: maven-${{ runner.os }}
+      - # Cache will be saved by one job from matrix that completes first,
+        # to maximize number of dependencies in cache
+        # below command resolves statically declared dependencies
+        # (for example misses dynamically downloaded such as formatter for sptoless-maven-plugin)
+        # for modules in active by default profiles
+        # (for example includes dependencies of `org.jacoco.core.test.validation.kotlin`, but misses ecj)
+        run: ./mvnw --quiet dependency:go-offline
```
marchof, line 66: I wonder whether cache prepulation should be better done only once in a separate `Cache` job where `Linux` and `Windows` depends on?

.github/workflows/ci.yml
```diff
@@ -49,6 +49,21 @@ jobs:
...
             ${{ matrix.jdk }}
             21
+      - &cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.m2/wrapper
+          key: maven-${{ runner.os }}-${{ hashFiles('**/pom.xml', '.mvn/wrapper/maven-wrapper.properties') }}
+          restore-keys: maven-${{ runner.os }}
```
marchof, line 59: For my understanding: Why do we need a separate Maven cache for different operating systems? That would mean we get different Maven artifacts with the *same* name for different OSs? Same question for our POMs.

Godin: > That would mean we get different Maven artifacts with the same name for different OSs?

No

> Why do we need a separate Maven cache for different operating systems?

There are additional requirements to make cache cross-os - see https://github.com/actions/cache/blob/v5.0.5/tips-and-workarounds.md#cross-os-cache
And I was planning to satisfy them, but later step-by-step - see "future improvements" in PR description.

> Same question for our POMs.

Not sure I understand this part of question.

Dependencies and their versions are reproducible/fixed in respect to active profiles

* `mvn -Pecj` activates `ecj` profile without which ECJ not downloaded
* `mvn -Dbytecode.version=6` activates `java6-bytecode` profile that excludes `org.jacoco.core.test.validation.kotlin` module, so won't download Kotlin compiler
* etc

Godin: > Dependencies and their versions are reproducible/fixed in respect to active profiles

and of course properties - ie `mvn -Dkotlin.version=2.4.0` overrides version specified in POM

marchof: Ok, but the downloaded files stay unique (if downloaded), so why do we need a different cache for different profiles? Can't they share a single cache?

.github/workflows/ci.yml
Godin: See "future improvements" in description of this PR - this might be one of the ways to achieve them.

Setup proposed currently in this PR is not ideal, but benefits already visible, and as usual I prefer baby-steps.

.github/workflows/ci.yml
Godin: > the downloaded files stay unique

yes

> why do we need a different cache for different profiles? Can't they share a single cache?

They do share.

As you can see in https://github.com/jacoco/jacoco/actions/caches currently there are only two cache entries - one per os

Linux entry is used by all jobs in matrix after it is created/updated by one of them: in absence of cache - they all race to create it, in case of dependency update - they all race to create new one using previous one.

Godin: ah, probably now I see from where question comes...

```
hashFiles('**/pom.xml', '.mvn/wrapper/maven-wrapper.properties')
```

cache entry is immutable once uploaded, so another entry should be created when dependency changes, and that's why cache entry key is based on such hash

marchof: Ok, I think I misunderstood the the functionality of the cache: The entire cache folder is saved and restored, not individual files. So you can't simply add additional files from dependencies.

Thanks for your patience with my stupid questions...

Godin: @marchof glad that clarity was reached!

Wondering if there anything else to do or you just forgot to press button to approve this PR? :wink:

marchof (approved):
I forgot 😔

PR #2157, opened 2026-06-08
Fix typo

Godin:


marchof (approved):

org.jacoco.ant/src/org/jacoco/ant/CoverageTask.java
```diff
@@ -57,7 +57,7 @@ public CoverageTask() {
 	public void addTask(final Task task) {
 		if (childTask != null) {
 			throw new BuildException(
-					"Only one child task can be supplied to the coverge task",
+					"Only one child task can be supplied to the coverage task",
```
Godin, line 60: @marchof this is kinda user-visible change so wondering if it deserves changelog entry? 😅

marchof: Nope, I sincerely hope this is not a breaking change for anyone.

PR #2158, opened 2026-06-09
Fix typo

Godin:
"an" should be used instead of "a" when the
following word starts with a vowel sound.

---

[I did my best last time](https://github.com/jacoco/jacoco/pull/2088), and [before](https://github.com/jacoco/jacoco/pull/1799)... still overlooked this one 😞 so unfortunately doubt that can guarantee that this is the last existing typo of the same kind 😅


marchof (approved):
No change log entry required 😅

PR #2168, opened 2026-06-10
Use strict equality in allowAnywhereFoldCaseAndRen

tejas-ae:
Not sure how often allowAnywhereFoldCaseAndRenumberGroups hits the edge case, but The loose equality here can coerce values in ways that are hard to spot later. this patch tightens it to strict equality. Close this if it’s not worth the noise.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.

### DIFF
--- a/org.jacoco.report/src/org/jacoco/report/internal/html/resources/prettify.js
+++ b/org.jacoco.report/src/org/jacoco/report/internal/html/resources/prettify.js
@@ -483,7 +483,7 @@ window['_pr_isIE6'] = function () {
               + '|\\\\x[A-Fa-f0-9]{2}'  // a hex escape
               + '|\\\\[0-9]+'  // a back-reference or octal escape
               + '|\\\\[^ux0-9]'  // other escape sequence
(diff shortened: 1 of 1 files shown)


org.jacoco.report/src/org/jacoco/report/internal/html/resources/prettify.js
```diff
@@ -483,7 +483,7 @@ window['_pr_isIE6'] = function () {
               + '|\\\\x[A-Fa-f0-9]{2}'  // a hex escape
               + '|\\\\[0-9]+'  // a back-reference or octal escape
               + '|\\\\[^ux0-9]'  // other escape sequence
-              + '|\\(\\?[:!=]'  // start of a non-capturing group
+              + '|\\(\\?[:!==]'  // start of a non-capturing group
```
Godin, line 486: https://regex101.com/?regex=%5C%28%5C%3F%5B%3A%21%3D%3D%5D&testString=&flags=&flavor=pcre2&delimiter=%2F

PR #2170, opened 2026-06-11
Add experimental support for Java 28 class files

### LINKED ISSUE #2162: Add experimental support for Java 28 class files

Godin:
Closes #2162

### DIFF
--- a/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
+++ b/org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
@@ -273,9 +273,9 @@ public final class InstrSupport {
 	 */
 	public static ClassReader classReaderFor(final byte[] b) {
 		final int originalVersion = getMajorVersion(b);
-		if (originalVersion == Opcodes.V26 + 1) {
+		if (originalVersion == Opcodes.V27 + 1) {
 			// temporarily downgrade version to bypass check in ASM
-			setMajorVersion(Opcodes.V26, b);
+			setMajorVersion(Opcodes.V27, b);
 		}
 		final ClassReader classReader = new ClassReader(b);
 		setMajorVersion(originalVersion, b);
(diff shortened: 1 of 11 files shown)


org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
```diff
@@ -273,9 +273,9 @@ public static void push(final MethodVisitor mv, final int value) {
 	 */
 	public static ClassReader classReaderFor(final byte[] b) {
 		final int originalVersion = getMajorVersion(b);
-		if (originalVersion == Opcodes.V26 + 1) {
+		if (originalVersion == Opcodes.V27 + 1) {
 			// temporarily downgrade version to bypass check in ASM
-			setMajorVersion(Opcodes.V26, b);
+			setMajorVersion(Opcodes.V27, b);
```
marchof, line 278: As we need these bytecode version in many places I wonder whether we should introduce constants e.g. in `InstrSupport` or `JaCoCo`. Then this would read:

```suggestion
		if (originalVersion > InstrSupport.LATEST_ASM_BYTECODE_VERSION && originalVersion <= InstrSupport.LATEST_JACOCO_BYTECODE_VERSION) {
// temporarily downgrade version to bypass check in ASM
			setMajorVersion(InstrSupport.LATEST_ASM_BYTECODE_VERSION, b);
```

Godin: @marchof I like this idea a lot 😍 at the same time implementation most likely will entail further discussions - I myself got some thoughts/questions while trying to apply it. So WDYT about merging this PR as is, while I immediately open another one "Simplify future updates of bytecode versions"?

marchof: Sure! The advantage is that the second PR is a pure refactoring for now.

marchof (approved):

org.jacoco.core/src/org/jacoco/core/internal/instr/InstrSupport.java
Godin: @marchof here you go - https://github.com/jacoco/jacoco/pull/2187

PR #2172, opened 2026-06-13
Add validation test for Java 25 JEP 513: Flexible Constructor Bodies

Godin:


org.jacoco.core.test.validation.java25/src/org/jacoco/core/test/validation/java21/FlexibleConstructorBodiesTest.java
```diff
@@ -10,16 +10,18 @@
  *    Evgeny Mandrikov - initial API and implementation
  *
  *******************************************************************************/
-package org.jacoco.ant;
+package org.jacoco.core.test.validation.java21;
```
marchof: I guess the package should be `java25`?

marchof (changes requested):

org.jacoco.core.test.validation.java25/src/org/jacoco/core/test/validation/java21/FlexibleConstructorBodiesTest.java
Godin: @marchof right 🤦 thanks for spotting 👍 fixed

marchof (approved):

PR #2175, opened 2026-06-15
Add unit tests for misleading values of boolean agent options

Godin:


marchof (approved):

Godin:
@marchof I am wondering whether this should be fixed, or just documented? WDYT?

marchof:
@Godin Just documented. We're using the java standard convention here for boolean string interpretation.

PR #2182, opened 2026-06-17
Add Lilliput 2 simulation to LineImplMemoryTest

Godin:
Quoting https://bugs.openjdk.org/browse/JDK-8360700?focusedId=14862778&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14862778

> Also, we have Lilliput 2 (4-byte headers) coming ... we have a shot at upstreaming Lilliput 2 starting in 30 with a default target of 33

### DIFF
--- a/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
+++ b/org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
@@ -83,6 +83,53 @@ public class LineImplMemoryTest {
 	}
 
 	/**
+	 * <a href=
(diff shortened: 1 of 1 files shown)


org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
```diff
@@ -81,6 +81,52 @@ public void currentVM() throws Exception {
 		}
 	}
 
+	/**
+	 * Potentially experimental in JDK 30 and default in JDK 33
```
marchof: AFAIK it is already integrated in Java 27.

marchof (approved):

org.jacoco.benchmarks/src/org/jacoco/core/internal/analysis/LineImplMemoryTest.java
Godin: @marchof Lilliput v1 is integrated and this is about v2 which is coming 😉

PR #2184, opened 2026-06-18
Add validation test for Java 21 JEP 440: nested record pattern

Godin:


marchof (approved):
@Godin I like the assert<Compiler> pattern you're using here. It makes the test fixture very readable because you see directly in the source code what is expected for the different compilers.

Something worth pulling up in the base class and use it at other places too (separate PR of course)?

PR #2187, opened 2026-06-24
Simplify future updates of bytecode versions

Godin:


org.jacoco.core.test/src/org/jacoco/core/internal/instr/InstrSupportTest.java
```diff
@@ -139,7 +152,7 @@ public void needFrames_should_return_true_for_versions_greater_than_or_equal_to_
 		assertTrue(InstrSupport.needsFrames(Opcodes.V25));
 		assertTrue(InstrSupport.needsFrames(Opcodes.V26));
 		assertTrue(InstrSupport.needsFrames(Opcodes.V27));
-		assertTrue(InstrSupport.needsFrames(Opcodes.V27 + 1));
+		assertTrue(InstrSupport.needsFrames(InstrSupport.BYTECODE_VERSION_MAX));
```
Godin, line 155: Looks awkward to me. Also AFAIR in past I even challenged value of this test written this way.

@marchof given that you suggested to reduce number of places requiring update - maybe you can suggest something for this place too?

org.jacoco.core.test/src/org/jacoco/core/instr/ClassFileVersionsTest.java
```diff
@@ -207,7 +207,7 @@ public void test_27() throws IOException {
 
 	@Test
 	public void test_28() throws IOException {
-		testVersion(V27 + 1, true);
```
marchof, line 210: I would rework this class and just have two methods:

* `should_not_generate_frame_until_java_5`
* `should_generate_frame_since_java_6`

Both are implemented with a for loop.

org.jacoco.core.test/src/org/jacoco/core/internal/instr/InstrSupportTest.java
marchof: For loop?

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -112,11 +116,24 @@ protected Collection<String> additionalClassesForAnalysis() {
 		return Collections.emptyList();
 	}
 
+	/**
```
marchof, line 119: I'm confused about the purpose of this test here? Isn't that property already tested in the referenced tests? If you see an added value here please add a comment about the purpose.

PR #2188, opened 2026-06-24
Add validation test for Java 27 JEP 532: primitive types in patterns

Godin:


org.jacoco.core.test.validation.java27/src/org/jacoco/core/test/validation/java27/targets/PrimitivePatternsTarget.java
```diff
@@ -0,0 +1,51 @@
...
+		default -> // assertEmpty()
+			"default"; // assertFullyCovered()
+		};
+	}
+
+	public static void main(String[] args) {
+		nestedInRecordInInstanceof(new R("", 1));
+		nestedInRecordInInstanceof(new Object());
+
+		nop(nestedInRecordInSwitch(new R("", 1)));
```
marchof, line 46: What is the reason for the `nop()` wrapper? Are we forced to consume the return value?

marchof (approved):

org.jacoco.core.test.validation.java27/src/org/jacoco/core/test/validation/java27/targets/PrimitivePatternsTarget.java
Godin: @marchof otherwise IntelliJ IDEA shows warnings that result is ignored 😞

Issue #287, opened 2015-02-24
Add exec file Compatibility Test


marchof:
To avoid regressessions like #286 we need a automated compatibility test for exec files.

marchof:
To verify compatibility we need the following setup:
* A stable set of class files with all different kind of control flows: The TargetXX classes in org.jacoco.core.test.validation.targets would be a representative set. To keep the classes stable a JAR file should be created and versioned.
* A stable exec file created for the baseline of the latest exec file version: By loading all TargetXX classes a exec file will be created as a baseline and versioned.
* A new compatibility test: For every build a test loads the classes from the JAR and verifies that the versioned exec file has the exact same content.

Whenever we increase the exec file version on purpose a new baseline needs to be created:
* Optional: Build a new JAR with the reference classes (just to capture all new TargetXX classses).
* An new versioned exec file as a baseline

@Godin Any comments regarding the test setup?

Godin:
@marchof I need to think a little bit about automated compatibility test...

But I don't think that improvement of a test suite should block release of 0.7.4 with fixes for observed regressions. Also we probably should do 0.7.4 really soon?

Godin:
@marchof This sounds like an extension of idea of filter-samples, but in its case you wanted to do a single shot to generate ASM source code which would define test scenarios, whereas in this case you'd like to lock content of class-file. I consider that storage of binary files in repository - is a bad idea, so I propose following setup:
* Dedicated project to hold sources of validation classes, let's call it "jacoco-test-cases". To keep it stable, versioned and usable for our other projects - releases should go into Maven Central. However it can have separate lifecycle and so versioning. Binary diff with its previous version(s) could help to ensure stability of releases. Releases will also help to track differences between different versions of compilers. And so we probably would be able to rely on automated builds to detect changes in newer versions of compilers. I expect tracking different compilers automatically using different Maven coordinates such as groupId = "org.jacoco.jacoco-test-cases" artifactId="jdk5" without duplication of source code for those test cases and allowing to release at any time later artifactId="jdk6" or even something like "jdk6_u45" for the same version of test cases without name conflicts. And why not - it can be used for creation of ASM source code for test scenarios.
* Dedicated project to hold baselines with a dependency on previous one and on JaCoCo release, let's call it "jacoco-baseline". Again - released into Maven Central. Its version can strictly follow baseline version of JaCoCo from which it was built.
* JaCoCo itself will depend on previous two to ensure compatibility.

[...]

PR #2192, opened 2026-06-26
Add utilities for creating snapshots of method bytecode

Godin:
#1770 is helpful, but has some limitations. In particular comparison of cases (diff) is not straightforward.

This is the next step to simplify creation and maintenance of filters and their tests.

The following step after this will be to allow replacement of current filters tests by snapshots with assertions written as comments directly in snapshots ala comments in validation tests.

---

After which for example will be much easier than now to understand what and how to fix/improve in `RecordPatternFilter` for modern javac versions - see below examples of some snapshots diffs.

---

Also snapshots look useful for places other than filters - see for example https://github.com/jacoco/jacoco/issues/287

## Examples of snapshot diffs after renaming of labels

Side-effect of https://openjdk.org/jeps/455 preview in Java 23:

```diff
--- 1/org.jacoco.core.test/snapshots/RecordPatternsTarget/example.javac_21_22.txt
+++ 2/org.jacoco.core.test/snapshots/RecordPatternsTarget/example.javac_23_24_25.txt
@@ -13,6 +13,10 @@
    L_component_end_x
     ISTORE 4
     ILOAD 4
+    ISTORE 5
+    ICONST_1
+    IFEQ L6
+    ILOAD 4
     ISTORE 2
    L7
     ALOAD 1
@@ -21,6 +25,10 @@
    L_component_end_y
     ISTORE 4
     ILOAD 4
+    ISTORE 5
+    ICONST_1
+    IFEQ L6
+    ILOAD 4
     ISTORE 3
    L_body
     ILOAD 2
```

```diff
--- 1/org.jacoco.core.test/snapshots/UnnamedPatternTarget/example.javac_22.txt
+++ 2/org.jacoco.core.test/snapshots/UnnamedPatternTarget/example.javac_23_24_25.txt
@@ -11,12 +11,20 @@
    L_component_start_x
     INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.x ()I
    L_component_end_x
-    ISTORE 4
+    ISTORE 2
+    ILOAD 2
+    ISTORE 3
+    ICONST_1
+    IFEQ L6
     ALOAD 1
    L_component_start_y
     INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.y ()I
    L_component_end_y
-    ISTORE 4
+    ISTORE 2
+    ILOAD 2
+    ISTORE 3
+    ICONST_1
+    IFEQ L6
    L_body
     INVOKESTATIC org/jacoco/core/test/validation/targets/Stubs.nop ()V
    L6
```

https://openjdk.org/jeps/456 in Java 22:

```diff
--- 1/org.jacoco.core.test/snapshots/RecordPatternsTarget/example.javac_21_22.txt
+++ 2/org.jacoco.core.test/snapshots/UnnamedPatternTarget/example.javac_22.txt
@@ -2,31 +2,23 @@
     TRYCATCHBLOCK L_component_start_y L_component_end_y L_handler java/lang/Throwable
    L_method_start
     ALOAD 0
-    INSTANCEOF org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point
+    INSTANCEOF org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point
     IFEQ L6
     ALOAD 0
-    CHECKCAST org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point
+    CHECKCAST org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point
     ASTORE 1
     ALOAD 1
    L_component_start_x
-    INVOKEVIRTUAL org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point.x ()I
+    INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.x ()I
    L_component_end_x
     ISTORE 4
-    ILOAD 4
-    ISTORE 2
-   L7
     ALOAD 1
    L_component_start_y
-    INVOKEVIRTUAL org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point.y ()I
+    INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.y ()I
    L_component_end_y
     ISTORE 4
-    ILOAD 4
-    ISTORE 3
    L_body
-    ILOAD 2
-    ILOAD 3
-    IADD
-    INVOKESTATIC org/jacoco/core/test/validation/targets/Stubs.nop (I)V
+    INVOKESTATIC org/jacoco/core/test/validation/targets/Stubs.nop ()V
    L6
     GOTO L_after
    L_handler
@@ -43,8 +35,6 @@
    L_after
     RETURN
    L_method_end
-    LOCALVARIABLE x I L7 L6 2
-    LOCALVARIABLE y I L_body L6 3
     LOCALVARIABLE o Ljava/lang/Object; L_method_start L_method_end 0
     LINENUMBER 5 L_method_start
     LINENUMBER 6 L_body
```

Removal of unnecessary instructions in javac 26:

```diff
--- 1/org.jacoco.core.test/snapshots/RecordPatternsTarget/example.javac_23_24_25.txt
+++ 2/org.jacoco.core.test/snapshots/RecordPatternsTarget/example.txt
@@ -12,8 +12,6 @@
     INVOKEVIRTUAL org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point.x ()I
    L_component_end_x
     ISTORE 4
-    ILOAD 4
-    ISTORE 5
     ICONST_1
     IFEQ L6
     ILOAD 4
@@ -24,8 +22,6 @@
     INVOKEVIRTUAL org/jacoco/core/test/validation/java21/targets/RecordPatternsTarget$Point.y ()I
    L_component_end_y
     ISTORE 4
-    ILOAD 4
-    ISTORE 5
     ICONST_1
     IFEQ L6
     ILOAD 4
```

```diff
--- 1/org.jacoco.core.test/snapshots/UnnamedPatternTarget/example.javac_23_24_25.txt
+++ 2/org.jacoco.core.test/snapshots/UnnamedPatternTarget/example.txt
@@ -12,8 +12,6 @@
     INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.x ()I
    L_component_end_x
     ISTORE 2
-    ILOAD 2
-    ISTORE 3
     ICONST_1
     IFEQ L6
     ALOAD 1
@@ -21,8 +19,6 @@
     INVOKEVIRTUAL org/jacoco/core/test/validation/java22/targets/UnnamedPatternTarget$Point.y ()I
    L_component_end_y
     ISTORE 2
-    ILOAD 2
-    ISTORE 3
     ICONST_1
     IFEQ L6
    L_body
```


org.jacoco.core.test/src/org/jacoco/core/test/snapshot/MethodSnapshot.java
```diff
@@ -0,0 +1,166 @@
...
+	 * {@code newSnapshotFile} and throws {@link ComparisonFailure}.
+	 *
+	 * {@code snapshotFile} can be modified by
+	 * <ul>
+	 * <li>renaming labels</li>
+	 * <li>adding comments between instructions</li>
+	 * </ul>
+	 * without impact on this comparison.
+	 */
+	public static void compare(final Class<?> targetClass,
```
marchof: In Java core `compare` has a return value and does not throw an exception. In JUnit methods which throw an exception are typically prefixed with `assert`. So I propose to call this method `assertSnapshotEquals()`or just `assertEquals()`.

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
```diff
@@ -149,6 +150,19 @@ private void saveBytecodeRepresentations(final byte[] classBytes,
 		asmWriter.close();
 	}
 
+	/**
+	 * {@link MethodSnapshot#compare(Class, String, File, File)} with predefined
+	 * file paths.
+	 */
+	protected final void snapshot(final Class<?> targetClass,
```
marchof: See other comment about naming:

```suggestion
	protected final void assertEquals(final Class<?> targetClass,
```

org.jacoco.core.test.validation.java21/src/org/jacoco/core/test/validation/java21/RecordPatternsTest.java
```diff
@@ -26,6 +29,19 @@ public RecordPatternsTest() {
...
 	}
 
+	@Test
+	public void bytecodeSnapshot() throws Exception {
+		if (!isJDKCompiler) {
+			return;
+		}
+		snapshot(RecordPatternsTarget.class, "instanceofOperator",
+				JavaVersion.current().isBefore("23") //
+						? "example.javac_21_22"
```
marchof: Is example "example" the final name? Or better "snapshot"?

marchof (approved):

org.jacoco.core.test/src/org/jacoco/core/test/validation/ValidationTestBase.java
Godin: @marchof IMO `assertEquals` is way too generic name, heavily overloaded, and from signature unclear equals to what, so what about `assertSnapshot`?

org.jacoco.core.test.validation.java21/src/org/jacoco/core/test/validation/java21/RecordPatternsTest.java
Godin: This is just an example, which will be reverted prior to merge of this PR and actual snapshots will be added in separate PRs.

org.jacoco.core.test.validation.java22/src/org/jacoco/core/test/validation/java22/UnnamedPatternTest.java
```diff
@@ -25,6 +27,19 @@ public UnnamedPatternTest() {
 		super(UnnamedPatternTarget.class);
 	}
 
+	@Test
+	public void bytecodeSnapshot() throws Exception {
+		if (!isJDKCompiler) {
+			return;
+		}
+		snapshot(UnnamedPatternTarget.class, "instanceofPattern",
```
marchof: Idea: Maybe we can have some generic lookup tool that iterates through the Java versions and picks the file which belongs to the current Java version?

PR #2219, opened 2026-07-31
Add validation test for ranges in Kotlin `when`

### LINKED ISSUE #2140: Add validation test for ranges in Kotlin `when`
See also https://youtrack.jetbrains.com/issue/KT-19162/Cover-enum-cases-by-ranges-in-when-statement

Eljees:
Closes #2140

Adds `KotlinWhenRangeTarget` / `KotlinWhenRangeTest`. No production code is touched — the test documents the coverage that JaCoCo currently produces for range conditions in `when`.

What the new target pins down:

* a range case is **not** compiled into a `tableswitch` / `lookupswitch` like a case with constants. Each `in a..b` case is compiled into a comparison with both bounds whose result is materialized as a boolean, which is then compared with zero:

  ```
  ICONST_1
  ILOAD 2
  IF_ICMPGT out_of_range
  ILOAD 2
  BIPUSH 6
  IF_ICMPGE out_of_range
  ...
  IFEQ next_case
  ```

  so a **single range case carries six branches**, of which the last pair duplicates the outcome of the two comparisons. `assertFullyCovered(0, 6)` in the target records exactly that.

* covering those six branches requires a subject below the lower bound, not only one above the upper bound — the `main` method is written accordingly (for example `whenRangeAndConstant(-1)`, because the preceding `0 ->` case would otherwise hide that branch).

* the same shape appears for `!in`, for a `when` statement without `else`, for a range case that follows a constant case (2 branches for the constant, 6 for the range), and for a `when` without a subject.

Verified with the Kotlin version currently used by the module (2.4.10):

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 -- KotlinWhenRangeTest
[INFO] Tests run: 288, Failures: 0, Errors: 0, Skipped: 0 -- org.jacoco.core.test.validation.kotlin
[INFO] BUILD SUCCESS
```

See also https://youtrack.jetbrains.com/issue/KT-19162, referenced in the issue.

---
Written with AI assistance; reviewed, built and tested by me.

### DIFF
--- a/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/KotlinWhenRangeTest.java
+++ b/org.jacoco.core.test.validation.kotlin/src/org/jacoco/core/test/validation/kotlin/KotlinWhenRangeTest.java
@@ -1,0 +1,27 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2026 Mountainminds GmbH & Co. KG and Contributors
+ * This program and the accompanying materials are made available under
+ * the terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Evgeny Mandrikov - initial API and implementation
+ *
+ *******************************************************************************/
+package org.jacoco.core.test.validation.kotlin;
+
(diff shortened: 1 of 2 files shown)


Eljees:
Ping - open since 31 July with no review. It only adds a validation test for ranges in a Kotlin `when`, so there is no production code change.

